PR #2, opened 2015-10-28
README correction for publicIp flag

tgross:
@misterbisson this corrects the README to match the behavior of the `publicIp` flag in the code, and also adds this flag to the Nginx portion of the demo.


misterbisson:
:peach:es to me (sorry for the delay)

PR #4, opened 2015-10-30
First steps to allowing multiple discovery backends

bbox-kula:
I believe this is enough of a first step to removing the assumption that the only discovery backend is consul. The general idea is that each backend implements the DiscoveryService interface, and then code is added in config.go to pull that backend's config from the appropriately named json field. 

May not quite be idiomatic go (I'm still getting familiar with it), but I think this is the bulk of the work.


src/containerbuddy/config.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 type Config struct {
-	DiscoveryUri string           `json:"consul"` // TODO: allow choice of discovery here
+	Consul	     *string          `json:"consul"`
```
tgross: We should mark this field as `omitempty` so that it's not serialized if we want to do that in the future. I'd also prefer we keep this as a string rather than a pointer and do a check for `""` rather than `nil` in the switch block on line 76. What you have here works but avoiding a scenario where we pass around `nil` will reduce the risk of future bugs.

tgross:
>  The general idea is that each backend implements the DiscoveryService interface, and then code is added in config.go to pull that backend's config from the appropriately named json field.

I like the approach you took here. I just have the one minor comment and then I'd be happy to merge this in.

bbox-kula:
Good point, and a minor enough change to make. I've sent it back up.

tgross:
LGTM. Thanks for the contribution.

Issue #3, opened 2015-10-29
starting the example locally does not work 


aschmid:
so there are multiple issues with the local deployment…

```
cd ./examples/nginx
./start -p example -f docker-compose-local.yml
```
this should be 
```
cd ./examples
./start.sh -p example -f docker-compose-local.yml
```
there is no executable called `start` in `./examples/nginx`, it is in `./examples` and its called `start.sh`

the build and start of nginx is broken.
when it tries to start nginx i get this error

```
Successfully built a09f1b400979
Creating example_nginx_1...
Cannot start container c0bd9d5f1e31674146c0800c884a273580fa35b6f93356c66aaeea2e0fe449d9: [8] System error: exec: "opt/containerbuddy/containerbuddy": stat opt/containerbuddy/containerbuddy: no such file or directory
```

this is obviously because there is no file like `/opt/containerbuddy/containerbuddy` in the repo nor in the built image.

tgross:
@aschmid https://github.com/joyent/containerbuddy/pull/5 should fix these problems.

mjgsilva:
I guess you will find another bug after solving this one, check my [issue](https://github.com/joyent/containerbuddy/issues/8).

tgross:
@aschmid that PR has been merged in if you want to give it another go.

misterbisson:
@aschmid I'm going to close this, if you don't mind. @tgross' PR fixes this for me.

Please let us know if you're still seeing a problem after updating, and re-open or open a new issue with any problems.

PR #5, opened 2015-11-02
README fixes for running example locally

tgross:
For issue #3.
- start script is named ./start.sh
- we need to make sure build artifacts are present for docker-compose-local.yml
  to be able to build the containers
- note that scale-up doesn't work when the containers don't have their own IPs
- (also added contribution guidelines)

@misterbisson and/or @aschmid to review please

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ coverage.out
 build/
 release/
 containerbuddy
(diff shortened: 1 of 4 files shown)


misterbisson:
:+1:

PR #12, opened 2015-11-09
new docker version change ip on container restart

frascuchon:
Hi, 
When service restarts, no ip info it being updated on consul, so change hostname by host ip on service id for service registration.

Best regards

### DIFF
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -91,12 +91,12 @@ func loadConfig() *Config {
 		backend.onChangeArgs = strings.Split(backend.OnChangeExec, " ")
 	}
 
-	hostname, _ := os.Hostname()
 	for _, service := range config.Services {
-		service.Id = fmt.Sprintf("%s-%s", service.Name, hostname)
+		hostIP := getIp(service.IsPublic)
+		service.Id = fmt.Sprintf("%s-%s", service.Name, hostIP)
 		service.discoveryService = discovery
 		service.healthArgs = strings.Split(service.HealthCheckExec, " ")
-		service.ipAddress = getIp(service.IsPublic)
+		service.ipAddress = hostIP
 	}
 
 	return config


src/containerbuddy/config.go
```diff
@@ -91,12 +91,12 @@ func loadConfig() *Config {
 		backend.onChangeArgs = strings.Split(backend.OnChangeExec, " ")
 	}
 
-	hostname, _ := os.Hostname()
 	for _, service := range config.Services {
-		service.Id = fmt.Sprintf("%s-%s", service.Name, hostname)
+		hostIP := getIp(service.IsPublic)
+		service.Id = fmt.Sprintf("%s-%s", service.Name, hostIP)
```
misterbisson: This changes the format of the `service.Id`, replacing the `hostname` with the `hostIP`, which breaks expectations elsewhere that this is used.

src/containerbuddy/config.go
```diff
@@ -91,12 +91,12 @@ func loadConfig() *Config {
 		backend.onChangeArgs = strings.Split(backend.OnChangeExec, " ")
 	}
 
-	hostname, _ := os.Hostname()
```
misterbisson: The hostname defaults to the Docker container ID, which is important in being able to interact with the container elsewhere. Removing it here breaks other uses.

frascuchon:
Hi,

So, when a container is restarted from docker, the service ip in consul should be updated, since this may change.

Thanks !

tgross:
> So, when a container is restarted from docker, the service ip in consul should be updated, since this may change.

@frascuchon I just want to make sure I understand the issue you're seeing here: when a container is stopped and then started elsewhere, its IP address may change. (This doesn't happen when I run `docker-machine` locally or on Joyent's Triton platform, but I agree that we should assume that there might be a platform where this does happen.) So you're proposing this change so that the IP address is updated in Consul.

In that scenario, the [service registration](https://github.com/joyent/containerbuddy/blob/master/src/containerbuddy/consul.go#L36-L45) will happen again and any IP address changes will be updated in Consul. I've verified independently that [`Agent.ServiceRegister`](https://godoc.org/github.com/hashicorp/consul/api#Agent.ServiceRegister]) will overwrite correctly. This means that, for example, if we have the container `app-4fe70bfdef25` running at `172.17.0.20:8000` and it's moved to `172.17.0.21:8000` with the same container ID, Consul will pick up the change.

Am I misunderstanding the scenario you've outlined?

frascuchon:
Hi @tgross. 

You're right. I test in a local environment with a docker-machine (version 0.4.1) and i can see this behavior

tgross:
Ok, looks like it's safe to close this then?

PR #21, opened 2015-12-01
Add interface option to ServiceConfig

### LINKED ISSUE #20: Override interface for ip discovery
I'm trying to combine [Weave](http://weave.works/) and containerbuddy to register weave IPs in consul-dns. Unfortunately, discovery uses the first IP that it finds from the list of interfaces - which in my case is usually the docker bridge, not the one I want (weave interface).

I'd like to configure the containerbuddy to use `ethwe` instead directly, because I know my containers will be on the weave network: 

**Example of my container's interfaces on the Weave network**
```
root@aab54fefc215:/# ifconfig
eth0      Link encap:Ethernet  HWaddr 02:42:ac:11:00:02
          inet addr:172.17.0.2  Bcast:0.0.0.0  Mask:255.255.0.0
          inet6 addr: fe80::42:acff:fe11:2/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:8 errors:0 dropped:0 overruns:0 frame:0
          TX packets:8 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:648 (648.0 B)  TX bytes:648 (648.0 B)

ethwe     Link encap:Ethernet  HWaddr ee:71:f3:9b:ed:4a
          inet addr:10.130.16.1  Bcast:0.0.0.0  Mask:255.255.224.0
          inet6 addr: fe80::ec71:f3ff:fe9b:ed4a/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1410  Metric:1
          RX packets:8 errors:0 dropped:0 overruns:0 frame:0
          TX packets:9 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:648 (648.0 B)  TX bytes:690 (690.0 B)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
```

Adding an interface override like `"interface": "ethwe"` may help me get the correct IP:

```
{
  "consul": "consul:8500",
  "services": [
    {
      "name": "nginx",
      "port": 80,
      "interface": "ethwe",
      "health": "/usr/bin/curl --fail -s http://localhost/health.txt",
      "poll": 10,
      "ttl": 25
    }
  ],
  "backends": [ ... ]
}
```

justenwalker:
- Adds the ability to override the interface from which containerbuddy
  gets its IP address. If the interface is found, it takes precedence
  over the "publicIp" option.

- Fixes #20

### DIFF
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -26,7 +26,7 @@ type ServiceConfig struct {
 	HealthCheckExec  string   `json:"health"`
 	Port             int      `json:"port"`
 	TTL              int      `json:"ttl"`
-	Interfaces       []string `json:"interface"`
+	Interfaces       []string `json:"interfaces"`
 	discoveryService DiscoveryService
 	healthArgs       []string
 	ipAddress        string
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -26,7 +26,6 @@ type ServiceConfig struct {
 	HealthCheckExec  string `json:"health"`
 	Port             int    `json:"port"`
 	TTL              int    `json:"ttl"`
-	IsPublic         bool   `json:"publicIp"` // will default to false
 	Interface        string `json:"interface"`
 	discoveryService DiscoveryService
 	healthArgs       []string
@@ -108,7 +107,9 @@ func loadConfig() (*Config, error) {
 		service.Id = fmt.Sprintf("%s-%s", service.Name, hostname)
 		service.discoveryService = discovery
 		service.healthArgs = strings.Split(service.HealthCheckExec, " ")
-		service.ipAddress = getIp(service.IsPublic,service.Interface)
+		if service.ipAddress, err = getIp(service.Interface); err != nil {
+			return nil, err
+		}
 	}
 
 	return config, nil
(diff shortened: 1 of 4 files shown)


src/containerbuddy/config.go
```diff
@@ -27,6 +27,7 @@ type ServiceConfig struct {
 	Port             int    `json:"port"`
 	TTL              int    `json:"ttl"`
 	IsPublic         bool   `json:"publicIp"` // will default to false
```
tgross: I commented in https://github.com/joyent/containerbuddy/issues/20#issuecomment-161279931 to this effect, but if we're going to have an `Interface` option then we've obviated the need to have the `IsPublic` option. Let's get rid of that in this same PR.

src/containerbuddy/config.go
```diff
@@ -137,7 +138,7 @@ func parseConfig(configFlag string) (*Config, error) {
 }
 
 // determine the IP address of the container
-func getIp(usePublic bool) string {
+func getIp(usePublic bool, interfaceName string) string {
```
tgross: If we get rid of the no-longer-needed `usePublic` arg, we can simplify this function a lot and eliminate the `isPublicIp` function below.

src/containerbuddy/config.go
```diff
@@ -146,6 +147,10 @@ func getIp(usePublic bool) string {
 		// We're assuming each interface has one IP here because neither Docker
 		// nor Triton sets up IP aliasing.
 		ipAddr, _, _ := net.ParseCIDR(ipAddrs[0].String())
+		// Use specific interface if given
+		if interfaceName == intf.Name {
```
tgross: If `interface` has been elided in the config, we should have a safe default.

tgross:
@justenwalker this looks great. Two minor items to keep us in shipping state on master:
- Can you do me a favor and run `go fmt` over it? (`config.go` and `config_test.go` both have complaints.)
- It'd help if you could update the sections of the README that refers to `publicIp`.

misterbisson:
I've been thinking about how the PR will work when I'm moving across environments. When developing on my laptop I might not have Weave running, but the config will fail if it's looking for `ethane`.

Perhaps it needs to accept an array of interfaces in priority order?

justenwalker:
@tgross Updated the formatting. README was updated in e31fffa
@misterbisson Interesting idea. Would it be sufficient to fall back to eth0? Or is it better to supply multiple interfaces.

justenwalker:
@misterbisson I changed it to use a list of interfaces instead of a single interface - let me know if that works for you.

misterbisson:
@justenwalker looks good to me, but I'll defer to @tgross for his review.

justenwalker:
@tgross Not a nitpick at all - I documented it as `interfaces` but forgot to change it in the JSON mapping.

Fixed with 77fce1f

tgross:
All tests pass, `go fmt` is happy. LGTM! Let's merge it!

tgross:
Released in https://github.com/joyent/containerbuddy/releases/tag/0.0.3

PR #22, opened 2015-12-03
Deregister services on SIGTERM

justenwalker:
## Changes

- Handle SIGTERM, send to application
- Add `stopTimeout` to config to manage wait time
- Extend DiscoveryService - add Deregister
- Deregister all services when application exits
- OnStop command to run after application exits

## Use Case

When i run my Containerbuddy images, i want them to automatically deregister from consul when they die, so that they do not clutter up the service namespace. Especially problematic when scaling these containers out with Mesos where applications may be moved around frequently.

## Notes

I'm a bit of a golang noob, so I'm not sure how to add tests for this, since this has some external effects. FWIW, I tested this empirically using a local Mesos and Consul cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ docker exec myapp_1 kill -USR1 1
 
 ```
 
-Docker will automatically deliver a `SIGTERM` with `docker stop`; Not when using `docker kill`.  When Containerbuddy receives a `SIGTERM`, it will propagate this signal to the application and wait for `stopTimeout` seconds before forcing the application to stop. Make sure this timeout is less than the docker stop timout period or services may not deregister from the discovery service backend. If `-1` is given for `stopTimeout`, Containerbuddy will kill the application immediately with `SIGKILL`, but it will still deregister the services.
+Docker will automatically deliver a `SIGTERM` with `docker stop`, not when using `docker kill`.  When Containerbuddy receives a `SIGTERM`, it will propagate this signal to the application and wait for `stopTimeout` seconds before forcing the application to stop. Make sure this timeout is less than the docker stop timeout period or services may not deregister from the discovery service backend. If `-1` is given for `stopTimeout`, Containerbuddy will kill the application immediately with `SIGKILL`, but it will still deregister the services.
 
 *Caveat*: If Containerbuddy is wrapped as a shell command, such as: `/bin/sh -c '/opt/containerbuddy .... '` then `SIGTERM` will not reach Containerbuddy from `docker stop`.  This is important for systems like Mesos which may use a shell command as the entrypoint under default configuration.
 
--- a/README.md
+++ b/README.md
@@ -88,15 +88,16 @@ Other fields:
 
 - `consul` is the hostname and port of the Consul discovery service.
 - `onStart` is the executable (and its arguments) that will be called immediately prior to starting the shimmed application. This field is optional. If the `onStart` handler returns a non-zero exit code, Containerbuddy will exit.
-- `stopTimeout` Optional amount of time in seconds to wait before killing the application. (defaults to `5`) When Containerbuddy receives a SIGTERM, it will propagate this signal to the application and wait this many seconds before forcing the application to stop. Make sure this timeout is less than the docker stop timout period or services may not deregister from the discovery service backend.
+- `stopTimeout` Optional amount of time in seconds to wait before killing the application. (defaults to `5`). Providing `-1` will kill the application immediately.
 
 *Note that if you're using `curl` to check HTTP endpoints for health checks, that it doesn't return a non-zero exit code on 404s or similar failure modes by default. Use the `--fail` flag for curl if you need to catch those cases.*
 
 ### Operating Containerbuddy
 
-Containerbuddy accepts POSIX signals to change its runtime behavior. Currently, Containerbuddy accepts the following signals.
+Containerbuddy accepts POSIX signals to change its runtime behavior. Currently, Containerbuddy accepts the following signals:
+
 - `SIGUSR1` will cause Containerbuddy to mark its advertised service for maintenance. Containerbuddy will stop sending heartbeat messages to the discovery service. The discovery service backend's `MarkForMaintenance` method will also be called (in the default Consul implementation, this deregisters the node from Consul).
-- `SIGTERM` will cause Containerbuddy to send SIGTERM to the application, and eventually exit in a timely manner (as specified by StopTimeout).
+- `SIGTERM` will cause Containerbuddy to send `SIGTERM` to the application, and eventually exit in a timely manner (as specified by `stopTimeout`).
 
 Delivering a signal to Containerbuddy is most easily done by using `docker exec` and relying on the fact that it is being used as PID1.
 
@@ -105,6 +106,9 @@ docker exec myapp_1 kill -USR1 1
 
 ```
 
+Docker will automatically deliver a `SIGTERM` with `docker stop`; Not when using `docker kill`.  When Containerbuddy receives a `SIGTERM`, it will propagate this signal to the application and wait for `stopTimeout` seconds before forcing the application to stop. Make sure this timeout is less than the docker stop timout period or services may not deregister from the discovery service backend. If `-1` is given for `stopTimeout`, Containerbuddy will kill the application immediately with `SIGKILL`, but it will still deregister the services.
+
+*Caveat*: If Containerbuddy is wrapped as a shell command, such as: `/bin/sh -c '/opt/containerbuddy .... '` then `SIGTERM` will not reach Containerbuddy from `docker stop`.  This is important for systems like Mesos which may use a shell command as the entrypoint under default configuration.
 
 ### Contributing
 
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ Other fields:
 
 - `consul` is the hostname and port of the Consul discovery service.
 - `onStart` is the executable (and its arguments) that will be called immediately prior to starting the shimmed application. This field is optional. If the `onStart` handler returns a non-zero exit code, Containerbuddy will exit.
+- `stopTimeout` Optional amount of time in seconds to wait before killing the application. (defaults to `5`) When Containerbuddy receives a SIGTERM, it will propagate this signal to the application and wait this many seconds before forcing the application to stop. Make sure this timeout is less than the docker stop timout period or services may not deregister from the discovery service backend.
 
 *Note that if you're using `curl` to check HTTP endpoints for health checks, that it doesn't return a non-zero exit code on 404s or similar failure modes by default. Use the `--fail` flag for curl if you need to catch those cases.*
 
@@ -95,6 +96,7 @@ Other fields:
 
 Containerbuddy accepts POSIX signals to change its runtime behavior. Currently, Containerbuddy accepts the following signals.
 - `SIGUSR1` will cause Containerbuddy to mark its advertised service for maintenance. Containerbuddy will stop sending heartbeat messages to the discovery service. The discovery service backend's `MarkForMaintenance` method will also be called (in the default Consul implementation, this deregisters the node from Consul).
+- `SIGTERM` will cause Containerbuddy to send SIGTERM to the application, and eventually exit in a timely manner (as specified by StopTimeout).
 
 Delivering a signal to Containerbuddy is most easily done by using `docker exec` and relying on the fact that it is being used as PID1.
 
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ The format of the JSON file configuration is as follows:
 {
   "consul": "consul:8500",
   "onStart": "/opt/containerbuddy/onStart-script.sh",
+  "stopTimeout": 5,
   "services": [
     {
       "name": "app",
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -72,7 +72,7 @@ func (s *ServiceConfig) Deregister() {
 
 const (
 	// Amount of time to wait before killing the application
-	DEFAULT_STOP_TIMEOUT int = 5
+	defaultStopTimeout int = 5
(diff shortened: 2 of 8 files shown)


tgross:
Normally I'd suggest we open an issue first, but the way you've structured this PR makes it ideal for discussing the feature idea so this is great. Thanks for the PR @justenwalker. Some points of discussion:

> i want them to automatically deregister from consul when they die

Just to be specific, this patch catches `SIGTERM` which means we're catching `docker stop` (or really the [equivalent remote API call](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#stop-a-container)) but not `docker kill` or involuntary exit like a crash. I do like that split in behavior though -- it means that containers intentionally removed will deregister but ones that have crashed will remain in Consul until you reap them separately, which will help out for debugging.

I'd like if we documented the intended behavior with respect to Docker commands specifically. What you have in the README is accurate but it'll improve usability if we explicitly say "this is what happens with `docker stop` vs `docker kill`" (ex. the Docker docs say they fire SIGTERM + SIGKILL so you don't have to dig into source to verify the behavior).

> so I'm not sure how to add tests for this

Tricky... You don't terminate Containerbuddy itself in the signal handler (which is correct, because the Containerbuddy process will clean itself up once it gets the exit code from the shimmed application). So you could write a test that runs an `executeAndWait` on something like `sleep` and then fires the appropriate signals and checks the exit code to make sure they were caught.

src/containerbuddy/config.go
```diff
@@ -63,6 +66,10 @@ func (s *ServiceConfig) MarkForMaintenance() {
 	s.discoveryService.MarkForMaintenance(s)
 }
 
+func (s *ServiceConfig) Deregister() {
```
tgross: I like that you've extended the `DiscoveryService` such that if `MarkForMaintenance` and `Deregister` need to be separate implementations for a non-Consul discovery backend we can do that easily.

src/containerbuddy/discovery.go
```diff
@@ -4,4 +4,5 @@ type DiscoveryService interface {
 	SendHeartbeat(*ServiceConfig)
 	CheckForUpstreamChanges(*BackendConfig) bool
 	MarkForMaintenance(*ServiceConfig)
+	Deregister(*ServiceConfig)
```
tgross: I like that you've extended the `DiscoveryService` such that if `MarkForMaintenance` and `Deregister` need to be separate implementations for a non-Consul discovery backend we can do that easily.

src/containerbuddy/main.go
```diff
@@ -60,6 +62,14 @@ func main() {
 	select {}
 }
 
+func deregisterServices(config *Config) {
+	log.Println("Deregister All Services")
```
tgross: Once we've got this closer to being merged we might want to consider cleaning up some of the log noise in these functions.

src/containerbuddy/config.go
```diff
@@ -95,6 +102,10 @@ func loadConfig() (*Config, error) {
 		return nil, errors.New("More than one discovery backend defined")
 	}
 
+	if config.StopTimeout == 0 {
+		config.StopTimeout = 10
```
tgross: The README says we default to 5. Also, [`terminate`](https://github.com/joyent/containerbuddy/pull/22/files#diff-536bffba9bba22f8afea4ecb1036fe5cR30) checks for `> 0`. It looks like we're trying to catch the struct's default value here so that we default to 5. I think this makes it impossible to intentionally set the value to 0. Maybe that's ok?

justenwalker: I think maybe I should document this a bit better. My intention here was to allow someone to set stopTimeout = -1 and get SIGKILL semantics on the managed application (while still allowing Containerbuddy to clean up)

tgross: > My intention here was to allow someone to set stopTimeout = -1

Ah, ok I like the intent. Let's just make it clear.

justenwalker:
I'm going to add some documentation like the following in Operating Container Buddy:


Docker will automatically deliver a SIGTERM when you use `docker stop`; Not when using `docker kill`. *Caveat*: If container buddy is executed in a shell such as: `/bin/sh -c '/opt/containerbudy .... '` then SIGTERM will not reach containerbuddy from `docker stop`.  This is important for systems like Marathon on Mesos which do this by default.

tgross:
> This is important for systems like Marathon on Mesos which do this by default.

Just for clarity: does Mesos/Marathon just fire SIGTERM to the shell process and pray cleanup works, or are you saying that you don't want to use `docker stop` directly on a container launched by Marathon/Mesos?

justenwalker:
Mesos/Marathon will issue a `docker stop` on the container when shutting it down, (at least if you specify `docker_stop_timeout` in your mesos-slave config).  However, the problem is when launching the container - the default is to use a shell executor, which blocks signals from propagating.

From Mesos Docs - [CommandInfo to run Docker images](http://mesos.apache.org/documentation/latest/docker-containerizer/)
> To run a docker image with the default command (ie: docker run image), the CommandInfo’s value must not be set. If the value is set then it will override the default command.
> To run a docker image with an entrypoint defined, the CommandInfo’s shell option must be set to false. If shell option is set to true the Docker Containerizer will run the user’s command wrapped with `/bin/sh -c` which will also become parameters to the image entrypoint.

tgross:
> However, the problem is when launching the container - the default is to use a shell executor, which blocks signals from propagating.

That's unfortunate but I don't see any way around it.

justenwalker:
It's not Containerbuddy's responsibility to manage that - but I think it would be a good idea to make people aware of this because it manifests as "My services are not deregistering from Consul when Mesos shuts them down".  I spent all night tracking this down, so i'd hate for anyone else to have to suffer through it.

tgross:
> It's not Containerbuddy's responsibility to manage that - but I think it would be a good idea to make people aware of this because it manifests as "My services are not deregistering from Consul when Mesos shuts them down". I spent all night tracking this down, so i'd hate for anyone else to have to suffer through it.

I'm in total agreement there.

justenwalker:
Added some tests for the SIGTERM handle.  

Also, found out that we can only call handleSignals once, otherwise we'll get duplicate handles - so I had to dump the SIGTERM test into a single function along with SIGUSR1

justenwalker:
@tgross Updated tests and removed panic on cmd not found.

misterbisson:
This opens up the door to more than just deregistering services. Consider running Couchbase in Docker for example. The [blueprint and demo makes deploying Couchbase and scaling it **up** easy](https://www.joyent.com/blog/couchbase-in-docker-containers) [repo](https://github.com/misterbisson/clustered-couchbase-in-containers), but scaling down requires more steps.

If the `SIGTERM` handler could also execute a user-defined executable (and wait for it), it would allow us to mark the Couchbase node for removal from the cluster and automatically rebalance the data to the remaining nodes before stopping this one.

I haven't tested it, but I think the right command to call would be:

```
couchbase-cli rebalance -c 127.0.0.1:8091 -u $COUCHBASE_USER -p $COUCHBASE_PASS --server-remove=${IP_PRIVATE}:8091
```

And when that is done, it should be safe to stop the container.

justenwalker:
@misterbisson Added another config for onStop support - Could be used for any arbitrary commands necessary to clean up after a service exits.

misterbisson:
@justenwalker I was returning to the ticket to suggest I should move that to a ticket and let this PR move forward, but you were too quick!

> `onStop` is the executable (and its arguments) that will be called immediately after the shimmed application exits. This field is optional. If the `onStop` handler returns a non-zero exit code, Containerbuddy will exit with this code rather than the application's exit code.

I like this, a lot. I do have a question about "will be called immediately after the shimmed application exits," though. My usage scenario requires that this handler be called _before_  the shimmed application exits, and that Containerbuddy wait for its exit before continuing. 

Semantically, `onStart` is actually executed _before_ the application starts. Would it be okay for `onStop` to similarly execute _before_ the actual stop? Or, did you have a specific usage in mind that requires calling `onStop` after the shimmed app's exit?

src/containerbuddy/main.go
```diff
@@ -48,10 +48,21 @@ func main() {
 		// Run our main application and capture its stdout/stderr.
 		// This will block until the main application exits and then os.Exit
 		// with the exit code of that application.
-		code, err := run(flag.Args())
+		config.Command = getCmd(flag.Args())
+		code, err := executeAndWait(config.Command)
 		if err != nil {
 			log.Println(err)
 		}
+		deregisterServices(config)
```
misterbisson: Should this be called in the top of [`terminate`](https://github.com/joyent/containerbuddy/pull/22/files#diff-536bffba9bba22f8afea4ecb1036fe5cR30) in `signals.go` so that the requests to the service can be stopped before the service truly stops responding to them?

In this position, the shimmed app has been terminated before being deregistered, yes?

src/containerbuddy/config.go
```diff
@@ -63,6 +68,15 @@ func (s *ServiceConfig) MarkForMaintenance() {
 	s.discoveryService.MarkForMaintenance(s)
 }
 
+func (s *ServiceConfig) Deregister() {
+	s.discoveryService.Deregister(s)
+}
+
+const (
+	// Amount of time to wait before killing the application
+	DEFAULT_STOP_TIMEOUT int = 5
```
tgross: Golang style is not to all-caps these. Should just be lower-case so that we're not exporting it.

justenwalker:
@misterbisson In my mind, OnStop makes more sense to happen after the application exits rather than before - so that we know the service isn't interacting with anything before we send cleanup commands (like Finally block).  Perhaps I should undo this commit and move it to a separate PR - as I've kind of incorporated too many changes into this PR at once?

src/containerbuddy/main.go
tgross: > Should this be called in the top of terminate in signals.go so that the requests to the service can be stopped before the service truly stops responding to them?

If we move the call to `deregisterServices` then we'll miss cases where the application exits normally without a `docker stop` (processes don't call `SIGTERM` on themselves when they exit).

But your concern is well-founded. Maybe we need to *also* include the `deregisterServices` in the beginning of `terminate`; we'd need to either handle the error from the discovery service on the second call or set a flag that we've already terminated the process.

tgross:
> In my mind, OnStop makes more sense to happen after the application exits rather than before - so that we know the service isn't interacting with anything before we send cleanup commands (like Finally block)

I agree. A use case here would be backing up state to somewhere external after the application has exited (and therefore given up hold on file handles).

> Perhaps I should undo this commit and move it to a separate PR - as I've kind of incorporated too many changes into this PR at once?

I think that's probably a good idea. That way we can discuss that new feature separately.

src/containerbuddy/main.go
justenwalker: I don't see why the application should continue running after deregister is called. My intention for the cleanup method is to be analagous to a try / finally block.  

stopTimeout is just the time that Containerbuddy will wait before killing the application itself (after sending SIGTERM) and proceeding with the cleanup.

tgross: Right, but as-written you're running the application and then running deregister afterwards. Which means the application exits and can't serve requests but the discovery service will still think it's healthy.

justenwalker:
@tgross @misterbisson backed out the OnStop change. Will do separate PR

justenwalker:
@misterbisson #28 created to discuss OnStop

src/containerbuddy/main.go
justenwalker: I agree. I can move it to signals.go after terminate.

tgross: > Is it possible that while this is going on, the poll function re-registers it though?

No because we check for maintenance mode for running any poll function: https://github.com/joyent/containerbuddy/blob/288c18cc9df4bd15f64be59456f9f362655af30e/src/containerbuddy/main.go#L74

justenwalker:
@tgross Im stopping all of the polling threads, then deregistering the service, and finally terminating the application.

I did not use the maintenance mode - maybe I should?

tgross:
> I did not use the maintenance mode - maybe I should?

My first pass thru implementing maintenance mode I did this the same way you did here but then I realized we couldn't un-stop the polling threads so I used the `toggleMaintenanceMode` function at the top of the SIGUSR1 handler.

What you've done probably works just as well or better for the SIGTERM handler. I'm a little worried about implementation drift between maintenance and deregister... the `deregisterServices` code is similar to the loop in the SIGUSR1 handler except with a different function called on each service. Maybe we can factor that loop out and have func as one of its arguments?

justenwalker:
@tgross I factored out that logic into its own function: forAllServices

README.md
```diff
@@ -88,13 +88,16 @@ Other fields:
 
 - `consul` is the hostname and port of the Consul discovery service.
 - `onStart` is the executable (and its arguments) that will be called immediately prior to starting the shimmed application. This field is optional. If the `onStart` handler returns a non-zero exit code, Containerbuddy will exit.
+- `stopTimeout` Optional amount of time in seconds to wait before killing the application. (defaults to `5`). Providing `-1` will kill the application immediately.
```
tgross: Last comment: we should show this field in the JSON example above.

tgross:
Ok, @justenwalker I've verified the tests pass, that we're `go fmt` clean, and that the example application builds and runs. I added one last documentation comment and then I think we can merge this in.

tgross:
> Add stopTimeout to example json

Ok, this looks good to merge! Thanks for the hard work in bringing this home @justenwalker!

misterbisson:
@justenwalker thanks for creating the new ticket. I've [picked up the conversation there](https://github.com/joyent/containerbuddy/issues/28#issuecomment-162005735), as you've probably already seen ;-)

This is a good feature to add, thanks for working and thinking it through with us!

tgross:
Released in https://github.com/joyent/containerbuddy/releases/tag/0.0.3

PR #23, opened 2015-12-03
Bump version to 0.0.3

tgross:
Also tweaked release makefile target to push tags to joyent/origin to make doing releases involve one less manual step.

cc @misterbisson to give this a LGTM so we can make a new release.

### DIFF
--- a/makefile
+++ b/makefile
@@ -5,7 +5,7 @@ SHELL := /bin/bash
 
 .PHONY: clean test consul run example ship
 
-VERSION := 0.0.2-alpha
+VERSION := 0.0.3-alpha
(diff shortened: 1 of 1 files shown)


justenwalker:
Noticed a bug with #20 and fixed it with #25.  If you're releasing a version, it might be good to get this in before this PR.

tgross:
Will do, thanks @justenwalker

tgross:
I ended up with some annoying spurious merge commits in this branch. I'm going to un-muss it up for cleanliness and close this PR.

tgross:
Nevermind, my history is too muddy in this branch so fixing it up would be evil at this point.

PR #25, opened 2015-12-03
Fix interface configuration in nginx.json example

justenwalker:
PR to add interfaces had a bug in the example application which used the old "interface" configuration.  Example still works, but would be confusing, since it is silently ignored.


tgross:
LGTM. Thanks for catching this.

PR #29, opened 2015-12-04
Clean up release process

tgross:
@misterbisson this cleans up the release process a bit so that it's more obvious what needs to be done and to prevent missing parts (like shipping the updated example app images).

Updates to makefile:
- fix a bug in pushing tags to repo
- ensure we ship new example app with each release
- document use of VERSION for `release` target


misterbisson:
:house_with_garden: :walking:

Issue #30, opened 2015-12-04
Add TravisCI Support


justenwalker:
It would be great if we could get automated builds on TravisCI

Could make reviewing easier, since Travis will update the build status on the PR.

misterbisson:
It's interesting you should ask that ;-) @tgross and I were just discussing it earlier.

justenwalker:
Great minds, yadda yadda.

tgross:
I pushed the badge to the README directly so that we can push a build into Travis and get this kicked off. Once the first build is good I'll close this issue.

tgross:
Interesting... the test for the new signal handler failed: https://travis-ci.org/joyent/containerbuddy/builds/94954762

I'm going to close this ticket because TravisCI integration is working, and I'll open a ticket for the bug. (I think it's a race, which is why it worked before.)

PR #31, opened 2015-12-04
Add support for TravisCI

justenwalker:
Adds .travis.yml and alters the makefile to support building on TravisCI (#30)

### DIFF
--- a/makefile
+++ b/makefile
@@ -54,6 +54,7 @@ ifeq "$(TRAVIS)" "true"
 build: .godeps
 	mkdir -p ${ROOT}/build
 	export GOPATH=${ROOT}/.godeps:${ROOT}/src && \
+	export CGO_ENABLED=0 && \
 		cd ${ROOT}/src/containerbuddy && \
 		go build -a -o ${ROOT}/build/containerbuddy -ldflags ${LDFLAGS}
 	chmod +x ${ROOT}/build/containerbuddy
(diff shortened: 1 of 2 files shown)


justenwalker:
Example build: https://travis-ci.org/justenwalker/containerbuddy/builds/94940405

makefile
```diff
@@ -47,6 +49,15 @@ consul:
 # build and release
 
 # build our binary in a container
+
+ifeq "$(TRAVIS)" "true"
```
tgross: What's the reason for this conditional?

justenwalker: In my testing, there was a permissions problem with chmod after it builds the binary with docker.

Also, I was attempting to use the native go binary on the travis CI machine, since it allows us to build it against different go versions

tgross: Sounds reasonable. But we'll need to put `CGO_ENABLED=0` back. It makes a difference with portability between glibc and musl libc:

without CGO_ENABLED=0:

```sh
$ ldd containerbuddy
        linux-vdso.so.1 (0x00007fff711bf000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f5f33092000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5f32ce9000)
        /lib64/ld-linux-x86-64.so.2 (0x000055f1a2073000)
```

with CGO_ENABLED=0:

```sh
$ ldd containerbuddy
        not a dynamic executable
```

justenwalker: Added back in 474f3fb

tgross: Awesome. Merging!

PR #33, opened 2015-12-04
Signal test fix

tgross:
For https://github.com/joyent/containerbuddy/issues/32

Signals are asynchronous and we needed to force the test runner to wait long enough for the signal to propagate in order to guarantee that the test would pass on all platforms. In the process of debugging this, I've refactored the signal setup process so as to reduce the confusion around set up and teardown of the signal handlers.

ref https://travis-ci.org/tgross/containerbuddy/builds/94976042 for a passing build
cc @justenwalker and @misterbisson

### DIFF
--- a/src/containerbuddy/signals.go
+++ b/src/containerbuddy/signals.go
@@ -74,17 +74,15 @@ func handleSignals(config *Config) {
 			case syscall.SIGUSR1:
 				toggleMaintenanceMode()
 				if inMaintenanceMode() {
-					log.Println("we are paused!")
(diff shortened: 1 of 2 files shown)


justenwalker:
Check out this:

https://github.com/justenwalker/containerbuddy/commit/50b761d55a2c08b18ca7c9b0dde8a67088d95bbc

I was working on it in parallel.  I'm pretty sure we came to the same conclusion though.

I did add a way to re-register handlers more easily.

tgross:
ref https://github.com/joyent/containerbuddy/issues/32#issuecomment-162211064

Issue #28, opened 2015-12-04
Add more custom script hooks


justenwalker:
From @misterbisson:

> #22 opens up the door to more than just deregistering services. Consider running Couchbase in Docker. The blueprint and demo makes deploying Couchbase and scaling it up easy (repo), but scaling down requires more steps that haven't been automated.

> A `SIGTERM` handler could be exactly what's needed to add that automation. If it could also execute a user-defined executable (and wait for it), it would allow us to mark the Couchbase node for removal from the cluster and automatically rebalance the data to the remaining nodes before stopping it.

> I haven't tested it, but I think the right command to call would be:

> `couchbase-cli rebalance -c 127.0.0.1:8091 -u $COUCHBASE_USER -p $COUCHBASE_PASS --server-remove=${IP_PRIVATE}:8091`

> And when that is done, it should be safe to stop (and remove/delete) the container.

misterbisson:
From @justenwalker in https://github.com/joyent/containerbuddy/pull/22#issuecomment-161963349:

> In my mind, OnStop makes more sense to happen after the application exits rather than before - so that we know the service isn't interacting with anything before we send cleanup commands (like Finally block)

That raises questions about the inconsistency with of the language compared to `onStart`. `onStart` is actually executed before the application starts. The following is a crappy solution, in my mind, but it avoids changing any existing code:

- `onStart` before application start
- `onStarted` after application start
- `onStop` before application stop
- `onStopped` _after_ application stop

What we really want, perhaps, is:

- `preStart` before application start, the current behavior of `onStart`
- `postStart` after application start
- `preStop` before application stop
- `postStopped` _after_ application stop

misterbisson:
@justenwalker the title of the ticket may be too specific given our similar, but slightly different interests. Would you be okay with retitling it in a way that fits the conversation about pre and post stopped execution?

tgross:
> `onStarted` after application start

This is going to be difficult to actually implement because we block execution flow on the application. Even if we could figure out a way around that, the application can "start" (as in fork) but not actually be "running" for any useful purpose. e.g. slow Java startup times.

> - `onStop` before application stop

Services deregister and then stop. Does this happen before or after the service is deregistered?

justenwalker:
So do we need something like:

- `onStart` - Unchanged
- `preStop` - Runs before Containerbuddy terminates the application - blocks
- `postStop` - Runs after the application terminates

tgross:
> It's not symmetric

[...]

PR #34, opened 2015-12-05
Add preStop and postStop hooks

justenwalker:
- Refactor hook calling into runCommand method
- Add docs and example for preStop and postStop
- Add runCommand tests

See #28

### DIFF
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -20,8 +20,9 @@ var (
 type Config struct {
 	Consul       string `json:"consul,omitempty"`
 	OnStart      string `json:"onStart"`
+	PreStop      string `json:"preStop"`
+	PostStop     string `json:"postStop"`
 	StopTimeout  int    `json:"stopTimeout"`
-	onStartArgs  []string
 	Command      *exec.Cmd
 	QuitChannels []chan bool
 	Services     []*ServiceConfig `json:"services"`
@@ -37,7 +38,6 @@ type ServiceConfig struct {
 	TTL              int      `json:"ttl"`
 	Interfaces       []string `json:"interfaces"`
 	discoveryService DiscoveryService
-	healthArgs       []string
 	ipAddress        string
 }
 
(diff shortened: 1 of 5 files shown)


justenwalker:
Build fails because of signal race condition - relevant tests do pass though:

```
=== RUN   TestRunCommand
Running doStuff with args: --debug
Running failStuff with args: --debug
2015/12/05 04:08:44 exit status 255
--- PASS: TestRunCommand (0.00s)
```

justenwalker:
Rebased on the #36 fix so the tests would pass and to incorporate the refactoring.

src/containerbuddy/config.go
```diff
@@ -20,8 +20,12 @@ var (
 type Config struct {
 	Consul       string `json:"consul,omitempty"`
 	OnStart      string `json:"onStart"`
+	PreStop      string `json:"preStop"`
+	PostStop     string `json:"postStop"`
 	StopTimeout  int    `json:"stopTimeout"`
 	onStartArgs  []string
+	preStopArgs  []string
+	postStopArgs []string
```
tgross: I did a search thru the code and it looks like `onStartArgs`, `preStopArgs`, and `postStopArgs` are never used. You're passing the equivalent public member of the struct to `runCommand` instead, so there's no reason to have these.

src/containerbuddy/main.go
```diff
@@ -111,6 +112,18 @@ func getCmd(args []string) *exec.Cmd {
 	}
 }
 
+func runCommand(commandString string) (int, error) {
```
tgross: I feel like it makes more sense to move the string splitting into the `run`, and then we can eliminate `ServiceConfig.healthArgs` and `Backend.onChangeArgs` members too (and the splitting code in config.go).

tgross:
cc'ing in @misterbisson here too.

justenwalker:
@tgross incorporated the changes you made in the last update - rebased on latest master.

misterbisson:
lgtm

tgross:
Apparently TravisCI is having some build-time issues. We'll wait for them to catch up and then merge this once we get the +1 from them.

justenwalker:
@tgross - I kicked the build with another rebase and Travis finally tested it

tgross:
Thanks @justenwalker. Merging!

Issue #37, opened 2015-12-08
Consider supporting command args with JSON Array


justenwalker:
All commands in the app.json are just strings.  Currently, Containerbuddy splits these strings on space.
This could be fragile if the command requires an argument to have a space.

This leaves three options:

1. Spaces in arguments are not supported - rewrite your app please
2. Make the arg splitter more intelligent (i.e. support quoting/escaping arguments)
3. Allow the user to enter a JSON array of arguments

## Proposal

Supporting a JSON array of arguments makes practical sense:

1. Option 1 makes the application less usable
2. Option 2 can be complicated to implement, and may introduce command parsing bugs
3. JSON Arrays are easy enough to add, and it uses the JSON syntax for argument splitting, which is already well-defined and implemented.

#### Commands & arguments

All executable fields, such as `onStart` and `onChange`, accept both a string or an array. If a string is given, the command and its arguments are separated by spaces; otherwise, the first element of the array is the command path, and the rest are its arguments.

**String Command**

```json
"health": "/usr/bin/curl --fail -s http://localhost/app"
```

**Array Command**

```json
"health": [
  "/usr/bin/curl",
  "--fail",
  "-s",
  "http://localhost/app"
]
```

tgross:
> Both forms can be supported at the same time:

I like where you're going with the overall concept here but I think supporting two different field names is inelegant. I'd love if we could do something like what the Docker folks did with the `CMD` and `ENTRYPOINT` Dockerfile stanzas, where they were able to parse both `"/usr/bin/my/health"` and `["/usr/bin/my/health", "--with", "some", "-args"]`.

The typical way to do this would be to deserialize the JSON into `map[string][]json.RawMessage`, parse the pieces we need, and then complete serialization of the nested `ServiceConfig` and `BackendConfig` structs (ref http://mattyjwilliams.blogspot.com/2013/01/using-go-to-unmarshal-json-lists-with.html, in the "option C" for an example of how this could be done). This means the code that parses the configuration will be less efficient but honestly we only parse it once so if it's mildly hairy that's fine as long as it's well-tested.

If we *weren't* going to do that, then I'd say we just break backwards compat (this is still marked alpha) and use a JSON array that will sometimes be a one-argument array. But if we can make the former work that'd be ideal.

tgross:
[...]

PR #38, opened 2015-12-08
Add support for command args with JSON Array

justenwalker:
See [README.md](https://github.com/justenwalker/containerbuddy/blob/array-args/README.md) for a [better description](https://github.com/justenwalker/containerbuddy/blob/array-args/README.md#commands--arguments)

- Refactor getCmd into argsToCmd and strToCmd
- Add *Cmd parameters to all config structs
- Add parseCommandArgs function which supports JSON array or JSON string
- Support parsing interfaces as a string with parseInterfaces
- Do parsing of all commands once - during loadConfig()
- Update README

For #37

### DIFF
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -281,3 +281,21 @@ func getInterfaceIps() []InterfaceIp {
 	}
 	return ifaceIps
 }
+
+func argsToCmd(args []string) *exec.Cmd {
+	if len(args) == 0 {
+		return nil
+	}
+	if len(args) > 1 {
+		return exec.Command(args[0], args[1:]...)
+	} else {
+		return exec.Command(args[0])
+	}
+}
+
+func strToCmd(command string) *exec.Cmd {
+	if command != "" {
+		return argsToCmd(strings.Split(command, " "))
+	}
+	return nil
+}
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -87,7 +87,7 @@ const (
 
 func parseInterfaces(raw json.RawMessage) ([]string, error) {
 	if raw == nil {
-		return nil, nil
+		return []string{}, nil
 	}
 	// Parse as a string
 	var jsonString string
@@ -100,7 +100,7 @@ func parseInterfaces(raw json.RawMessage) ([]string, error) {
 		return jsonArray, nil
 	}
 
-	return nil, errors.New("interfaces must be a string or an array")
+	return []string{}, errors.New("interfaces must be a string or an array")
 }
 
 func parseCommandArgs(raw json.RawMessage) (*exec.Cmd, error) {
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -144,19 +144,19 @@ func loadConfig() (*Config, error) {
 
 	onStartCmd, err := parseCommandArgs(config.OnStart)
 	if err != nil {
-		return nil, errors.New(fmt.Sprintf("Could not parse `onStart`: %s", err.Error()))
+		return nil, fmt.Errorf("Could not parse `onStart`: %s", err)
 	}
 	config.onStartCmd = onStartCmd
 
 	preStopCmd, err := parseCommandArgs(config.PreStop)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


src/containerbuddy/main.go
```diff
@@ -88,23 +88,26 @@ func poll(config Pollable, fn pollingFunc, command string) chan bool {
...
+func checkForChanges(pollable Pollable) {
 	backend := pollable.(*BackendConfig) // if we pass a bad type here we crash intentionally
 	if backend.CheckForUpstreamChanges() {
-		run(command)
+		run(backend.onChangeCmd)
 	}
 }
 
-func getCmd(args []string) *exec.Cmd {
+func argsToCmd(args []string) *exec.Cmd {
```
tgross: I feel like `argsToCmd` and `strToCmd` should be moved into config.go now, just because all but one of their callers are in config.go.

src/containerbuddy/config_test.go
```diff
@@ -52,8 +59,87 @@ func TestValidConfigParse(t *testing.T) {
...
-	if config.OnStart != "/bin/to/onStart.sh" {
-		t.Errorf("onStart not configured")
+
+	validateCommandParsed(t, "onStart", config.onStartCmd, []string{"/bin/to/onStart.sh", "arg1", "arg2"})
+	validateCommandParsed(t, "preStop", config.preStopCmd, []string{"/bin/to/preStop.sh", "arg1", "arg2"})
+	validateCommandParsed(t, "postStop", config.postStopCmd, []string{"/bin/to/postStop.sh"})
+	validateCommandParsed(t, "health", config.Services[0].healthCheckCmd, []string{"/bin/to/healthcheck/for/service/A.sh"})
+	validateCommandParsed(t, "health", config.Services[1].healthCheckCmd, []string{"/bin/to/healthcheck/for/service/B.sh"})
+	validateCommandParsed(t, "onChange", config.Backends[0].onChangeCmd, []string{"/bin/to/onChangeEvent/for/upstream/A.sh"})
+	validateCommandParsed(t, "onChange", config.Backends[1].onChangeCmd, []string{"/bin/to/onChangeEvent/for/upstream/B.sh"})
```
tgross: I :heart: the table-style tests like this.

src/containerbuddy/config.go
```diff
@@ -102,6 +142,24 @@ func loadConfig() (*Config, error) {
 		return nil, err
 	}
 
+	onStartCmd, err := parseCommandArgs(config.OnStart)
+	if err != nil {
+		return nil, errors.New(fmt.Sprintf("Could not parse `onStart`: %s", err.Error()))
```
tgross: Just from the standpoint of verbosity: you don't need to call the `.Error()` method on these. Errors just wrap strings in golang so they have the interfaces to be `%s`-formatted. In fact, as I just discovered when looking up something to give you as a reference, this particular construct is supported in the stdlib (ref https://golang.org/pkg/errors/#example_New_errorf). So try:

```go
if err != nil {
    return nil, fmt.Errorf("Could not parse `onStart`: %s", err)
}
```

src/containerbuddy/config.go
```diff
@@ -80,6 +85,41 @@ const (
 	defaultStopTimeout int = 5
 )
 
+func parseInterfaces(raw json.RawMessage) ([]string, error) {
+	if raw == nil {
+		return nil, nil
```
tgross: Do we want to return `nil, nil` here or should we be returning an empty array? I feel like the latter is going to be less likely to result in a null-pointer exception down the road.

src/containerbuddy/config.go
```diff
@@ -123,6 +181,11 @@ func loadConfig() (*Config, error) {
 	}
 
 	for _, backend := range config.Backends {
+		cmd, err := parseCommandArgs(backend.OnChangeExec)
+		if err != nil {
+			return nil, errors.New(fmt.Sprintf("Could not parse `onChange` in backend %s: %s", backend.Name, err.Error()))
```
tgross: Break up long lines please (plus this can be shorted once we use `fmt.Errorf`)

tgross:
I like the approach here. Just a few comments to fix I think.

justenwalker:
@tgross I agree with and have addressed your comments

tgross:
LGTM! Merging!

PR #41, opened 2015-12-11
Refresh exec.Cmd for Health and OnChange

### LINKED ISSUE #40: Polling Bug - exec.Cmd is not reusable
Unfortunately, I did not do adequate testing for #38 and introduced bug.

Parsing the config yields a Cmd object for all commands. However, health checks and backend change hooks are run more than once. Once run, the Cmd object is effectively dead - cannot be run again.

The result is that we hit the failure case in the executeAndWait function and exit prematurely.

This does not affect the preStop, postStop, and onStart Commands since they are only ever run once.

justenwalker:
exec.Cmd can only be run once. Introduce two new methods, one on
ServiceConfig and one on BackendConfig that encapsulates running the
command and creating a new exec.Cmd object for subsequent runs.

Fixes #40

### DIFF
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -65,6 +65,13 @@ func (b *BackendConfig) CheckForUpstreamChanges() bool {
 	return b.discoveryService.CheckForUpstreamChanges(b)
 }
 
+func (b *BackendConfig) OnChange() (int, error) {
+	exitCode, err := run(b.onChangeCmd)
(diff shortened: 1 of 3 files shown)


tgross:
Ok, we didn't catch this in test (and it's definitely an easy bug to introduce and miss in review), so we should make sure we've added tests to exercise the bug in this PR.

tgross:
LGTM. Thanks for jumping on this quickly, @justenwalker!

Issue #16, opened 2015-11-18
support environment interpolation in the config file


bkcsfi:
It would be nice to be able to reference environment variables within the configuration file. e.g. if I want to one Dockerfile that can be used on a 'dev consul' and later in production, I might choose to change (future) tag support or use a different consul master port.

tgross:
I like this idea.

From an implementation standpoint we'd probably want to walk the [Config struct](https://github.com/joyent/containerbuddy/blob/101a5a857ef2a508b0cf29504735c8a0bfaaa5e5/src/containerbuddy/config.go#L14-L40) after we've [parsed it](https://github.com/joyent/containerbuddy/blob/101a5a857ef2a508b0cf29504735c8a0bfaaa5e5/src/containerbuddy/config.go#L120-L123). For each string in the config, we'd need to:

- parse out "things that look like environment variables"
- ask `os.Getenv` for their values
- inject the new values

Complications from a design standpoint are:

- Go's `os.Getenv` doesn't make any distinction between if the environment doesn't have a value, and if it was never set in the first place. If `os.Getenv` returns an empty string, do we not replace the variable with empty string or do we not do the replacement?
- How to we pass environment variable literals to the underlying applications? Ex. if our user-defined health check is `curl -s --fail localhost:8000/api/health?${SOME_CONFIG_VALUE}` then it looks like we'll need a way to escape those values.

Seems doable though. I'm going to mark this as an open enhancement where we'd encourage a PR.

That being said and just in case it's not obvious, we do currently support passing the entire Containerbuddy configuration in as an environment variable (ref https://github.com/joyent/containerbuddy/blob/master/src/containerbuddy/config.go#L65-L69). You can pass in either a config file path or JSON (as I've done in this [other work-in-progress project](https://github.com/tgross/triton-nginx/blob/6d9c97b3022289db1613ea97e3b2e5bd6f3635c7/example/start.sh#L51)). I realize that's not quite what you're asking for here but there's a bit of runtime configuration available in that way.

bkcsfi:
instead of adding support for environment interpolation directly in containerbuddy, could you add a command line option something like --configfromexec  "some_program_path some_args more_args"

where some_program_path is a program in the container that will print to stdout the configuration file

if some_program_path returns non-zero rc you can exit the container with that error.. also if some_program_path prints to stderr that could go to the container log.

[...]

PR #46, opened 2015-12-14
Add support for templates in config

justenwalker:
This is my first pass at supporting templates in configurations:

- Config is parsed as a Go template before being unmarshaled from JSON.
- Environment variables are available for interpolation. ie: `{{ .ENV_VAR_NAME }}`

For #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ The format of the JSON file configuration is as follows:
 ```json
 {
   "consul": "consul:8500",
-  "onStart": "/opt/containerbuddy/onStart-script.sh",
+  "onStart": "/opt/containerbuddy/onStart-script.sh {{.ENV_VAR_NAME}}",
   "stopTimeout": 5,
   "preStop": "/opt/containerbuddy/preStop-script.sh",
   "postStop": "/opt/containerbuddy/postStop-script.sh",
@@ -124,6 +124,22 @@ All executable fields, such as `onStart` and `onChange`, accept both a string or
   "-s",
   "http://localhost/app"
 ]
+```
+
+### Template Configuration
+
+Containerbuddy configuration has template support, such as substituting environment variables in the config file. This makes it possible to use the same configuration and alter the behavior per environment. Full documentation about the template language used can be found in the [Go text/template Docs](https://golang.org/pkg/text/template/).
+
+**Example**
+
+```json
+{
+  "consul": "consul:8500",
+  "onStart": "/opt/containerbuddy/onStart-script.sh {{.URL_TO_SERVICE}} {{.API_KEY}}",
+
+  ...
+
+}
 ```
 
 ### Operating Containerbuddy
--- a/README.md
+++ b/README.md
@@ -128,19 +128,18 @@ All executable fields, such as `onStart` and `onChange`, accept both a string or
 
 ### Template Configuration
 
-Containerbuddy configuration has template support, such as substituting environment variables in the config file. This makes it possible to use the same configuration and alter the behavior per environment. Full documentation about the template language used can be found in the [Go text/template Docs](https://golang.org/pkg/text/template/).
+Containerbuddy configuration has template support. If you have an environment variable such as `FOO=BAR` then you can use `{{.FOO}}` in your configuration file and it will be substituted with `BAR`.
 
-**Example**
+**Example Usage**
 
 ```json
 {
   "consul": "consul:8500",
   "onStart": "/opt/containerbuddy/onStart-script.sh {{.URL_TO_SERVICE}} {{.API_KEY}}",
-
-  ...
-
 }
 ```
+
+_Note:  If you need more than just variable interpolation, check out the [Go text/template Docs](https://golang.org/pkg/text/template/)._
 
 ### Operating Containerbuddy
 
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -303,6 +303,11 @@ func unmarshalConfig(data []byte) (*Config, error) {
 	return config, nil
 }
 
+func newJSONParseError(js []byte, syntax *json.SyntaxError) error {
+	line, col, err := highlightError(js, syntax.Offset)
+	return fmt.Errorf("Parse error at line:col [%d:%d]: %s\n%s", line, col, syntax, err)
+}
+
 func highlightError(data []byte, pos int64) (int, int, string) {
 	prevLine := ""
 	thisLine := ""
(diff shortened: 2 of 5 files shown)


tgross:
There's some behavior in failure modes that seems to undo some of the UX improvements that we had in https://github.com/joyent/containerbuddy/pull/45. If I change the example application so that the Nginx port looks like this:

```
{
  "consul": "consul:8500",
  "services": [
    {
      "name": "nginx",
      "port": {{ .NGINX_PORT }},
      "interfaces": ["eth0"],
      "health": "/usr/bin/curl --fail -s http://localhost/health.txt",
      "poll": 10,
      "ttl": 25
    }
...
```

And then don't add `NGINX_PORT=80` to my `docker-compose-local.yml` I get this somewhat unintuitive-looking error:

> 2015/12/14 18:33:48 Could not parse configuration: invalid character ',' looking for beginning of value

It works just fine if I do pass the variable. But the failure mode is cryptic.

README.md
```diff
@@ -126,6 +126,22 @@ All executable fields, such as `onStart` and `onChange`, accept both a string or
 ]
 ```
 
+### Template Configuration
+
+Containerbuddy configuration has template support, such as substituting environment variables in the config file. This makes it possible to use the same configuration and alter the behavior per environment. Full documentation about the template language used can be found in the [Go text/template Docs](https://golang.org/pkg/text/template/).
```
tgross: I feel like we should be a bit more straightforward about how environment variables are interpreted. The `text/template` docs are accurate but are dense and talk about "fields" (which is an implementation detail) but what we're really trying to say is "if you set an environment variable `FOO=BAR` then you can read it into the template like `{{ .FOO }}`". We should make that common case front-and-center of this section and then follow up with something along the lines of "oh and this is Go `text/template` language if you want more".

justenwalker:
To your comment about the cryptic error message.  I'm not sure how to resolve this, since it seems to happen inside the `json.Unmarshal` - where the checks happen after those values have already been parsed.

Any ideas on where I should look to resolve this?

tgross:
> I'm not sure how to resolve this, since it seems to happen inside the json.Unmarshal
> The UX checks happen after those values have already been parsed.

Hm. Maybe we should be using `missingkey=error` in that case? That way we get an error at the point it happens and can give the user a sensible error message about there being a missing value?

justenwalker:
I'm on board with the idea of making it an explicit error. 

I liked the blank value first because it closely mirrored what I expect to happen in bash scripts. If we leave it as `missingkey=zero` it will allow people who are using containerbuddy to have optional environment variables in their config more easily.

misterbisson:
> I liked the blank value first because it closely mirrored what I expect to happen in bash scripts. If we leave it as missingkey=zero it will allow people who are using containerbuddy to have optional environment variables in their config more easily.

I think we can demo some `Dockerfile` patterns to go with this that include default values for environment vars, see [example](https://github.com/joyent/mesos-dockerfiles/blob/master/agent/Dockerfile#L28-L31).

tgross:
> If we leave it as missingkey=zero it will allow people who are using containerbuddy to have optional environment variables in their config more easily.

I think unfortunately that's only going to work for string variables, because they get quoted and so the JSON parser has at the very least `""` to work with.

justenwalker:
In terms of environment variables - they are always strings, no?

tgross:
> In terms of environment variables - they are always strings, no?

Right, but they aren't strings in the JSON, which appears to be the problem here. The zero-value for the environment variable is indeed `""`, but that results in the JSON:

```
{
  "consul": "consul:8500",
  "services": [
    {
      "name": "nginx",
      "port": ,
      "interfaces": ["eth0"],
      "health": "/usr/bin/curl --fail -s http://localhost/health.txt",
      "poll": 10,
      "ttl": 25
    }
...
```

tgross:
Sorry... missed this comment prior to my one above somehow.

> Perhaps we can add a default function that can be used to provide sane defaults to numeric values?

Strictly speaking we can do this already with `{{if .NGINX_PORT }}{{ .NGINX_PORT }}{{else}}80{{end}}` but I in keeping with the "this should be easy to understand" theme, I like the idea of having a pipeline function that does it better.

justenwalker:
@tgross Still wont solve the issue entirely though, since if the environment variable is omitted - it will fail with this cryptic error. Maybe we can parse out the syntax error message and make some more meaningful error in this case (see https://groups.google.com/forum/#!topic/golang-nuts/fizimmXtVfc)

misterbisson:
Would it be possible to pre-parse the config file to find the referenced env vars and test their availability?

tgross:
@misterbisson that's what I proposed [here](https://github.com/joyent/containerbuddy/pull/46#issuecomment-164531709)

tgross:
> Still won't solve the issue entirely though, since if the environment variable is omitted - it will fail with this cryptic error

Sure, that's true. We're stuck with the possibility of misconfiguration regardless of what we do, and with @misterbisson's suggestion about Dockerfile defaults and the existing template language options there are options for the user to reduce errors as it is. So doing our darndest to return a sensible error message seems like the priority.

misterbisson:
> that's essentially what I proposed

Sorry, I didn't understand [your comment](https://github.com/joyent/containerbuddy/pull/46#issuecomment-164531709) in that context, thanks for clarifying @tgross.

justenwalker:
I pushed up a change that should:

1. Allow default values with pipeline {{ .ENV_VAR | default "hello!" }}
2. Give us better error messages on json parse errors:

```
2015/12/14 21:48:44 Parse error on line 6: invalid character ',' looking for beginning of value
      "port": ,
              ^
```

Line number code based off that google groups discussion that I mentioned above.

justenwalker:
Updated the implementation of the JSON error pretty-printer

Implementation is a little easier to follow and maintain with Scanners. 
Plus I added some additional context to the error message:


**Examples**

```
Parse error at line:col [1:1]: invalid character '<' looking for beginning of value
    1: <>
-------^

Parse error at line:col [2:13]: invalid character ',' looking for beginning of value
    1: {
    2:     "test": ,
-------------------^

Parse error at line:col [5:5]: invalid character 't' looking for beginning of object key string
    4:     "test3": false,
    5:     test2: "hello"
-----------^
```

out.txt
```diff
@@ -0,0 +1,75 @@
+docker rm -f containerbuddy_consul || true
```
tgross: was this file committed accidentally?

justenwalker: yes - removed

src/containerbuddy/config.go
```diff
@@ -271,25 +273,69 @@ func parseConfig(configFlag string) (*Config, error) {
...
+			highlight = fmt.Sprintf("%s^", strings.Repeat("-", int(7+col-1)))
+			break
+		}
+		col -= readBytes + 1
+		line++
+	}
+	return line, int(col), fmt.Sprintf("%s%s%s", prevLine, thisLine, highlight)
+}
+
+func newJSONParseError(js []byte, syntax *json.SyntaxError) error {
```
tgross: This is just me being picky, but can we move this function above `highlightError`? The only call chain for this is `unmarshalConfig --> newJSONParseError --> highlightError` so legibility is helped if you don't have to jump down and then back up again.

justenwalker: Done

tgross:
Nice work. LGTM.

PR #48, opened 2015-12-14
Support configuration reload

tgross:
@misterbisson @justenwalker this PR supports reloading the configuration (ref https://github.com/joyent/containerbuddy/issues/9). This is not a completed implementation, but I figure getting in front of you guys would be a good idea. There will also almost certainly be merge conflicts with https://github.com/joyent/containerbuddy/pull/46 so I'll be sure to clean those up first.

We accept SIGHUP to reload the configuration. This stops all polling and forces all advertised services to be deregistered (and then reloaded) in the discovery service. We have to do this because otherwise we can't actually update service configurations in Consul.

~~Right now the fairly large hole in the implementation that I'm planning on taking on starting tomorrow morning is that we don't properly cover the `preStop` and `postStop` functions. We pass their config struct in `main`, which is outside the loop for getting the config. It's just a bit of refactoring so we can stick the config in at the appropriate moment.~~ *Done*

### DIFF
--- a/src/containerbuddy/main.go
+++ b/src/containerbuddy/main.go
@@ -21,19 +21,9 @@ func main() {
 		os.Exit(onStartCode)
 	}
 
-	// Set up signal handler for placing instance into maintenance mode
+	// Set up handlers for polling and to accept signal interrupts
 	handleSignals(config)
-
-	// Set up polling operations
 	handlePolling(config)
-
-	// gracefully clean up so that our docker logs aren't cluttered after an exit 0
-	// TODO: do we really need this?
-	defer func() {
-		for _, ch := range config.QuitChannels {
-			close(ch)
-		}
-	}()
 
 	if len(flag.Args()) != 0 {
 		// Run our main application and capture its stdout/stderr.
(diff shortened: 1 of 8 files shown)


misterbisson:
@tgross looks solid to me

src/containerbuddy/main.go
```diff
@@ -21,25 +21,9 @@ func main() {
...
-	for _, backend := range config.Backends {
-		quit = append(quit, poll(backend, checkForChanges))
-	}
-	for _, service := range config.Services {
-		quit = append(quit, poll(service, checkHealth))
-	}
-	config.QuitChannels = quit
-
-	// gracefully clean up so that our docker logs aren't cluttered after an exit 0
-	// TODO: do we really need this?
```
tgross: Tested this and it turns out the answer is no, so I cut it.

tgross:
@misterbisson and @justenwalker I've got this wrapped up. I've added a global (protected by a mutex) that stores the configuration so that top-level functions can get the new config at the time they're called, rather than at the time their caller is called. I'm not super-wild about how whether we're calling `getConfig` or just taking a pointer to the config as context are dependent on where they're being called just in terms of consistency. But keeping the global state out of almost all the functions does make this a lot more testable... I'm definitely open to suggestion as far as improvements.

My integration test (which I definitely do want to get added to our automation at some point as per https://github.com/joyent/containerbuddy/issues/42):

```shell
$ make run
...
$ docker exec -it example_nginx_1 /bin/bash
root@17f38ba6e2c6:/# sed -i 's/80/8080/' /opt/containerbuddy/nginx.json
root@17f38ba6e2c6:/# kill -HUP 1
root@17f38ba6e2c6:/# exit
```

And then it's easy to verify by watching what Consul is doing.

misterbisson:
:house_with_garden: :walking:

PR #55, opened 2015-12-26
Reformat code to pass golint checks

justenwalker:
This PR fixes any warnings that [golint](https://github.com/golang/lint) prints when run on the containerbuddy sources.  

**This does have a good chance of causing merge errors with in-progress branches.**

## Changes

- Fix identifier name casing
- Add doc comments to exported objects
- Refactor some code to satisfy golint style suggestions (else blocks, post-increment)

## How to lint the containerbuddy sources

```
$ go get -u github.com/golang/lint/golint
$ $GOPATH/bin/golint -min_confidence 0.8 src/containerbuddy
```

*Note:* **golint** is not a panacea for style errors, but it's a good place to start catching them.  It makes it easier to be idiomatic & consistent with the styles described in [Effective Go](https://golang.org/doc/effective_go.html). If it is not too onerous, it could also help remind us to include documentation (although I do find it easy to write terse 1-liners just to make the warning go away)

### DIFF
--- a/src/containerbuddy/signals.go
+++ b/src/containerbuddy/signals.go
@@ -68,8 +68,8 @@ func terminate(config *Config) {
 func reloadConfig(config *Config) *Config {
 	signalLock.Lock()
 	defer signalLock.Unlock()
-	newConfig, err := loadConfig()
 	log.Printf("Reloading configuration.\n")
+	newConfig, err := loadConfig()
 	if err != nil {
 		log.Printf("Could not reload config: %v\n", err)
 		return nil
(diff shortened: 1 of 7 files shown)


src/containerbuddy/signals.go
```diff
@@ -68,26 +68,25 @@ func terminate(config *Config) {
 func reloadConfig(config *Config) *Config {
 	signalLock.Lock()
 	defer signalLock.Unlock()
-
+	newConfig, err := loadConfig()
```
tgross: This moves the log statement to after the attempt to reload the config, whereas I wanted it beforehand so as to signal that's what was happening should there be a problem.

tgross:
This LGTM except for the one minor comment I made here https://github.com/joyent/containerbuddy/pull/55#discussion_r48478339

justenwalker:
Rebased on master

tgross:
Awesome, thanks @justenwalker

PR #57, opened 2015-12-26
Factor the go make stuff into a docker makefile

justenwalker:
I had this idea to factor out the build logic into a docker-specific makefile.  IMO it makes the build tasks easier to maintain. Does this seem like a good idea?

## Changes

- Factor build / test logic into Makefile.docker
- Create Dockerfile for creating the build environment
- Make coverage generate html file instead of opening a browser

The main `makefile`:

1. Create a build environment: `containerbuddy_build`
2. Set up middleware for integration tests: `containerbuddy_consul`
3. All the release logic for post-build

The new `Makefile.docker` contains all of the targets for building, testing, format/linting, coverage, etc...
It is able to assume it runs in a consistent and initialized build environment - so there doesn't need to be any setup logic inside of it.

### Modified makefile targets

- `make clean` - Cleans up build artifacts and removes build environment docker images
- `make build` - **Default target**, builds containerbuddy
- `make test` - Runs unit tests
- `make cover` - Generates coverage report HTML
- `make lint` - Runs go vet, go fmt, and golint

### DIFF
--- a/makefile
+++ b/makefile
@@ -50,7 +50,7 @@ test: docker cover/coverage.out
 cover/coverage.out:
 	${DOCKERMAKE} test
 
-cover: cover/coverage.html
+cover: docker cover/coverage.html
 
 cover/coverage.html:
 	${DOCKERMAKE} cover
--- a/makefile
+++ b/makefile
@@ -29,7 +29,7 @@ clean:
 
 build/containerbuddy_build:
 	mkdir -p ${ROOT}/build
-	docker rmi -f containerbuddy_build || true
+	docker rmi -f containerbuddy_build > /dev/null 2>&1 || true
 	docker build -t containerbuddy_build ${ROOT}
 	docker inspect -f "{{ .ID }}" containerbuddy_build > build/containerbuddy_build
 
--- a/makefile
+++ b/makefile
@@ -3,75 +3,64 @@ SHELL := /bin/bash
 .SHELLFLAGS := -eu -o pipefail
 .DEFAULT_GOAL := build
 
-.PHONY: clean test consul run example ship
+.PHONY: clean test consul run example ship dockerfile docker cover lint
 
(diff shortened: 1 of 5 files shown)


makefile
```diff
@@ -3,75 +3,64 @@ SHELL := /bin/bash
...
 	docker run -d -m 256m --name containerbuddy_consul \
 		progrium/consul:latest -server -bootstrap-expect 1 -ui-dir /ui
 
-
-# ----------------------------------------------
-# build and release
-
-# build our binary in a container
-
-ifeq "$(TRAVIS)" "true"
```
tgross: I think we can now eliminate this section of the `.travis.yml` file as well? https://github.com/joyent/containerbuddy/blob/69ad0afd0fefc830da47225f6ccbe290c0b869ab/.travis.yml#L4-L6

makefile
```diff
@@ -3,75 +3,64 @@ SHELL := /bin/bash
...
-	docker rm -f containerbuddy_consul || true
+test: docker cover/coverage.out
 
-cover: test
-	@sed -i 's/_\/root\/src\///' coverage.out
-	go tool cover -html=coverage.out
+cover/coverage.out:
+	${DOCKERMAKE} test
 
+cover: cover/coverage.html
```
tgross: Cover should depend on the `docker` target as well.

tgross:
The thing I really like about this approach is that it gets rid of TravisCI-specific code in the makefile and that completely Dockerizing the build environment will eliminate a lot of support headaches. Added a couple of comments.

makefile
justenwalker: @tgross Removed. Also made travis do lint as well:

`make clean lint build test`

justenwalker:
Rebased on master

tgross:
Nice work. Thanks again, @justenwalker

Issue #53, opened 2015-12-19
Support selecting service IP matching CIDR


justenwalker:
If Triton allows multiple IPs per address (@bahamat on #50), and we potentially have IPV6 (#52) Then perhaps we need to allow services to select their IP in a more flexible way.

Propose something like:

```json
{
  "cidr": "192.168.0.0/16"
}
```

instead of (or in addition to) the `"interfaces"` option

Proposed behavior:
1. Both not given - pick the *first IP* on eth0 (sane default) - *Already Supported*
2. `interfaces` given - pick the *first IP* on the first found interface listed in `interfaces` - *Already Supported*
3. `cidr` given - pick the *first IP* address on any interface matching the CIDR
4. Both given - pick the *first IP* address on any interface (limited to `interfaces`) matching the CIDR

*Note*: *first IP* could be IPV4 or IPV6 - Maybe that's a flag or environment variable that dictates which one is preferred? Depends on what is decided in #52 I suppose. In the `cidr` case, it will always be determined by the whatever version of CIDR is given, so this disambiguation logic applies to 1 and 2

Also, case 4 may be unnecessary if each interface will only have at most 1 v4 and 1 v6 address. In this case it should probably be a parse error instead of handling it specially.

bahamat:
I did say "*Triton* supports", however that's inaccurate (I've edited my comment in #50 to reflect this). SmartOS allows for multiple addresses. This isn't exposed in Triton (JPC) yet.

However, IPv6 is upon us. Recent changes in SmartOS and SDC are leading up to making IPv6 work across the entire stack. When it's available, there may be multiple addresses on any one interface. Most likely in Triton/JPC there will be one IPv4 and one IPv6 address per `nic` object.

justenwalker:
If it's only 1 IPv4 and 1 IPv6 per interface, then probably case 4 is unnecessary.

bahamat:
My comments above are limited to JPC only. On-prem SDC are unrestricted as to how many IPs can be on a single interface.

tgross:
> On-prem SDC are unrestricted as to how many IPs can be on a single interface.

And we're not limiting ourselves to Triton in any case, so who knows what someone is going to want to deploy on.

[...]

PR #58, opened 2015-12-29
Implement interface specification syntax

justenwalker:
For #53

## Changes

- Factor out IP/Interface logic into its own go file
- Implement stable sorting of interfaces and their IP
- Implement Interface Specification syntax for the `interfaces` option

## Interface Specification 

- `eth0` : Match the first IPv4 address on `eth0` (alias for `eth0:inet`)
- `eth0:inet6` : Match the first IPv6 address on `eth0`
- `eth0[1]` : Match the 2nd IP address on `eth0`
- `10.0.0.0/16` : Match the first IP that is contained within the IP Network
- `fdc6:238c:c4bc::/48` : Match the first IP that is contained within the IPv6 Network
- `inet` : Match the first IPv4 Address (excluding `127.0.0.0/8`)
- `inet6` : Match the first IPv6 Address (excluding `::1/128`)

## Stable Sorting

Interfaces and their IP addresses are ordered alphabetically by interface name, then by IP address (lexicographically by bytes).

- eth0 10.2.0.1 192.168.1.100
- eth1 10.0.0.100 10.0.0.200
- eth2 10.1.0.200 fdc6:238c:c4bc::1
- lo ::1 127.0.0.1


src/containerbuddy/ips_test.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	if ip, _ := GetIP([]string{"lo", "eth0"}); ip != "127.0.0.1" {
+		t.Errorf("Expected to find loopback ip, but found: %s", ip)
+	}
+	if ip, err := GetIP([]string{"interface-does-not-exist"}); err == nil {
+		t.Errorf("Expected interface not found, but instead got an IP: %s", ip)
+	}
+}
+
+func TestOnChangeCmd(t *testing.T) {
```
tgross: I think this `TestOnChangeCmd` and `TestHealthCheck` got moved to this file accidentally.

tgross:
We should add `cover/coverage.out` to the `.PHONY` block of the Makefile. I got confused trying to check something in the tests locally because I didn't have a clean test environment.

src/containerbuddy/ips.go
```diff
@@ -0,0 +1,253 @@
...
+	 * an error state. */
+	if interfaceIPsErr != nil && len(interfaceIPs) < 1 {
+		return "", interfaceIPsErr
+	}
+	/* We had error(s) and there were interfaces returned, this is potentially
+	 * recoverable. Let's pass on the parsed interfaces and log the error
+	 * state. */
+	if interfaceIPsErr != nil && len(interfaceIPs) > 0 {
+		log.Printf("We had a problem reading information about some network "+
+			"interfaces. If everything works, it is safe to ignore this"+
```
tgross: incredibly tiny nitpick: you'll want a space here after "this"

src/containerbuddy/ips.go
```diff
@@ -0,0 +1,253 @@
...
+		spec, err := parseInterfaceSpec(iface)
+		if err != nil {
+			errors = append(errors, err.Error())
+			continue
+		}
+		specs = append(specs, spec)
+	}
+	if len(errors) > 0 {
+		err := fmt.Errorf(strings.Join(errors, "\n"))
+		println(err.Error())
```
tgross: This should be `log.Println(err)` I think?

src/containerbuddy/ips.go
```diff
@@ -0,0 +1,253 @@
...
+	if spec == "inet6" {
+		return interfaceSpec{IPv6: true, Name: "*"}, nil
+	}
+
+	if match := ifaceSpec.FindStringSubmatch(spec); match != nil {
+		name := match[1]
+		index := match[2]
+		inet := match[3]
+		if index != "" {
+			i, _ := strconv.Atoi(index)
```
tgross: I get that you're swallowing this error because `i` will be 0 if we get an error back, but why 1-index this instead of 0-indexing it in the first place?

src/containerbuddy/ips.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	return iip.IP.String()
+}
+
+func (iip interfaceIP) String() string {
+	return fmt.Sprintf("%s:%s", iip.Name, iip.IPString())
+}
+
+// Queries the network interfaces on the running machine and returns a list
+// of IPs for each interface. Currently, this only returns IPv4 addresses.
```
tgross: I believe this comment is now stale: `Currently, this only returns IPv4 addresses.` ?

src/containerbuddy/ips.go
```diff
@@ -0,0 +1,253 @@
...
+	if match := ifaceSpec.FindStringSubmatch(spec); match != nil {
+		name := match[1]
+		index := match[2]
+		inet := match[3]
+		if index != "" {
+			i, _ := strconv.Atoi(index)
+			if i > 0 {
+				return interfaceSpec{Spec: spec, Name: name, Index: i}, nil
+			}
+			return interfaceSpec{Spec: spec}, fmt.Errorf("Unable to parse interface spec: %s. Index should be > 0", spec)
```
tgross: As much as I hate the common practice in golang of returning `nil, err`... it is idiomatic. Why return an object if there's an error here? (Similar for the final return on line 165).

src/containerbuddy/ips.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// Stable Sort the interface IPs so that selecting the correct IP in GetIP
+	// can be consistent
+	sort.Stable(ByInterfaceThenIP(ifaceIPs))
+
+	/* If we had any errors parsing interfaces, we accumulate them all and
+	 * then return them so that the caller can decide what they want to do. */
+	if len(errors) > 0 {
+		err := fmt.Errorf(strings.Join(errors, "\n"))
+		println(err.Error())
```
tgross: `log` rather than `println` here, I think

src/containerbuddy/ips.go
tgross: If you combine both comments above, I think this collapses nicely to:

```
if index != "" {
    if i, err := strconv.Atoi(index); err != nil {
        return nil, fmt.Errorf("Unable to parse...")
    } else {
        return interfaceSpec{Spec: spec, Name: name, Index: i}, nil
    }
}
```

src/containerbuddy/ips.go
justenwalker: I made it 1-index for two reasons.

1. It was easier to explain in the readme. (1 is the 1st, 2 is the 2nd, etc...)
2. It made the code a little cleaner - I don't have to use -1 instead of 0 for Index - which i have to set explicitly when creating the struct.

However; the target audience understands 0-index, so I can switch it easily enough.

src/containerbuddy/ips.go
justenwalker: I can't return nil

> cannot use nil as type interfaceSpec in return argument

I think I'd have to switch the return type to `*interfaceSpec`

src/containerbuddy/ips.go
justenwalker: I carried this over from this:
https://github.com/joyent/containerbuddy/blob/master/src/containerbuddy/config.go#L467

However, I do agree it should be logged. I'll change it in both places.

src/containerbuddy/ips.go
tgross: hm, yeah. On second thought I think you've got the right idea here.

src/containerbuddy/ips.go
tgross: Yeah, I'd prefer the 0-index.

src/containerbuddy/ips_test.go
justenwalker: Moved it (	0e0a934)

justenwalker:
@tgross Found a better way to handle interface spec matching using interfaces.  Refactored the code accordingly (9b3cd71)

src/containerbuddy/ips.go
```diff
@@ -0,0 +1,282 @@
...
+
+	// Interface not found, return error
+	return "", fmt.Errorf("None of the interface specifications were able to match\nSpecifications: %s\nInterfaces IPs: %s",
+		specs, interfaceIPs)
+}
+
+// Interface Spec
+type interfaceSpec interface {
+	Match(index int, iip interfaceIP) bool
+	String() string
```
tgross: As far as I can tell this is an unused method.

src/containerbuddy/ips.go
```diff
@@ -0,0 +1,282 @@
...
+	return s.Spec
+}
+
+func (s baseInterfaceSpec) Match(index int, iip interfaceIP) bool {
+	return false
+}
+
+// -- matches inet, inet6, interface:inet, and interface:inet6
+type inetInterfaceSpec struct {
+	interfaceSpec
```
tgross: What is this getting us? We only need the struct to have implement the `Match` method for it to be an implementation of the `interfaceSpec` interface.

src/containerbuddy/ips.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// Interface Spec
+type interfaceSpec interface {
+	Match(index int, iip interfaceIP) bool
+	String() string
+}
+
+// -- Base Struct
+type baseInterfaceSpec struct {
```
tgross: It's not clear to me why we need a semi-empty version of this struct if we're only using it as an empty-value return. Adding `Spec string` to the "real" implementations seems like less indirection.

src/containerbuddy/ips.go
justenwalker: Isn't String() used implicitly when doing fmt.Printf ? https://golang.org/pkg/fmt/

src/containerbuddy/ips.go
justenwalker: Was trying to keep things DRY so I wouldn't have to re-implement String() for each class.

src/containerbuddy/ips.go
justenwalker: Can do - just means a lot of duplicate func String() methods.

justenwalker:
@tgross Most of the new comments relate to the String() method. I am still new to go - so maybe I'm probably coding Java in golang instead of doing something more idiomatic.  If you want me to cut out String() completely, let me know.

src/containerbuddy/ips.go
tgross: Ok, but `interfaceSpec` isn't a struct, it's an interface. So you'd need to embed `baseInterfaceSpec` here for that, right?

src/containerbuddy/ips.go
tgross: Yeah, but we only care about that if the `fmt` is user-visible, in which case we could just pass `someSpecInstance.Spec` to the formatter.

src/containerbuddy/ips.go
justenwalker: Yeah, thats a mistake. I'll fix that.

justenwalker:
@tgross I've removed String() and baseInterfaceSpec in 3379712. Since I'm using an interface not instead of a struct type, we can return nil instead of needing some kind of base class. I think this is for the best.

tgross:
> Most of the new comments relate to the String() method (and inheriting it through baseInterfaceSpec). I am still new to go - so I'm probably coding Java in golang instead of doing something more idiomatic.

Honestly, if this were a non-technical-user-facing application, I'd say it'd be the right way to go because you're going to want to return pretty-printed errors to the user.

But for Containerbuddy's audience, returning the default struct formatting by default is probably a better approach when it comes to debugging and logging -- this way the user gets whatever fields in the struct that we've managed to construct at that point instead of just the `Spec` field, which is (in a sense) just the input to the constructor.

tgross:
Ok, this is good-to-go. Nice work, @justenwalker

Issue #10, opened 2015-11-06
Support tags for service registration


tgross:
Allow the configuration of a service registration to include tags that the discovery backend can expose for consumption by `backend` handlers. (ex. "prod", "dev")

tgross:
@misterbisson this was one of your suggestions from my original notes. Do you have further thoughts on what this looks like?

misterbisson:
I'd leave this one in the icebox for now. Tagging for `dev` or `prod` is discussed in some contexts, but I'd rather isolate those environments in entirely different networks with separate discovery catalogs.

tgross:
Merged in https://github.com/joyent/containerbuddy/pull/59 and will be in the next release (0.0.6).

tgross:
Released in https://github.com/joyent/containerbuddy/releases/tag/0.1.0

PR #59, opened 2015-12-30
Add support for tags

justenwalker:
For #10 

Add support for `tags` in the service configuration. Didn't seem like too much work to add, so I'm afraid I may have glossed over some important details. 

Another set of eyes appreciated.

### DIFF
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -76,6 +76,7 @@ type BackendConfig struct {
 	Name             string          `json:"name"`
 	Poll             int             `json:"poll"` // time in seconds
 	OnChangeExec     json.RawMessage `json:"onChange"`
+	Tag              string          `json:"tag,omitempty"`
 	discoveryService DiscoveryService
 	lastState        interface{}
 	onChangeCmd      *exec.Cmd
--- a/src/containerbuddy/config.go
+++ b/src/containerbuddy/config.go
@@ -65,6 +65,7 @@ type ServiceConfig struct {
 	Port             int             `json:"port"`
 	TTL              int             `json:"ttl"`
 	Interfaces       json.RawMessage `json:"interfaces"`
+	Tags             []string        `json:"tags,omitempty"`
 	discoveryService DiscoveryService
 	ipAddress        string
 	healthCheckCmd   *exec.Cmd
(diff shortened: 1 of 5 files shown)


src/containerbuddy/config.go
```diff
@@ -65,6 +65,7 @@ type ServiceConfig struct {
 	Port             int             `json:"port"`
 	TTL              int             `json:"ttl"`
 	Interfaces       json.RawMessage `json:"interfaces"`
+	Tags             []string        `json:"tags,omitempty"`
```
tgross: I think we're going to need to include tags for the `BackendConfig` as well. It's my understanding that the primary use case for this that @misterbisson was talking about in #10 was being able to split out services for e.g. dev vs prod. So when we query Consul we're probably going to want to have a list of tags to send to [`checkHealth`](https://github.com/joyent/containerbuddy/blob/7249deb6e69fd66469a3f4ed366186d28937769b/src/containerbuddy/consul.go#L84) and then on to [`Health.Service`](https://godoc.org/github.com/hashicorp/consul/api#Health.Service)

justenwalker:
@tgross What would the filter behavior be for tags on a Backend. Would they filter with AND (Must have every tag) or with OR (must have one of the tags)

justenwalker:
I checked the [consul codebase](https://github.com/hashicorp/consul/blob/fed14582145987a6f56f4e21c841f9bfd97d411f/consul/state/state_store.go) for clarification.  It just accepts 1 tag (ignoring case) and filters if that tag is found.  So I think we should just use a single tag.

tgross:
I'm feeling like there's going to come a point where we want query-with-multiple-tags support but without a backend that supports it I feel that'd be YAGNI. It's easy enough to add it in later.

src/containerbuddy/config_test.go
```diff
@@ -85,6 +87,15 @@ func TestValidConfigParse(t *testing.T) {
 		t.Errorf("Expected 3 args but got unexpected results: %v", args)
 	}
 
+	expectedTags := []string{"tag1", "tag2"}
```
tgross: Can we make sure we have a test for empty tags too? Just to verify that we don't regress with it down the road?

tgross:
Ok, this LGTM.

PR #60, opened 2015-12-30
I can't believe we never added @justenwalker to the contributors list

tgross:
Credit where credit is due! cc @misterbisson @justenwalker


misterbisson:
:+1:

Issue #11, opened 2015-11-06
Add an alternate service discovery backend


tgross:
The PR https://github.com/joyent/containerbuddy/pull/4 provided by @bbox-kula made the changes we needed to have a generic interface for service discovery so we can have pluggable backends as intended. Including a second implementation of this backend as an example would make sure this works out in practice.

TODO on this issue: pick a second backend

stonevil:
+1

I'm interesting to add etcd backend support.

tenfourty:
+1 on etcd as the next backend to be supported.

tgross:
@stonevil did you want to take that on?

bbox-kula:
@stonevil I've got some ideas, if you wanted to bounce ideas off of someone. Otherwise, after things settle down here I'd take a crack at adding it.

tgross:
Released in https://github.com/joyent/containerbuddy/releases/tag/0.1.0

PR #61, opened 2016-01-01
Add support for etcd

justenwalker:
Adds etcd as an alternate discovery backend

## Changes
- Add dependency on `github.com/coreos/etcd/client`
- Implement DiscoveryService for etcd
- Create etcd example images
- Clean up makefile and examples to support multiple backend examples
- Support running both examples from the makefile
  - `make run-consul` to run the consul example
  - `make run-etcd` to run the etcd example
  - `make clean-consul` to stop and clean up the consul example
  - `make clean-etcd` to stop and clean up the etcd example

For #11 

🎊 **Happy New Year** 🎊


src/containerbuddy/etcd.go
```diff
@@ -0,0 +1,240 @@
...
+	ttl, _ := time.ParseDuration(fmt.Sprintf("%ds", service.TTL))
+	_, err := c.API.Set(context.Background(), key, "", &client.SetOptions{Dir: true, TTL: ttl, PrevExist: client.PrevExist})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c Etcd) deregisterService(service *ServiceConfig) error {
+	_, err := c.API.Delete(context.Background(), c.getNodeKey(service), &client.DeleteOptions{Dir: true, Recursive: true})
```
tgross: can you break this long line?

src/containerbuddy/etcd.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// MarkForMaintenance removes this instance from the registry
+func (c Etcd) MarkForMaintenance(service *ServiceConfig) {
+	c.deregisterService(service)
+}
+
+// SendHeartbeat refreshes the TTL of this associated etcd node
+func (c Etcd) SendHeartbeat(service *ServiceConfig) {
+	if c.checkServiceExists(service) {
```
tgross: According to https://godoc.org/github.com/coreos/etcd/client#KeysAPI the `Set` endpoint returns an error. Rather than making a call to test for existence and then making a second call, we should just make the `Set` call and catch the error, falling back to `registerService` if we get the error.

src/containerbuddy/etcd.go
```diff
@@ -0,0 +1,240 @@
...
+		// but we do want to make sure we've written the key for this map
+		etcdUpstreams[backend.Name] = services
+	}
+	return didChange
+}
+
+func (c Etcd) checkServiceExists(service *ServiceConfig) bool {
+	key := c.getNodeKey(service)
+	_, err := c.API.Get(context.Background(), key, nil)
+	if err != nil {
```
tgross: The golang-idiomatic way to do this test if you don't need the return values to live outside the scope of the `if` block is:

```go
if _, err := c.API.Get(context.Background(), key, nil); err != nil {
    // stuff
}
```

src/containerbuddy/etcd.go
```diff
@@ -0,0 +1,240 @@
+package main
```
tgross: `discovery_test.go` really just tests the Consul service interface. We should move most or all of that into `consul_test.go` and include `etcd_test.go`.

src/containerbuddy/etcd.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (c Etcd) registerService(service *ServiceConfig) error {
+	key := c.getNodeKey(service)
+	value := encodeEtcdNodeValue(service)
+	ttl, _ := time.ParseDuration(fmt.Sprintf("%ds", service.TTL))
+	_, err := c.API.Set(context.Background(), key, "", &client.SetOptions{Dir: true, TTL: ttl})
+	if err != nil {
+		return err
+	}
+	_, err = c.API.Set(context.Background(), fmt.Sprintf("%s/%s", key, "/service"), value, nil)
```
tgross: Is having to make two calls here potentially exposing us to getting into an inconsistent state if the second one fails? Will it just register safely on the next health check? We should probably have some notes here about what the expected behavior is.

src/containerbuddy/etcd.go
```diff
@@ -0,0 +1,240 @@
...
+	switch endpoints := config["endpoints"].(type) {
+	case string:
+		etcdConfig.Endpoints = []string{endpoints}
+	case []string:
+		etcdConfig.Endpoints = endpoints
+	default:
+		log.Fatal("Must provide etcd endpoints")
+	}
+
+	prefix, ok := config["prefix"].(string)
```
tgross: This will result in a panic if `prefix` isn't set: http://play.golang.org/p/v1uYdrWXAM

src/containerbuddy/etcd.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (c Etcd) getServices(appName string) ([]EtcdServiceNode, error) {
+	var services []EtcdServiceNode
+
+	key := c.getAppKey(appName)
+	resp, err := c.API.Get(context.Background(), key, &client.GetOptions{Recursive: true})
+	if err != nil {
+		if etcdErr, ok := err.(client.Error); ok {
+			if etcdErr.Code == client.ErrorCodeKeyNotFound {
```
tgross: I'm not super-familiar with etcd... what's the circumstance where we'd see this?

README.md
```diff
@@ -95,9 +95,32 @@ Backend fields:
...
+    "etcd": {
+        "endpoints": [
+            "http://etcd:4001"
+        ],
+        "prefix": "/containerbuddy"
+    }
+    ```
+
+    - `endpoints` is the list of etcd nodes in your cluster
+    - `prefix` is the path that will be prefixed to all service discovery keys.
```
tgross: `prefix` should be marked as optional and defaulting to `/containerbuddy`, if I'm reading the intent of the code right.

src/containerbuddy/etcd.go
justenwalker: The panic in the example refers to line 12, the setup code.  Seems fine when I try it.

http://play.golang.org/p/mHGCoL02Kt

src/containerbuddy/etcd.go
justenwalker: KeyNotFound errors are like 404's.  Need to check for them since they are not errors, just an indication that there are no service.

Use Case:

Backend health check is run before the backend has a chance to register itself with etcd. Rather than result in an error, it will return an empty array instead - which is more indicative of the actual state.

src/containerbuddy/etcd.go
tgross: D'oh! :grinning: Ok!

src/containerbuddy/etcd.go
tgross: Just for the record, our equivalent section of the Consul code is [here](https://github.com/joyent/containerbuddy/blob/aacd26d596d5eecdb9fa0164618273ac87f27e15/src/containerbuddy/consul.go#L84) but yeah I checked in the [Consul API code](https://github.com/hashicorp/consul/blob/0d81e4bab4e0fc8fdbaabf99050468591edad128/api/health.go#L100-L103 
) and the overall behavior should be identical -- an empty array returned on non-error states.

src/containerbuddy/etcd.go
justenwalker: :+1: done

README.md
```diff
@@ -201,7 +224,7 @@ $ curl -Lo containerbuddy-0.0.4.tar.gz \
 $ tar -xf containerbuddy-0.0.4.tar.gz
 $ cp ./containerbuddy ./nginx/opt/containerbuddy/
 $ cp ./containerbuddy ./app/opt/containerbuddy/
-./start.sh -p example -f docker-compose-local.yml
+./run.sh -p example -f docker-compose-local.yml
```
tgross: The paths aren't right here anymore. We have `./examples/etcd/docker-compose-local.yml` and `./examples/consul/docker-compose-local.yml`. Also, we're missing the non-local compose for the etcd example.

tgross:
@justenwalker we've got a merge conflict here now that https://github.com/joyent/containerbuddy/pull/59 has been merged.

justenwalker:
@tgross rebased

README.md
justenwalker: I can fix up this in the readme, but I don't have a good way of creating the non-local compose for etcd since it has some external dependencies for which I dont have access. 

Aside - the readme make reference to a specific version of Containerbuddy for the tarball. Should this be updated when the version gets bumped?

tgross: > I can fix up this in the readme, but I don't have a good way of creating the non-local compose for etcd since it has some external dependencies to which I don't have access.

It really should just be the same as the one in the `consul` directory with the appropriate `etcd` vs `consul` sections swapped.

> Aside - the readme make reference to a specific version of Containerbuddy for the tarball. Should this be updated when the version gets bumped?

Ugh. When we decide to cut a 1.0 release that problem will go away because we can link to "latest". But for now, yeah.

examples/etcd/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+    hostname: etcd
+    expose:
+    - 2379
+    - 2380
+    - 4001
+    ports:
+    - 4001:4001
+    restart: always
+
+nginx:
```
tgross: The ports/expose spec for this should be:

```
ports:
- 80
```

without an `expose` stanza.

examples/etcd/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+    command: >
+      /opt/containerbuddy/containerbuddy
+      -config file:///opt/containerbuddy/nginx.json
+      nginx -g "daemon off;"
+
+app:
+    image: 0x74696d/containerbuddy-etcd-demo-app
+    mem_limit: 512m
+    expose:
+    - 8000
```
tgross: We don't need the `expose` stanza here.

justenwalker:
@tgross rebased

tgross:
Awesome. Let's ship it!

PR #65, opened 2016-01-05
Make golang dependencies a non-phony target

tgross:
For https://github.com/joyent/containerbuddy/issues/64

Mount the `.godeps/` directory (on the local machine) to a spot nested within the `/go/src` directory in the build container so that we can run our build without manipulating the `GOPATH` but still take advantage of caching the cloned dependencies on disk.

cc @justenwalker

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,4 +29,4 @@ release/
 cover/
 examples/*/opt/containerbuddy/containerbuddy
 *.tar.gz
-build-image
+.godeps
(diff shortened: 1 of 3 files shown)


tgross:
Odd... what I have here seems to work for the first run but then is treating `make test` like it's not PHONY. Looking into it.

tgross:
> Odd... what I have here seems to work for the first run but then is treating make test like it's not PHONY. Looking into it.

Ah, which you already ran into in https://github.com/joyent/containerbuddy/pull/58/files#diff-cd5dbf3629c558b0104ba8f0937d6816R48. I'll rebase on that and double-check.

tgross:
Ok, rebased on `joyent/master` and that seems to work as expected. Once TravisCI give us the OK, I think this is ready, @justenwalker

justenwalker:
LGTM - it's really nice to increase the iteration speed like this.

PR #70, opened 2016-01-08
Pin GOMAXPROCS=1

tgross:
The `GOMAXPROCS` runtime variable is set by default (in Go 1.5+) to be equal to the number of visible CPUs. The runtime uses M:N threading and so we can expect multiple threads. Containerbuddy has minimal performance requirements, so potentially creating a thread for each core will increase thread-switching for the main application that we care about. This might particularly be an issue in environments like Triton where `runtime.NumCPU` will see all the cores on the underlying CN and not the number that are likely to be available for scheduling.

References:

https://golang.org/pkg/runtime/#GOMAXPROCS
https://docs.google.com/document/d/1At2Ls5_fhJQ59kDK2DFVhFu3g5mATSXqqV5QrxinasI/edit

### DIFF
--- a/src/containerbuddy/main.go
+++ b/src/containerbuddy/main.go
@@ -5,11 +5,15 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"runtime"
 	"syscall"
 	"time"
 )
 
 func main() {
+	// make sure we use only a single CPU so as not to cause
+	// contention on the main application
+	runtime.GOMAXPROCS(1)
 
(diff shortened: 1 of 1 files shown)


tgross:
"ping" on this PR.

justenwalker:
LGTM @tgross I didn't see a problem with this when I looked and it's innocuous enough.  Container buddy should have a small cpu footprint anyway.

tgross:
Cool, let's merge it!

PR #71, opened 2016-01-08
Add signal handler for SIGCHLD so we reap children

tgross:
For https://github.com/joyent/containerbuddy/issues/62

Also fires the handler for SIGCHLD when we fire our SIGTERM handler so that any children of the terminated process get reaped.

### DIFF
--- a/integration_tests/fixtures/zombie_maker/Dockerfile
+++ b/integration_tests/fixtures/zombie_maker/Dockerfile
@@ -1,0 +1,11 @@
+FROM alpine:3.3
+
+ADD build/containerbuddy /bin/containerbuddy
+ADD app.json /app.json
+
+ADD zombie.sh /zombie.sh
+RUN chmod 755 /zombie.sh
+
+ENTRYPOINT [\
+  "/bin/containerbuddy","-config=file:///app.json",\
+  "tail","-f"]
--- a/integration_tests/fixtures/zombie_maker/app.json
+++ b/integration_tests/fixtures/zombie_maker/app.json
@@ -1,0 +1,13 @@
+{
+  "consul": "consul:8500",
+  "services": [
+    {
+      "name": "app",
+      "port": 8000,
+      "health": "/bin/sh /zombie.sh",
+      "poll": 1,
+      "ttl": 5,
+      "tags": ["application"]
+    }
+  ]
+}
--- a/integration_tests/fixtures/zombie_maker/zombie.sh
+++ b/integration_tests/fixtures/zombie_maker/zombie.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+sleep 1 & exec /bin/sleep 2
--- a/integration_tests/tests/test_reap_zombies/docker-compose.yml
+++ b/integration_tests/tests/test_reap_zombies/docker-compose.yml
@@ -1,0 +1,12 @@
+consul:
+    image: "cbfix_consul"
+    mem_limit: 256m
+    hostname: consul
+
+app:
+    image: "cbfix_zombie_maker"
+    mem_limit: 512m
+    links:
+      - consul:consul
+    volumes:
+      - '${CONTAINERBUDDY_BIN}:/bin/containerbuddy:ro'
--- a/integration_tests/tests/test_reap_zombies/run.sh
+++ b/integration_tests/tests/test_reap_zombies/run.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+docker-compose up -d consul app > /dev/null 2>&1
+APP_ID="$(docker-compose ps -q app)"
+sleep 6
+NUM_ZOMBIES=$(docker exec $APP_ID ps -o stat,ppid,pid,comm | awk '
+BEGIN { count=0 }
+$1 ~ /^Z/ && $2 ~ /1/ { count++ }
+END { print count }
+')
+if [ $NUM_ZOMBIES -gt 1 ]; then
+  echo "Number of zombies > 1: $NUM_ZOMBIES" >&2
+  docker exec $APP_ID ps -o stat,ppid,pid,args
+  docker logs $APP_ID
+  exit 1
+fi
+exit 0
--- a/integration_tests/tests/test_sighup_deadlock/run.sh
+++ b/integration_tests/tests/test_sighup_deadlock/run.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-docker-compose up -d consul app > /dev/null 2>&1
+docker-compose up -d consul app
 APP_ID="$(docker-compose ps -q app)"
 docker-compose run --no-deps test /go/bin/test_probe test_sighup_deadlock $APP_ID > /dev/null 2>&1
 result=$?
--- a/src/containerbuddy/main.go
+++ b/src/containerbuddy/main.go
@@ -22,6 +22,9 @@ func main() {
 	}
 
 	// Set up handlers for polling and to accept signal interrupts
+	if 1 == os.Getpid() {
+		reapChildren()
+	}
 	handleSignals(config)
 	handlePolling(config)
 
--- a/src/containerbuddy/main.go
+++ b/src/containerbuddy/main.go
@@ -116,8 +116,12 @@ func executeAndWait(cmd *exec.Cmd) (int, error) {
 				return status.ExitStatus(), err
 			}
 		}
-		// only happens if we misconfigure, so just die here
-		log.Fatal(err)
+		// Should only happen if we misconfigure or there's some more
+		// serious problem with the underlying open/exec syscalls. But
+		// we'll let the lack of heartbeat tell us if something has gone
+		// wrong to that extent.
+		log.Println(err)
+		return 1, err
 	}
 	return 0, nil
 }
--- a/src/containerbuddy/signals.go
+++ b/src/containerbuddy/signals.go
@@ -106,7 +106,7 @@ func forAllServices(config *Config, fn serviceFunc) {
 	}
 }
 
-// Listen for and capture signals from the OS
+// Listen for and capture signals used for orchestration
 func handleSignals(config *Config) {
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, syscall.SIGUSR1, syscall.SIGTERM, syscall.SIGHUP)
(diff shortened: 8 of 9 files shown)


tgross:
Looks like this is failing intermittently (worked on my machine! :grinning: ) because of a race in waiting for the child processes to exit. But that actually helps us find a bug, which is that we need to safely handle cases where the child processes have cleaned themselves up.

tgross:
Nevermind, I never registered the signal handler... let me fix, retest, and rebase.

tgross:
Still getting an error from TravisCI, but it doesn't match what I'm seeing:

Travis:

```
...
=== RUN   TestMaintenanceSignal
--- PASS: TestMaintenanceSignal (0.00s)
=== RUN   TestTerminateSignal
2016/01/08 15:30:57 Marking for maintenance: test-service
.2016/01/08 15:30:57 Deregistering service: test-service
2016/01/08 15:30:58 wait: no child processes
--- PASS: TestTerminateSignal (1.00s)
=== RUN   TestReloadSignal
exit status 1
FAIL	containerbuddy	7.102s
makefile:36: recipe for target 'test' failed
make: *** [test] Error 1
make: *** [test] Error 2
```

On my machine:
```
...
=== RUN   TestMaintenanceSignal
2016/01/08 15:31:29 Marking for maintenance: test-service
--- PASS: TestMaintenanceSignal (0.00s)
=== RUN   TestTerminateSignal
.2016/01/08 15:31:29 Deregistering service: test-service
--- PASS: TestTerminateSignal (1.00s)
=== RUN   TestReloadSignal
2016/01/08 15:31:30 Reloading configuration.
2016/01/08 15:31:30 Could not reload config: Parse error at line:col [1:1]: invalid character 'i' looking for beginning of value
    1: invalid
-------^
2016/01/08 15:31:30 Reloading configuration.
2016/01/08 15:31:30 Deregistering service: test-service
--- PASS: TestReloadSignal (0.00s)
=== RUN   TestSignalWiring
2016/01/08 15:31:30 Reloading configuration.
...
```

The order that I'm getting the "Reloading configuration" and "RUN TestSignalWiring" tells me that we're still seeing some kind of concurrency bug around the signal handlers, possibly in the tests themselves.

tgross:
Particularly tricky thing with this is that even Golang's stdlib skips any SIGCHLD tests: https://golang.org/src/os/signal/signal_test.go

justenwalker:
@tgross I was thinking earlier that this might need to be done as an integration test since its very difficult to get golang to behave when it comes to trapping signals in unit tests.

tgross:
>  I was thinking earlier that this might need to be done as an integration test

At least for the SIGTERM and SIGCHLD signal tests, I'm inclined to agree with you. Should we sit on this PR until we have the integration test rig set up maybe?

justenwalker:
> Should we sit on this PR until we have the integration test rig set up maybe?

I think so. It might make this significantly easier.

tgross:
I'm going to remove the unworkable unit test and rebase this branch on the updated master with our integration test suite.

tgross:
So after rebasing and running the integration tests I'm getting integration test failures, but I get different failure behaviors on TravisCI than I do locally. Notice that they run in a different order when I run them on my local `docker-machine`.

On Travis:
```
2016/01/14 13:23:01 ==== Running Integration Tests ====
2016/01/14 13:23:01 TEST: test_sigterm
2016/01/14 13:23:10 PASS: test_sigterm
2016/01/14 13:23:11 TEST: test_sighup_deadlock
2016/01/14 13:23:15 500 Internal Server Error: Cannot kill container 2a25d0c38dba962f5305f61527da85a747053442e30532f4a65317a865faa3ae: [2] Container does not exist: container destroyed
2016/01/14 13:23:15 FAIL: test_sighup_deadlock
2016/01/14 13:23:17 ==== Tests Failed! ====
make: *** [integration] Error 1
```

Locally:
```
2016/01/14 08:57:48 ==== Running Integration Tests ====
2016/01/14 08:57:48 TEST: test_sighup_deadlock
2016/01/14 08:58:03 PASS: test_sighup_deadlock
2016/01/14 08:58:04 TEST: test_sigterm
2016/01/14 13:58:37 Service app (tag:) count != 0
2016/01/14 08:58:38 FAIL: test_sigterm
2016/01/14 08:58:41 ==== Tests Failed! ====
make: *** [integration] Error 1
```

I first suspected ordering of tests, but if I tweak the `test.sh` script so as to force `test_sigterm` to be run first (same as on Travis), I still see only the `test_sigterm` test fail. So there's something about our environment that's leaking into the test runner.

tgross:
I've narrowed down the behavior a bit. Whenever the SIGCHLD handler block is [within the `select`](https://github.com/joyent/containerbuddy/pull/71/files#diff-536bffba9bba22f8afea4ecb1036fe5cR142), it is blocking the firing of *all* the other signal handlers. I suspect this has to do with our `signalLock` again, so I'm digging into that now.

tgross:
Ok, I put `strace` on Containerbuddy in one of the integration test targets and I think I understand the problem now. (ref https://gist.github.com/tgross/11e32dbd4b087ceabd09)

Our [`wait4`](http://linux.die.net/man/2/wait4) call (for which we'll need to read the [`waitpid` man page](http://linux.die.net/man/2/waitpid)) is being passed -1 for [the first arg](https://github.com/joyent/containerbuddy/pull/71/files#diff-536bffba9bba22f8afea4ecb1036fe5cR117) which means "wait for any child process." But we have *many* child processes for Containerbuddy because of the health checks. The `wait4` starts accumulating `SIGCHLD` events to process because the health check is running on another goroutine (see thread 38 in the gist). 

Compounding this is that we have a single goroutine catching all the signal handlers. This was intentional to avoid having signals interfere with each other, but it also means that as `wait4` accumulates a backlog of `SIGCHLD` events to handle (because of the health check running on another goroutine), it ends up blocking all our other signal handlers.

The fix for this is straightforward: move `reapChildren` into its own goroutine so that it doesn't block forward progress on the other signal notifications. Which was exactly what I'd suggested we not do in the original issue for what was a not-very-good reason. :blush: I've pulled the essential bits out of go-reaper again and run that goroutine starting from `main`.

After running the integration test a half-dozen more times to verify I'm not just missing a race, I've pushed a commit to do so and rebased. If TravisCI blesses this, I think we'll be good-to-go.

tgross:
Failed again, but with different failures on the two runs (my commit and this PR):

https://travis-ci.org/tgross/containerbuddy/builds/102441944
```
2016/01/14 20:57:12 ==== Running Integration Tests ====
2016/01/14 20:57:12 TEST: test_sigterm
2016/01/14 20:57:17 PASS: test_sigterm
2016/01/14 20:57:18 TEST: test_sighup_deadlock
2016/01/14 20:57:23 500 Internal Server Error: Cannot kill container 5ac72a7911ec144255b604a53fbc5d03a78f82a379a32860bd04c8ae2775cece: [2] Container does not exist: container destroyed
2016/01/14 20:57:23 FAIL: test_sighup_deadlock
2016/01/14 20:57:25 ==== Tests Failed! ====
```

https://travis-ci.org/joyent/containerbuddy/builds/102441965
```
2016/01/14 20:57:15 ==== Running Integration Tests ====
2016/01/14 20:57:15 TEST: test_sigterm
2016/01/14 20:57:20 PASS: test_sigterm
2016/01/14 20:57:21 TEST: test_sighup_deadlock
2016/01/14 20:58:05 Expected app to be healthy after SIGHUP: Service app (tag:) count != 1
2016/01/14 20:58:05 FAIL: test_sighup_deadlock
2016/01/14 20:58:07 ==== Tests Failed! ====
```

justenwalker:
So at some point the docker container for app_consul dies while getting the flood of SIGHUP signals?

tgross:
> So at some point the docker container for app_consul dies while getting the flood of SIGHUP signals?

It looks like it, yeah. I'm verifying that it's not dying before that in [build 107](https://travis-ci.org/joyent/containerbuddy/builds/102455993) now.

tgross:
Ok, [build 108](https://travis-ci.org/joyent/containerbuddy/builds/102460847) seems to confirm that. I'll pick this up again in the AM.

justenwalker:
FATAL executeAndWait(/reload-app.sh [/reload-app.sh]): wait: no child processes

I think perhaps we need to get rid of the log.Fatal because its causing the app to die when the polling checks are reaped.

justenwalker:
I was able to get it to pass and keep the `log.Fatal` by setting the options as in [simple_init.c](http://man7.org/tlpi/code/online/diff/namespaces/simple_init.c.html#child_handler)

```go
syscall.Wait4(-1, &wstatus, syscall.WNOHANG|syscall.WUNTRACED|syscall.WCONTINUED, nil)
```

tgross:
> I think perhaps we need to get rid of the log.Fatal because its causing the app to die when the polling checks are reaped.

I think you're referring to the `log.Fatal` in `main.go/executeAndWait`? The only time that should ever be reached is if the command can't start. So long as it returns an exit code of any kind, it returns `(int, error)`.

> I was able to get it to pass and keep the log.Fatal by setting the options as in simple_init.c

Looks like that did it! Thanks for helping out! I'll rebase to take out all the cruft of the TravisCI diagnostics and then we should be good-to-go here.

justenwalker:
> The only time that should ever be reached is if the command can't start.

So, if the health check, or onChange command can't start for some reason, the container dies.  Is that the behavior we want? Can we make it a special case for the main app instead?

tgross:
> So, if the health check, or onChange command can't start for some reason, the container dies. Is that the behavior we want? Can we make it a special case for the main app instead?

The primary case I was thinking of (and is mentioned in the comment) was where the path to the command was misconfigured. In that case, it's probably better to die than to try to continue. But I could see a case where the command might fail due to a transient IO error (i.e. the `open` syscall fails due to disk error) in which case yeah, maybe it would be better to remove that entirely and let the failure to send a heartbeat TTL be the thing that alerts us to failure.

I'll add a commit to this PR to that effect.

tgross:
That's done and as a bonus we got an extra test run on TravisCI just to make doubly-sure. :grinning:

justenwalker:
Nice. When I looked into the fatal call, the wait4 call that we originally had caused this code path to execute, so it leads me to believe that this error overlaps with some transient ones as well.

If we are concerned about a certain condition we can test for, perhaps we can test the existence/exec perm if the command before running and log.Fatal on that

tgross:
> If we are concerned about a certain condition we can test for, perhaps we can test the existence/exec perm if the command before running and log.Fatal on that

In retrospect, I think relying on the failed heartbeat makes for a more consistent interface for the operator. They'll see a service in the discovery DB but it'll be marked unhealthy. This way the behavior isn't special-cased.

tgross:
> I think we need to do
> Otherwise, this goroutine will exit after the first SIGCHLD

Good catch. We'd be able to handle more than one child process but not more than one signal. I've fixed that and added some commentary explaining why.

tgross:
I think this is ready to merge?

justenwalker:
code LGTM.  Should we add an integration test for killing zombie processes?

tgross:
> Should we add an integration test for killing zombie processes?

I like the idea of that. Do you know a reliable technique for creating them? I've looked into that a couple times in the past and wasn't able to figure out one that worked all the time.

justenwalker:
I've had success with this:

```sh
sleep 1 & exec /bin/sleep 10
```

justenwalker:
Check out justenwalker/containerbuddy@982c0d0

tgross:
> Check out justenwalker/containerbuddy@982c0d0

Awesome, thanks. I've cherry-picked that and tested it locally with success... waiting for Travis to give us the :+1:

tgross:
@justenwalker TravisCI build is green. Thanks for the assist on this... this issue has been a bit of a slog to get just a handful of lines of codes merged in!

Issue #67, opened 2016-01-05
Proposal: Integration Test Framework


justenwalker:
My goal is to make it somewhat easy to create new integration tests by factoring
out the common logic into a simple framework. 

It might be overkill though, maybe there is an easier way to do it.

### Folder Structure
- `integration_tests/fixtures` - Contains folder for each test harness
- `integration_tests/tests`  - Contains folders for each integration test

#### Fixtures

`integration_tests/fixtures/fixture_name` - Folder contains a `Dockerfile` 
for building a Conatainerbuddy app - resulting in an image named `fixture_name`

The `Dockerfile` should specify an ENTRYPOINT/CMD and should not require any arguments.

#### Tests

`integration_tests/tests/test_name` 
- Contains a `test.sh` script 
- `docker-compose.yml` for setting up the test environment
- Any other resources it might need. 

If `test.sh` returns success: `0` then the test passed, otherwise it failed

This script can make some assumptions:

- It's current directory PWD is the same as the script.
- `containerbuddy_etcd` and `containerbuddy_consul` are running and are fresh (no data)
- all fixtures in `integration_tests/fixtures` are created and are available as images - requiring no arguments

### Make target `integration_test`

```
integration_test: build
  ./test.sh

```

The script in the top-level folder named `test.sh` will:

### Setup Test Fixtures
- Scan through all folders in `integration_tests/fixtures` in alpha order
- For each fixture:
  - Copy `build` into `integration_test/fixtures/fixture_name/build` so it can easily be sourced by the `Dockerfile`
  - `cd integration_tests/fixtures/fixture_name`
  - `docker build -t fixture_name .`
  - Remove `build` folder to clean up (should add to .gitignore also)

*Note*: Since fixtures are created in alpha order, they can have FROM directives for previously
created images - so as to reduce duplication.

### Run tests
- Scan through all folders in `integration_tests/tests` in alpha order
- For each test:
  - Re-run `containerbuddy_etcd` and `containterbuddy_consul`
  - Run docker compose on the `docker-compose.yml` to bring up the test environment
  - Run `integration_test/tests/test_name/test.sh`
  - Stop/kill the compose environment

tgross:
What I might change is that it would be really nice/useful if the integration tests can use our example applications as test fixtures. This way the application that a new user might try out is one we know we've hit with testing.

[...]

PR #72, opened 2016-01-12
Integration Test Framework

justenwalker:
For #67 

Still a work in progress, but it's finally working for me in TravisCI.  

- There is a [README.md](https://github.com/justenwalker/containerbuddy/blob/integration_tests/integration_tests/README.md) in the integration_tests folder which describes how it operates and how to make tests.
- I've made an example test for SIGTERM. It's small golang application called [test_probe](https://github.com/justenwalker/containerbuddy/tree/integration_tests/integration_tests/fixtures/test_probe). It can be extended to support other tests easily - reducing the amount of bash wizardry needed to write tests.
- Added a test for #74 (but forced it to pass)

### DIFF
--- a/integration_tests/README.md
+++ b/integration_tests/README.md
@@ -1,0 +1,65 @@
+# Integration Tests
+
+## Running Tests
+
+- To run integration tests:
+
+`make integration` or `./test.sh test`
+
+- To run a single integration test:
+
+`./test.sh test test_name`
+
+- To clean fixtures. Next time tests are run they will be rebuilt.
+
+`make clean` or `./test.sh clean`
+
+## Making Tests
+
+### 1. Fixtures
+
+Fixtures Directory: `integration_tests/fixtures`
+
+- Folders in the fixtures directory will be built as docker images in alphabetical order
+- Each folder should contain a `Dockerfile` and any resources necessary to build the image
+- The resulting image will be named `cbfix_fixture_name`
+
+*Note*: Since fixtures are created in alpha order, they can have FROM directives for previously created images
+
+### 2. Tests
+
+Tests Directory: `integration_tests/tests`
+
+- Folders in the tests directory will be run as tests
+- Each folder must contain `run.sh`
+- The test folder can also contain `docker-compose.yml` for setting up the test environment and other resources it might need
+- If `run.sh` returns success: `0` then the test passed, otherwise it failed
+
+This script can make some assumptions:
+
+- It's current directory PWD is the same as the script.
+- Following environment variables are set:
+  - `COMPOSE_FILE`
+  - `COMPOSE_PROJECT_NAME` - The name of the test folder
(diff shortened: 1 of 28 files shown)


integration_tests/README.md
```diff
@@ -0,0 +1,56 @@
...
+- Each folder should contain a `Dockerfile` and any resources necessary to build the image
+- The resulting image will be named `cbfix_fixture_name`
+
+*Note*: Since fixtures are created in alpha order, they can have FROM directives for previously created images
+
+### 2. Tests
+
+Tests Directory: `integration_tests/tests`
+
+- Folders in the tests directory will be run as tests
```
tgross: I think it's clear to me and you but we should include a note here that these tests are run in the context of the testing host (i.e. TravisCI or your local environment), not inside a container.

tgross: This also implies (as the TravisCI config file says) that we need golang on the test-running host. We should make sure to document those assumptions as well. (Or do we want to run the tests themselves in a container?)

test.sh
```diff
@@ -0,0 +1,160 @@
...
+  find $TESTS_DIR -maxdepth 1 -mindepth 1 -type d |\
+    while read TDIR; do
+      cd $TDIR
+      export COMPOSE_PROJECT_NAME="$(basename $TDIR)"
+      export COMPOSE_FILE="./docker-compose.yml"
+      destroy_test_compose
+    done
+}
+
+case ${1:-"test"} in
```
tgross: Maybe we should have the option to run single tests here? I could imagine a case where if a developer is trying to fix a broken test that having to wait for the whole rig to build would be painful. That does mean that:

> *Note*: Since fixtures are created in alpha order, they can have FROM directives for previously created images

probably wouldn't work. But maybe that's a good thing?

integration_tests/README.md
justenwalker: Actually, we don't technically need Golang on the host environment.  the compose script can bring up a golang container to do testing instead of doing it locally.

Really I added the go language to `.travis.yml` because,  by default, TravisCI uses RVM and Ruby.  It just makes it clear which language we are targeting - has no actual effect the way we are using it.

test.sh
justenwalker: Easy enough.  Fixtures are created before tests.  I can make it run a specific test, but still build all fixtures.

justenwalker:
- Added option to run an individual integration test
- Made `CONTAINERBUDDY_BIN` available to docker-compose
- Altered `test_sighup_deadlock` to increase the chance of finding the issue
- Updated `README` to expand on how to run tests

integration_tests/README.md
tgross: Got it, understood.

tgross:
This looks great, nice work, @justenwalker

PR #74, opened 2016-01-12
Split maintenance mode lock from signal handler lock.

tgross:
For https://github.com/joyent/containerbuddy/issues/73.

If our polling functions are waiting on the `inMaintenanceMode` lock, then they can't select on their `quit` channel. This leads to a deadlock when a signal handler obtains a r/w lock on the same mutex and then tries to send on the `quit` channels (ex. `reloadConfig`). By splitting the `maintModeLock` from the `signalLock`, we can avoid this scenario.

### DIFF
--- a/src/containerbuddy/signals.go
+++ b/src/containerbuddy/signals.go
@@ -12,18 +12,21 @@ import (
 // globals are eeeeevil
 var paused bool
 var signalLock = sync.RWMutex{}
+var maintModeLock = sync.RWMutex{}
 
 // we wrap access to `paused` in a RLock so that if we're in the middle of
 // marking services for maintenance we don't get stale reads
 func inMaintenanceMode() bool {
-	signalLock.RLock()
-	defer signalLock.RUnlock()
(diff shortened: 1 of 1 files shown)


justenwalker:
LGTM - But it would be a good idea to make an integration test to catch this case.  I've started on the framework in #72 that might be helpful for this. I can try making a test for this.

justenwalker:
I added a test to #72 for catching that deadlock and it looks like it fails as expected - the service never comes back online from being in maintenance mode after receiving a SIGHUP.

tgross:
> I'm not sure how reliable this will be for reproducing the error

It probably won't be only inasmuch as it took probably 200-300 build/test cycles for the `triton-mysql` project I'm working on for it to first appear (although it was more frequent after that point due to whatever circumstance of timing. But it's better than nothing.

Unfortunately running `go test -race` doesn't work in our build -- there's what looks to be a glibc-specific component of the race detector. I'll open a new issue to see whether we can fix that in the builder-container.

Issue #76, opened 2016-02-06
Consider using godep for managing dependencies?


justenwalker:
[Godep](https://github.com/tools/godep) is pretty much the de-facto tool for managing dependencies in go projects. I suggest we consider using Godep instead of manually checking out and manipulating the source repositories.

This requires that the directory structure be altered to conform more closely with the [golang coding standards](http://golang.org/doc/code.html)

tgross:
I've talked about this notion at length [elsewhere](http://0x74696d.com/posts/go-get-considered-harmful/) but I think the Go community has their dependency management story wrong. A lot of this is because the core team works in a mono-repo environment, and lots of the community is coming from development communities where breaking simple monolithic applications down into a hundred tiny sublibraries (none of which are useful on their own) is considered best practice and where the dependency story is even worse.

Also, I'm not sure I agree that Godep is the de-facto complete solution at this point either. The Golang development team has introduced the ["vendor experiment"](https://docs.google.com/document/d/1Bz5-UB7g2uPBdOx-rw5t9MxJwkfpx90cqG9AFL0JAYo/edit?pref=2&pli=1) and it's almost certainly going to be the case that this is adopted as the standard. Godeps doesn't currently match up with this very well.

What's the value proposition for making this change?

justenwalker:
Mostly this came out of my inability to get my tooling to detect the libraries required for using autocompletion. Conforming to go's conventions seem to alleviate this for most tooling. 

I agree golang's dependency story is pretty bad, but it does seem that godep is used most often (a close second to using nothing). Vendor experiment is that - an experiment; however, godep supports this if you set an environment variable.

One use case is that I can run `godep restore` from the repository and my GOPATH will get all of the required libraries at their versions defined in Godeps.json - which allows my tooling (right now Atom Editor + go-plus) to find all the necessary libraries.

However; I concede that this proposal doesn't add much intrinsic value to the project or its users:
- It's a binary, not a library - so no one should be importing it into their own projects (yet?)
- It only really helps my workflow, but most likely doesn't help everyone (unless you're using atom + go-plus)

tgross:
As much as I'm skeptical of the official golang approach, realistically I'm can't oppose the idea of trying to bring things in-line with the community if it reduces friction for contributing. But it does look like Go 1.6 has [accepted](https://tip.golang.org/doc/go1.6#go_command) the vendor experiment:

> Go 1.6 keeps the vendoring support, no longer considered experimental, and enables it by default.

[...]

PR #77, opened 2016-02-06
Add support for godep

justenwalker:
- Conform to [golang standard](http://golang.org/doc/code.html) directory layout
- Install godep in Dockerfile
- Create initial Godeps workspace
- Add update-deps make command

For #76

### DIFF
--- a/Makefile.docker
+++ b/Makefile.docker
@@ -5,27 +5,23 @@ SHELL := /bin/bash
 .SHELLFLAGS = -o pipefail -euc
 .DEFAULT_GOAL := build
 
-.PHONY: all build check test cover update-deps restore
+.PHONY: all build check test cover update-deps vendor
 
 PACKAGE := github.com/joyent/containerbuddy
 SRC  := /go/src/${PACKAGE}
 CDWD := cd ${SRC}
 NO_VENDOR:= $(shell cd ${SRC} && go list ./... | grep -v /vendor/)
+GO15VENDOREXPERIMENT := 1
+export GO15VENDOREXPERIMENT
 
 all: build cover
 
-restore:
+vendor:
 	${CDWD} && godep restore
 
-build: restore
+build: vendor
 	${CDWD} && go build -a -o /build/containerbuddy -ldflags "$(LDFLAGS)" ./
 	chmod 755 /build/containerbuddy
-
-update-deps: restore
-	go get -u github.com/hashicorp/consul
-	go get -u github.com/coreos/etcd
-	${CDWD} && godep update github.com/hashicorp/consul
-	${CDWD} && godep update github.com/coreos/etcd
 
 lint:
 	${CDWD} && go vet ${NO_VENDOR}
--- a/Makefile.docker
+++ b/Makefile.docker
@@ -5,35 +5,35 @@ SHELL := /bin/bash
 .SHELLFLAGS = -o pipefail -euc
 .DEFAULT_GOAL := build
 
-.PHONY: all build check test cover
+.PHONY: all build check test cover update-deps restore
 
-CONSUL_REF := 158eabdd6f2408067c1d7656fa10e49434f96480
-ETCD_REF := release-2.2
-
-ROOT := /go
-CDWD := cd ${ROOT}/src/containerbuddy
+PACKAGE := github.com/joyent/containerbuddy
+SRC  := /go/src/${PACKAGE}
+CDWD := cd ${SRC}
+NO_VENDOR:= $(shell cd ${SRC} && go list ./... | grep -v /vendor/)
 
 all: build cover
 
-build: /go/src/github.com
-	${CDWD} && go build -a -o /build/containerbuddy -ldflags "$(LDFLAGS)"
+restore:
+	${CDWD} && godep restore
+
+build: restore
+	${CDWD} && go build -a -o /build/containerbuddy -ldflags "$(LDFLAGS)" ./
 	chmod 755 /build/containerbuddy
 
-/go/src/github.com:
-	mkdir -p "${ROOT}/src/github.com/hashicorp"
-	git clone https://github.com/hashicorp/consul.git ${ROOT}/src/github.com/hashicorp/consul
-	cd ${ROOT}/src/github.com/hashicorp/consul && git checkout ${CONSUL_REF}
-	mkdir -p "${ROOT}/src/github.com/coreos"
-	git clone https://github.com/coreos/etcd.git ${ROOT}/src/github.com/coreos/etcd
-	cd ${ROOT}/src/github.com/coreos/etcd && git checkout ${ETCD_REF}
+update-deps: restore
+	go get -u github.com/hashicorp/consul
+	go get -u github.com/coreos/etcd
+	${CDWD} && godep update github.com/hashicorp/consul
+	${CDWD} && godep update github.com/coreos/etcd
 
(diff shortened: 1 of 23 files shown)


tgross:
I know you're coming back to this but we definitely don't want to have the vendored dependencies committed in this repo, so we should add that to the `.gitignore`.

justenwalker:
@tgross I can ignore that folder; however, the `/vendor/` folder AFAIK should be included in in the repo.

In any case, godep restore would put all of the dependencies in the `GOPATH`, so the docker makefile will run - but it would break compatibility with `go get` or `go install`

justenwalker:
Alternatively, we could look into [govendor](https://github.com/kardianos/govendor) instead, which was designed specifically for the vendor spec.

tgross:
> I can ignore that folder; however, the /vendor/ folder AFAIK should be included in in the repo.

Only if we're using git subtrees or submodules. Vendoring the dependencies to the appropriate file system location is an orthogonal concern to the source control for those dependencies. Embedding them within our code obscures provenance and also fills our project history with someone else's project.

justenwalker:
amended commit - TravisCI build failed due to a curl error - no clear way to ask it to try again.

justenwalker:
> Only if we're using git subtrees or submodules. Vendoring the dependencies to the appropriate file system location is an orthogonal concern to the source control for those dependencies. Embedding them within our code obscures provenance and also fills our project history with someone else's project.

I agree - I hate vendoring in general. It seems like the way the go team wants you to treat `/vendor` is in [ruby/rails sense](http://stackoverflow.com/questions/11378921/what-does-the-term-vendoring-or-to-vendor-mean-for-ruby-on-rails). There is at least [some](http://blog.bithound.io/why-we-stopped-vendoring-our-npm-dependencies/) [dissent](https://gist.github.com/datagrok/8577287) for this practice. 

I've excluded it from this PR; However, in doing so - we are deviating from the intended use.

Makefile.docker
```diff
@@ -5,35 +5,35 @@ SHELL := /bin/bash
...
 	chmod 755 /build/containerbuddy
 
-/go/src/github.com:
-	mkdir -p "${ROOT}/src/github.com/hashicorp"
-	git clone https://github.com/hashicorp/consul.git ${ROOT}/src/github.com/hashicorp/consul
-	cd ${ROOT}/src/github.com/hashicorp/consul && git checkout ${CONSUL_REF}
-	mkdir -p "${ROOT}/src/github.com/coreos"
-	git clone https://github.com/coreos/etcd.git ${ROOT}/src/github.com/coreos/etcd
-	cd ${ROOT}/src/github.com/coreos/etcd && git checkout ${ETCD_REF}
+update-deps: restore
```
tgross: Am I misunderstanding this or does this update all our dependencies to the current tip? Can we put a giant "OMG don't do this" warning on the top of this?

tgross:
I'm seeing this in the Travis output:

> godep: WARNING: Go version (go1.5) & $GO15VENDOREXPERIMENT=1 wants to enable the vendor experiment, but disabling because a Godep workspace (Godeps/_workspace) exists

I think we want to make sure that directory isn't there.

tgross:
>  It seems like the way the go team wants you to treat /vendor is in ruby/rails sense

Ok, but they're profoundly wrong and I'm ok with taking a stance on that.

Makefile.docker
```diff
@@ -5,35 +5,35 @@ SHELL := /bin/bash
...
+	go get -u github.com/hashicorp/consul
+	go get -u github.com/coreos/etcd
+	${CDWD} && godep update github.com/hashicorp/consul
+	${CDWD} && godep update github.com/coreos/etcd
 
 lint:
-	${CDWD} && go vet
-	${CDWD} && go fmt
-	golint ${ROOT}/src/containerbuddy
+	${CDWD} && go vet ${NO_VENDOR}
```
tgross: What's this for?

tgross:
We should take a pass thru the top-level makefile too. We're mounting volumes that don't exist like `.godeps`

Makefile.docker
tgross: Could we maybe get a `make restore` at the top-level make too?

Makefile.docker
```diff
@@ -5,35 +5,35 @@ SHELL := /bin/bash
...
 
-CONSUL_REF := 158eabdd6f2408067c1d7656fa10e49434f96480
-ETCD_REF := release-2.2
-
-ROOT := /go
-CDWD := cd ${ROOT}/src/containerbuddy
+PACKAGE := github.com/joyent/containerbuddy
+SRC  := /go/src/${PACKAGE}
+CDWD := cd ${SRC}
+NO_VENDOR:= $(shell cd ${SRC} && go list ./... | grep -v /vendor/)
```
tgross: We're missing `GO15VENDOREXPERIMENT` here either in this makefile or in the Docker run environment.

Makefile.docker
```diff
@@ -5,35 +5,35 @@ SHELL := /bin/bash
...
 
 all: build cover
 
-build: /go/src/github.com
-	${CDWD} && go build -a -o /build/containerbuddy -ldflags "$(LDFLAGS)"
+restore:
```
tgross: I understand that the action here is to "restore" but maybe this should be called `vendor` so that we're taking advantage of make and not re-creating the dependencies?

makefile
```diff
@@ -6,18 +6,22 @@ SHELL := /bin/bash
...
 
 COMPOSE_PREFIX_ETCD := exetcd
 COMPOSE_PREFIX_CONSUL := exconsul
 
+
 DOCKERMAKE := docker run --rm \
 	--link containerbuddy_consul:consul \
 	--link containerbuddy_etcd:etcd \
-	-v ${ROOT}/src/containerbuddy:/go/src/containerbuddy \
 	-v ${ROOT}/.godeps:/go/src \
```
tgross: We're mounting volumes that don't exist like `.godeps` and we should be mounting the `vendor` directory so that we don't need to re-download.

tgross: I'm trying this out myself and I'm seeing some oddball behavior... godeps doesn't seem to respect GO15VENDORPATH at all, actually.

Makefile.docker
justenwalker: [Unfortunate artifact](https://github.com/golang/go/issues/11659) of how `/vendor` folder works.  ./... path expansion cannot be used because it would recurse into the vendor folder. you have to explicitly exclude it

Makefile.docker
justenwalker: It should be included in the Dockerfile, but I can add it here for redundancy/clarity

makefile
justenwalker: Did you mean `GO15VENDOREXPERIMENT` ? They dont reference GOVENDORPATH [anywhere in the code](https://github.com/tools/godep/search?utf8=%E2%9C%93&q=GO15VENDORPATH&type=Code)

Makefile.docker
justenwalker: I could remove this completely, we could take manual steps to upgrade the library ourselves.

makefile
tgross: D'oh, yes.

tgross:
Ok, thanks for slogging thru this @justenwalker. LGTM and I'm going to merge it in and then cut a hopefully-finally RC.

PR #84, opened 2016-02-24
Add --force flag to container tagging in makefile

tgross:
During releases, if there are existing local builds of the example containers then the tags won't be pushed to the Hub until the old builds are cleared up locally. This overrides this behavior. (The local copies of example container images aren't something we care about as build artifacts in-and-of-themselves.)


justenwalker:
LGTM

PR #86, opened 2016-02-24
Update package path for LDFLAGS

tgross:
For https://github.com/joyent/containerbuddy/issues/85

We need to have the fully-qualified path in the LDFLAGS in order to have it injected into the package's variables.

@misterbisson or @justenwalker to review please.


misterbisson:
makes sense to me

justenwalker:
Sorry about this @tgross. I think we might make an integration test for this to catch any regressions in the future?

tgross:
@justenwalker yeah, good idea. I'll do that.

Issue #88, opened 2016-02-26
Implement log levels


tgross:
Containerbuddy is a bit chatty. Health checks in particular pile up in the Docker logs. Adding (and documenting) log levels wouldn't be a bad idea. A couple options:

- https://github.com/Sirupsen/logrus is used by Docker
- I've also used https://github.com/op/go-logging in other projects

justenwalker:
My vote is for Logrus - i've used it in a couple of internal projects and it seems to be well-used among other golang projects i've seen.

Another thing that would be helpful for logging regardless of the framework: always log Containerbuddy stuff to STDERR to keep it distinct from any application-level logs.

tgross:
> My vote is for Logrus

Ok, let's do it!

> Another thing that would be helpful for logging regardless of the framework: always log Containerbuddy stuff to STDERR to keep it distinct from any application-level logs.

Does this change anything unexpected in the behavior of the Docker log drivers?

justenwalker:
> Does this change anything unexpected in the behavior of the Docker log drivers?
I don't believe so, but I only use `json-file`. I'm unsure about [other log drivers](https://docs.docker.com/engine/admin/logging/overview/). My assumption is that most of them just collapse both streams into whatever protocol they are supporting - perhaps marking each line as coming from stderr or stdout.

tgross:
Ok. For configuring it in the Containerbuddy config, what do we think about doing an associative array as a top-level attribute so that it's easier to add fields later in a backwards-compatible way if we've forgotten something?

```json
{
  "consul": "consul:8500",
  "logging": { "level": "INFO" }
}
```

Another question is whether we treat the addition of a log-level configuration option as a backwards-compatible change (i.e. minor version bump) or consider the change in the log format to be a non-backwards compatible  change (i.e. major version bump). I'm ok with not being a super-stickler over something like log formats, but we should decide deliberately.

justenwalker:
> For configuring it in the Containerbuddy config, what do we think about doing an associative array

Sounds like a good idea

> Another question is whether we treat the addition of a log-level configuration option as a backwards-compatible change (i.e. minor version bump) or consider the change in the log format to be a non-backwards compatible change (i.e. major version bump)

This is probably more critical for libraries than for applications - but my gut says its a Minor version update since its an added feature that shouldn't be backwards incompatible, except perhaps in the [XKCD](https://xkcd.com/1172/) sense.

justenwalker:
[...]

PR #93, opened 2016-03-01
Implement Log Levels with Logrus

justenwalker:
For #88

I tried to keep this as backwards compatible as possible by default - basically by delegating the default log formatter to the built-in log package targeting Stdout.

- Add Logrus as the logging framework
- Create "logging" config and add to README
- Add backwards-compatible "go" log format as the default
- Support "text" and "json" output formats
- Support "stderr" and "stdout" output target (Default stdout)

Since the output is going to a non TTY source, the TextFormatter is just a list of key=value entries - Not especially human readable. I suppose people could continue to use the default to the "go" format for a human readable output.

Also, Logrus has a "logstash" output formatter - although I'm not so sure we want such a specialized formatter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,13 +134,13 @@ Must supply only one of the following
     - `endpoints` is the list of etcd nodes in your cluster
     - `prefix` is the path that will be prefixed to all service discovery keys. This key is optional. (Default: `/containerbuddy`)
 
-Logging Config:
+Logging Config (Optional):
 
 The logging config adjust the output format and verbosity of Containerbuddy logs.
 
-- `level` adjusts the verbosity of the messages output by containerbuddy. Must be one of: DEBUG, INFO, WARN, ERROR, FATAL, PANIC (Default: INFO)
-- `format` adjust the output format for log messages. Can be "go", "text", or "json" (Default: "go")
-- `output` picks the output stream for log messages. Can be "stderr" or "stdout" (Default: "stdout")
+- `level` adjusts the verbosity of the messages output by containerbuddy. Must be one of: `DEBUG`, `INFO`, `WARN`, `ERROR`, `FATAL`, `PANIC` (Default is `INFO`)
+- `format` adjust the output format for log messages. Can be `go`, `text`, or `json` (Default is `go`)
+- `output` picks the output stream for log messages. Can be `stderr` or `stdout` (Default is `stdout`)
 
 Other fields:
 
--- a/README.md
+++ b/README.md
@@ -139,8 +139,47 @@ Logging Config (Optional):
 The logging config adjust the output format and verbosity of Containerbuddy logs.
 
 - `level` adjusts the verbosity of the messages output by containerbuddy. Must be one of: `DEBUG`, `INFO`, `WARN`, `ERROR`, `FATAL`, `PANIC` (Default is `INFO`)
-- `format` adjust the output format for log messages. Can be `go`, `text`, or `json` (Default is `go`)
+- `format` adjust the output format for log messages. Can be `default`, `text`, or `json` (Default is `default`)
 - `output` picks the output stream for log messages. Can be `stderr` or `stdout` (Default is `stdout`)
+
+Logging Format Examples:
+
+`default` - go log package with [LstdFlags](https://golang.org/pkg/log/)
+
+```
+2015/03/26 01:27:38 Started observing beach
+2015/03/26 01:27:38 A group of walrus emerges from the ocean
+2015/03/26 01:27:38 The group's number increased tremendously!
+2015/03/26 01:27:38 Temperature changes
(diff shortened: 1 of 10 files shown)


justenwalker:
Also, I took a first pass at actually using more specific log levels - please let me know if they make sense.  My feeling is there might be some decisions to make around Debug vs Info, and Error vs Warn

README.md
```diff
@@ -129,6 +134,14 @@ Must supply only one of the following
     - `endpoints` is the list of etcd nodes in your cluster
     - `prefix` is the path that will be prefixed to all service discovery keys. This key is optional. (Default: `/containerbuddy`)
 
+Logging Config:
+
+The logging config adjust the output format and verbosity of Containerbuddy logs.
+
+- `level` adjusts the verbosity of the messages output by containerbuddy. Must be one of: DEBUG, INFO, WARN, ERROR, FATAL, PANIC (Default: INFO)
+- `format` adjust the output format for log messages. Can be "go", "text", or "json" (Default: "go")
```
tgross: What do you think about including a link to the descriptions for these formats? These names are logrus-specific and "Go developers who use logrus" aren't the only audience for Containerbuddy.

containerbuddy/logging.go
```diff
@@ -0,0 +1,92 @@
...
+	Output string `json:"output,omitempty"`
+}
+
+var defaultLog = &LogConfig{
+	Level:  "INFO",
+	Format: "go",
+	Output: "stdout",
+}
+
+func init() {
```
tgross: So this makes sure we have some kind of sane logging to stdout even if the config file doesn't parse? Nice.

containerbuddy/signals.go
```diff
@@ -55,33 +56,33 @@ func terminate(config *Config) {
...
 
 func reloadConfig(config *Config) *Config {
 	signalLock.Lock()
 	defer signalLock.Unlock()
-	log.Printf("Reloading configuration.\n")
+	log.Infof("Reloading configuration.\n")
 	newConfig, err := loadConfig()
 	if err != nil {
-		log.Printf("Could not reload config: %v\n", err)
+		log.Warnf("Could not reload config: %v\n", err)
```
tgross: I'm going to nitpick and suggest this could be at the Error level, given that if this error is thrown the behavior of Containerbuddy will start to converge from what the user expects.

README.md
```diff
@@ -129,6 +134,14 @@ Must supply only one of the following
     - `endpoints` is the list of etcd nodes in your cluster
     - `prefix` is the path that will be prefixed to all service discovery keys. This key is optional. (Default: `/containerbuddy`)
 
+Logging Config:
+
+The logging config adjust the output format and verbosity of Containerbuddy logs.
```
tgross: We should probably note that this entire stanza is optional.

README.md
tgross: Should we even call the default "go"?

tgross:
Overall looks great, @justenwalker. I had a couple of nitpicks around documentation. Can we maybe get a unit test for the init section?

containerbuddy/logging.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	switch strings.ToLower(l.Output) {
+	case "stderr":
+		output = os.Stderr
+	case "stdout":
+		output = os.Stdout
+	default:
+		return fmt.Errorf("Unknown output type '%s'", l.Output)
+	}
+	log.Printf("=== SET FORMAT ===\n%#v\n", formatter)
```
justenwalker: This was some debug code. I'll remove this

containerbuddy/logging.go
justenwalker: yup :)

Makefile.docker
```diff
@@ -28,7 +28,7 @@ lint:
 	${CDWD} && go fmt ${NO_VENDOR}
 	golint ${SRC}
 
-test:
+test: vendor
```
tgross: This is triggering a `godep restore` for me on every test run; the vendor folder from the host should be available from `DOCKERMAKE` in the main makefile. If we are going to assert this (which seems reasonable) it should be done in the main makefile's `test` target.

tgross: Actually, I'm thinking that we should make sure that we're asserting the dependency here in only one makefile to avoid confusion. I'll move this item into https://github.com/joyent/containerbuddy/pull/95.

justenwalker: Good point. Maybe I'll just remove this and address it in another PR.  I think you were already doing some makefile changes anyway - so I'll avoid a clash.

tgross: Done.

tgross:
Ok, this LGTM once we get the AOK from TravisCI.

tgross:
Merging!

PR #94, opened 2016-03-01
Merge examples into integration tests

tgross:
For https://github.com/joyent/containerbuddy/issues/91. I've moved the example application into a pair of integration tests for the discovery services. This ends up being (nginx + app + consul) and (nginx + app + etcd). The tests stand up the stack, wait for it to converge, and then test that Nginx has the route we expect to `app/`. Changes this required:

- move the test.sh file we were using for unit tests into the source directory
- add both Consul and Etcd configuration for the App and Nginx fixtures
- added a new test runner to the `test_probe` fixture.

cc @misterbisson and @justenwalker


makefile
```diff
@@ -50,6 +50,8 @@ docker: build/containerbuddy_build consul etcd
 build: docker
 	${DOCKERMAKE} build
 
+build/containerbuddy: build
```
tgross: Neither this change nor the way it was seems to be working the way I'd expect with respect to rebuilds. Given that this is orthogonal to the point of this PR, I'm going to back this change out and fix it under separate issue.

tgross:
> Since you're moving the test.sh anyway - do you think that's a worthwhile pattern to follow?

Works for me. Done.

tgross:
After merging https://github.com/joyent/containerbuddy/pull/95 I've had to update to remove the `-v` mount for the examples dir.

tgross:
I've added the change per @justenwalker's suggestion and TravisCI is happy, so going to merge this in.

PR #95, opened 2016-03-01
Minimize rebuilds in makefile

tgross:
This PR takes advantage of make dependency tracking and that fact that `--link` doesn't require
the link be *valid* in order to reduce rebuilds of prereq targets and extra targets during integration test runs.

- `build` is a target with only dependencies; build/containerbuddy always ends up being newer than the build/ dir otherwise and we always rebuild.
- `integration` only requires the finished Containerbuddy binary, not etc or consul; the integration tests stand up discovery as-needed
- `lint` only requires the builder, not etcd/consul


tgross:
>  that fact that --link doesn't require the link be valid

This actually turns out to be wrong when I try out-of-order targets. Lemme fix this and take care of my comment from https://github.com/joyent/containerbuddy/pull/93/files#r54612339 as well.

tgross:
Ok fixed by splitting out `DOCKERMAKE` into `DOCKERBUILD` (non-test) and `DOCKERRUN` (tests). My verification process:

- run `make clean`
- run `make lint`: will create builder, do a `godep restore`, and then lint
- run `make lint` again: will just lint, without rebuilding builder or `godep restore`
- run `make test`: will start consul/etcd and then run unit tests, without rebuilding or `godep restore`
- run `make integration`: will run build without `godep restore`, and then run ./test.sh and build test fixtures
- run `make clean` again
- run `make build`: will create builder, do a `godep restore`, run the build

If we merge this prior to merging https://github.com/joyent/containerbuddy/pull/94 then I'll want to throw a commit into #94 removing the `-v` mount for examples.

containerbuddy/consul.go
```diff
@@ -3,8 +3,8 @@ package containerbuddy
 import (
 	"fmt"
 	"log"
-	"sort"
 	"os"
+	"sort"
```
tgross: fixed from a `make lint` test

tgross:
Added an important bit which allows `make build` to work without removing the Containerbuddy binary whenever one of our golang source files change.

justenwalker:
I gave this a run-through with your steps and it LGTM.  Nice work - will cut down on iteration time :+1:

PR #96, opened 2016-03-02
Remove example from makefile too

tgross:
@misterbisson @justenwalker I forgot to remove the example application from the makefile too.


justenwalker:
LGTM

PR #97, opened 2016-03-02
etcd: parse array endpoints correctly

### LINKED ISSUE #92: etcd config does not accept list of endpoints
Ran into this while working on https://github.com/joyent/containerbuddy/issues/91. The etcd configuration is supposed to accept a list of endpoints as documented in the README:

```json
{
  "etcd": {
    "endpoints": ["http://etcd1:4001"]
  }
}
```

But this throws the error `Must provide etcd endpoints`, which comes from [here where we switch over the config values](https://github.com/joyent/containerbuddy/blob/e392ab6b627f032822c80c88f40b9fc20b737121/containerbuddy/etcd.go#L43). Passing single string for a single host works fine. [This section of our etcd tests](https://github.com/joyent/containerbuddy/blob/e392ab6b627f032822c80c88f40b9fc20b737121/containerbuddy/etcd_test.go#L9-L11) also works, which suggests that the problem is in the configuration file parsing and not the switch where we're throwing the error.

justenwalker:
Fixes #92 

- Extract json marshaling into etcd
- Ensure correct handling of []interface{} endpoints
- Create unit tests for parsing both string and array endpoints

### DIFF
--- a/containerbuddy/config.go
+++ b/containerbuddy/config.go
@@ -39,15 +39,15 @@ func getConfig() *Config {
 
 // Config is the top-level Containerbuddy Configuration
 type Config struct {
-	Consul       string                 `json:"consul,omitempty"`
-	Etcd         map[string]interface{} `json:"etcd,omitempty"`
-	LogConfig    *LogConfig             `json:"logging,omitempty"`
-	OnStart      json.RawMessage        `json:"onStart,omitempty"`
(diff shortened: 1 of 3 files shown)


tgross:
I bet there's a JSON parsing library outside the stdlib that uses reflection and so sucks from a performance standpoint but doesn't require you jump thru silly hoops like our configuration syntax keeps making us do. :grinning: 

LGTM though. Merging and thanks!

PR #100, opened 2016-03-03
Add applications-on-autopilot blog link to examples

tgross:
@misterbisson this adds a link to our updated primary example with walkthru.


misterbisson:
:pickles:

PR #103, opened 2016-03-04
support setting the ip that will be advertised explicitly

wleese:
..in order to support bridged mode docker users.

https://github.com/joyent/containerbuddy/issues/102

### DIFF
--- a/containerbuddy/config.go
+++ b/containerbuddy/config.go
@@ -65,9 +65,8 @@ type ServiceConfig struct {
 	TTL              int             `json:"ttl"`
 	Interfaces       json.RawMessage `json:"interfaces"`
 	Tags             []string        `json:"tags,omitempty"`
-	Ip               string          `json:"ip,omitempty"`
+	IpAddress        string          `json:"ipaddress,omitempty"`
 	discoveryService DiscoveryService
-	ipAddress        string
 	healthCheckCmd   *exec.Cmd
 }
 
@@ -312,8 +311,8 @@ func initializeConfig(config *Config) (*Config, error) {
 			return nil, ifaceErr
 		}
 
-		if service.Ip == "" {
-			if service.ipAddress, err = GetIP(interfaces); err != nil {
+		if service.IpAddress == "" {
+			if service.IpAddress, err = GetIP(interfaces); err != nil {
 				return nil, err
 			}
 		}
--- a/containerbuddy/config.go
+++ b/containerbuddy/config.go
@@ -65,6 +65,7 @@ type ServiceConfig struct {
 	TTL              int             `json:"ttl"`
 	Interfaces       json.RawMessage `json:"interfaces"`
 	Tags             []string        `json:"tags,omitempty"`
+	Ip               string          `json:"ip,omitempty"`
 	discoveryService DiscoveryService
 	ipAddress        string
 	healthCheckCmd   *exec.Cmd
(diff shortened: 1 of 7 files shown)


containerbuddy/config.go
```diff
@@ -311,8 +312,10 @@ func initializeConfig(config *Config) (*Config, error) {
 			return nil, ifaceErr
 		}
 
-		if service.ipAddress, err = GetIP(interfaces); err != nil {
-			return nil, err
+		if service.Ip == "" {
```
tgross: We're setting the new `Ip` field on the `ServiceConfig` struct but then that field isn't being consumed anywhere. So this will let us get the raw value from the config file but then we're not doing anything with it.

tgross:
@wleese I've got the implementation comment about consuming the `Ip` field, but we'll need two more things before we can move this forward:
- we need to have tests exercising the behavior
- we need documentation about the design of the feature and its intended behavior (along with the caveat that it's not the primary use case)

wleese:
Thanks for your patience.

> we need to have tests exercising the behavior

I moved away from the dedicated ip field and instead reused the existing ipAddress field and exposed it by capitalization. That makes things more intuitive but looking at the existing tests for inspiration, doesn't leave me with any obvious places I could provide additional tests.

tgross:
> I moved away from the dedicated ip field and instead reused the existing ipAddress field and exposed it by capitalization.

I think that's the right approach.

> looking at the existing tests for inspiration, doesn't leave me with any obvious places I could provide additional tests.

There's a new category of user error that this feature introduces, which is adding an invalid IP address to the configuration. We should be validating this input and ensuring we handle it appropriately. So it looks like we need some code to make sure we're passing the IP address thru functions provided in [`ips.go`](https://github.com/joyent/containerbuddy/blob/master/containerbuddy/ips.go) to test that it parses as a valid IPv4 or IPv6 address. And we'll need a unit test that exercises that code path.

containerbuddy/config.go
```diff
@@ -311,8 +312,15 @@ func initializeConfig(config *Config) (*Config, error) {
...
 		}
 
-		if service.ipAddress, err = GetIP(interfaces); err != nil {
-			return nil, err
+		if service.IpAddress == "" {
+			if service.IpAddress, err = GetIP(interfaces); err != nil {
+				return nil, err
+			}
+		} else {
+			if err := net.ParseIP(service.IpAddress); err == nil {
```
tgross: The test you've written exercises the parsing of the config file but doesn't exercise this new code. This line should read `if err := net.ParseIP(service.IpAddress); err != nil {` and this wasn't caught by the tests.

wleese:
I'm confused. Docs say:" If s is not a valid textual representation of an IP address, ParseIP returns nil". My intention was to catch non-ip's and return an error.

Regarding the test, I'll rethink it.

tgross:
> I'm confused. Docs say:" If s is not a valid textual representation of an IP address, ParseIP returns nil". My intention was to catch non-ip's (so == nil) and return an error.

Ok. The name `err` is idiomatically always an `error` though... so I got confused by you calling it `err` rather than `ip` or something like that, given that [`ParseIP`](https://golang.org/pkg/net/#ParseIP) a [`net/IP` struct](https://golang.org/pkg/net/#IP). So what you're doing is technically correct but very misleading to the reader.

tgross:
Ok, this looks good. Thanks for the contribution, @wleese!

PR #107, opened 2016-03-10
Revert "support setting the ip that will be advertised explicitly"

tgross:
@misterbisson this reverts joyent/containerbuddy#103


tgross:
The plan is to release this as 1.2.1. Although this *technically* breaks semver, 1.2.0 has been out literally hours at most and my justification is that my reverting this we're fixing a bug (broken feature) as opposed to removing a feature. No one will have built anything on top of the code we're reverting here.

misterbisson:
Lgtm

misterbisson:
The limitations of the original patch are documented in https://github.com/joyent/containerbuddy/issues/102#issuecomment-194810938

PR #108, opened 2016-03-10
allow setting the service discovery endpoint

wleese:
A second go at https://github.com/joyent/containerbuddy/pull/103 but now it completely resolves my use case.

Need to add tests, but just wanted to pitch this first.


misterbisson:
@wleese I've got some questions in https://github.com/joyent/containerbuddy/issues/102#issuecomment-194936317 that I'd like to go through with you related to this. The summary is that I'm very cautious about scope expansion, so we need explore the problem further to see what solution is actually required.

containerbuddy/config.go
```diff
@@ -65,6 +66,7 @@ type ServiceConfig struct {
 	TTL              int             `json:"ttl"`
 	Interfaces       json.RawMessage `json:"interfaces"`
 	Tags             []string        `json:"tags,omitempty"`
+	Endpoint         string          `json:"endpoint,omitempty"`
```
misterbisson: The `endpoint` name seems ambiguous or possibly misleading. In this context, it's the hostname that you want Containerbuddy to resolve to an IP before registering the service in Consul.

containerbuddy/config.go
```diff
@@ -311,8 +313,19 @@ func initializeConfig(config *Config) (*Config, error) {
 			return nil, ifaceErr
 		}
 
-		if service.ipAddress, err = GetIP(interfaces); err != nil {
-			return nil, err
+		if service.Endpoint == "" && service.ipAddress ==  "" {
+			if service.ipAddress, err = GetIP(interfaces); err != nil {
```
misterbisson: I have concerns about doing name resolution in Containerbuddy. It's actually a very significant expansion of scope for the code.

Name resolution is hardly new technology, but the failure modes are often misunderstood or underestimated. Granted, my concerns stem from some deep work I did recently on CNS (https://www.joyent.com/blog/introducing-triton-container-name-service). Among those concerns:

1. The DNS TTL here cannot be honored
1. If we try to honor the TTL, how?
1. Multiple IPs may be returned for this A or AAA record, is it safe to assume only the first one should be used?

containerbuddy/config.go
```diff
@@ -311,8 +313,19 @@ func initializeConfig(config *Config) (*Config, error) {
...
+				return nil, err
+			}
+		} else {
+			if ok := net.ParseIP(service.Endpoint); ok == nil {
+				if ip, err := net.LookupHost(service.Endpoint); err == nil {
```
misterbisson: I have concerns about doing name resolution in Containerbuddy. It's actually a very significant expansion of scope for the code.

Name resolution is hardly new technology, but the failure modes are often misunderstood or underestimated. Granted, my concerns stem from some deep work I did recently on CNS (https://www.joyent.com/blog/introducing-triton-container-name-service). Among those concerns:

1. The DNS TTL here cannot be honored
1. If we try to honor the TTL, how?
1. Multiple IPs may be returned for this A or AAA record, is it safe to assume only the first one should be used?

tgross: Wanted to add here that the `ipAddress` field is just a string, and that's what we're giving to Consul in the `/v1/agent/service/register/` call. If we're allowing hostnames at all, why not just pass them through as-is to Consul rather than trying to look them up here. Doing a lookup is a misleading validation, because it can go away afterwards. And Containerbuddy doesn't do anything with the addresses other than give them to Consul anyways, whereas if you're going down this path then presumably your application has some mechanism to deal with DNS.

Also, from a code standpoint you never assign the service address if it parses correctly as an IP address, which means you're allowing the end user to pass in a DNS name but not an IP. Which probably wasn't the intent.

containerbuddy/config.go
wleese: Hostname or IP, which is why I could use neither. Any suggestions?

containerbuddy/config.go
wleese: > If we're allowing hostnames at all, why not just pass them through as-is to Consul rather than trying to look them up here.

Good idea, I'll rewrite once I have a clue what to call the parameter.

>  you never assign the service address if it parses correctly as an IP

fixed.

wleese:
Any objections left for this one?

misterbisson:
@wleese this is straightforward enough, and news that you're using it is good. It needs tests, though, yes?

We're doing some refactoring in https://github.com/joyent/containerbuddy/pull/119 and https://github.com/joyent/containerbuddy/pull/123 that we hope to finish very soon. Once those land, would you mind refactoring for that and adding tests? That'll make this ready for merging.

misterbisson:
@wleese the other PRs are merged. Can you update this PR and add tests?

justenwalker:
@tgross @wleese I'm wondering if this could be folded into interfaces so that we don't have yet another config value for IPs.

Something like this:
```json
{
  "interfaces": ["static:10.1.2.3"],
}
```

tgross:
> I'm wondering if this could be folded into interfaces so that we don't have yet another config section for IPs.

Not a bad idea.

wleese:
@justenwalker I know this sounds silly, but what if someone named their interface 'static'?

justenwalker:
> @justenwalker I know this sounds silly, but what if someone named their interface 'static'?

Theoretically that could pose a problem - in the same way that earth may be hit with a meteor and all life may become extinct at any moment. But as with either case - I'm not sure we should worry about it.

wleese:
..but does that still make that a better solution than a dedicated field which has different behaviour anyway? I don't care much either way, it just seemed to be a change that has it's own downsides.

justenwalker:
> ..but does that still make that a better solution than a dedicated field which has different behaviour anyway? I don't care much either way, it just seemed to be a change that has it's own downsides.

An additional field that configures the same thing seems like it has it's own drawbacks - specifically it overrides the behavior of the the `interfaces` field and therefore ignores it.

Perhaps we can print a warning if the user supplies both of these to indicate that we are ignoring `interfaces`, but I was thinking that it would fit nicely into the `interfaces` config since I could do something like:

```json
{ 
  "interfaces": [
    "ethwe",
    "static:10.0.0.1"
  ]
}
```

Which means, use `ethwe` if it exists, otherwise use the static IP provided.

tgross:
I like @justenwalker's proposed approach for all the reasons he's suggested. There's not really a reason to worry about someone naming an interface `static` here, because it'll always be followed by `:<some ip>`. `static` on it's own would be a valid interface spec if and only if there was an interface named `static` *and* the IP was omitted.

justenwalker:
> There's not really a reason to worry about someone naming an interface static here, because it'll always be followed by :<some ip>. static

One wrinkle is that a `:` is a valid character in an interface name when you bring up [multiple IPs on the same device](http://linux-ip.net/html/tools-ifconfig.html#tools-ifconfig-conclusion)
However that number will always be a non-negative integer, not an IP address.

tgross:
> However, it will always be followed by a non-negative integer, not an IP address.

That doesn't sound too terrible to deal with; our parsing logic is a little complex but it's not like it's on a hot code path -- it typically happens once per lifetime of the container.

utils/ips.go
```diff
@@ -122,6 +127,18 @@ type inetInterfaceSpec struct {
...
 
+// -- matches static
+type staticInterfaceSpec struct {
+	Spec string
+	Name string
+	IP net.IP
+}
+
+func (s staticInterfaceSpec) Match(index int, iip interfaceIP) bool {
+	// Never matches
```
justenwalker: What is the reason why this never matches?

It looks like the logic for this spec is in the inetInterfaceSpec instead.

utils/ips.go
```diff
@@ -187,6 +204,14 @@ func parseInterfaceSpec(spec string) (interfaceSpec, error) {
 	if spec == "inet6" {
 		return inetInterfaceSpec{Spec: spec, Name: "*", IPv6: true}, nil
 	}
+	if strings.HasPrefix(spec, "static:") {
+		ip := strings.SplitAfter(spec, "static:")
+		nip := net.ParseIP(ip[1])
+		if nip == nil {
+			return nil, fmt.Errorf("Unable to parse static ip %s in %s", ip[0], spec)
```
justenwalker: This doesn't seem to work for `static:0` which would be a valid interface.

wleese: aliases aren't supported properly yet, if I understand this regex correctly:
```
^(?P<Name>\w+)(?:(?:\[(?P<Index>\d+)\])|(?::(?P<Version>inet6?)))?$
```

justenwalker: Agreed.  Just don't want the static IP parser to trap potentially valid interfaces.

tgross:
I'm happy with this at this point. Thanks for your patience with this @wleese -- I think us all spending a little time making it right will be better for the project in the long run.

LGTM.

wleese:
@tgross not at all, thank you for the guidance.

PR #113, opened 2016-03-24
Parse scheme from Consul configuration string

tgross:
For https://github.com/joyent/containerbuddy/issues/112.

Users who are securing communication with Consul via TLS should be able to include the scheme in the URI provided and have that picked up by the config constructor. By parsing it out of the string rather than forcing the user to use a structure config like the Etcd backend, we can add support for TLS without breaking backwards compatibility. This means that all the following are valid Consul config entries:

```json
{
  "consul": "consul:8500",
  "consul": "https://consul:8500",
  "consul": "http://consul:8500",
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ Service Discovery Backends:
 
 Must supply only one of the following
 
-- `consul` configures discovery via [Hashicorp Consul](https://www.consul.io/). Expects `hostname:port` string:
+- `consul` configures discovery via [Hashicorp Consul](https://www.consul.io/). Expects `hostname:port` string. If you are communicating with Consul over TLS you may include the scheme (ex. `https://consul:8500`):
 
     ```
     "consul": "consul:8500"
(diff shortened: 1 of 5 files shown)


tgross:
The integration test for both this PR and unrelated #108 are failing at the same point but master, which suggests something is up with the test environment. I can replicate the problem locally so digging into that.

tgross:
The problem is that our `test_probe` integration test fixture is failing to build because we're not pinning the Consul dependency the same way we are in the containerbuddy application. This results in the following error during compilation because their master is broken:

> src/github.com/hashicorp/consul/consul/prepared_query/template.go:119: ct.re.Copy undefined (type *regexp.Regexp has no field or method Copy)

The fix for this will be to correctly pin our dependencies as we have for the main app. Will do.

tgross:
I've rebased this onto the fix in https://github.com/joyent/containerbuddy/pull/114 and now the integration tests pass. Should be good-to-go for review.

misterbisson:
:house_with_garden: :walking:

PR #114, opened 2016-03-25
Pin dependencies for test_probe fixture

tgross:
@misterbisson @justenwalker this pins the dependencies for the test_probe.go integration test fixture. This should prevent the Consul API dependency (which is currently broken at the HEAD of master) from breaking our integration tests. ref https://github.com/joyent/containerbuddy/pull/113#issuecomment-201021419


misterbisson:
:house_with_garden: :walking:

Issue #109, opened 2016-03-11
Deprecate "onStart" in favor of "preStart"


tgross:
We've had one issue on GitHub (https://github.com/joyent/containerbuddy/issues/98) and a couple of Joyent customer issues that ultimately came down to the name of the `onStart` flag. The problem seems to be confusing `onStart` with the main application that Containerbuddy is shimming. I'd like to propose that we rename the `onStart` option `preStart` to make it more clear. Because this breaks backwards compatibility I'd like to propose the following deprecation process:

Version 1.x.0:
- Alias the `onStart`/`preStart` option to the same behavior, with `preStart` winning out if both are configured.
- Remove `onStart` from the documentation in favor of `preStart`.
- Have the use of `onStart` log at WARN level: "The onStart option has been deprecated in favor of preStart. Containerbuddy will use the onStart option as a preStart."
- Have the simultaneous use of `onStart` and `preStart` log at the WARN level: "The onStart option has been deprecated in favor of preStart. Containerbuddy will use only the preStart option provided."

Version 2.x (future):
- Remove the `onStart` option handling entirely

cc @misterbisson and @justenwalker for their thoughts, particularly on the wording of the warnings.

justenwalker:
@tgross Seems reasonable and would maintain backwards compatibility to support both.  Also more inline with `preStop` wording that already exists.

It it worth pointing the user back to the documentation as part of the `WARN` message? Or reiterating the purpose of the `preStop` command?

```
preStop is the executable (and its arguments) that will be called immediately prior to starting the shimmed application. If the preStop handler returns a non-zero exit code, Containerbuddy will exit.
```

misterbisson:
That's a sound approach, I'm all for it.

We'll also need to update examples and documentation in the following:

- https://www.joyent.com/blog/applications-on-autopilot
- https://www.joyent.com/blog/containerbuddy-1.0
- https://www.joyent.com/blog/dbaas-simplicity-no-lock-in
- http://autopilotpattern.io/
- http://containerbuddy.io / https://github.com/joyent/containerbuddy/blob/master/README.md

As a related aside, this is interesting: https://www.google.com/search?client=safari&rls=en&q=%22onstart%22+containerbuddy&ie=UTF-8&oe=UTF-8

tgross:
> Is it worth pointing the user back to the documentation as part of the WARN message? Or reiterating the purpose of the preStart command?

Yeah, that sounds good.

tgross:
[...]

PR #115, opened 2016-03-25
Deprecate onStart in lieu of preStart

tgross:
For https://github.com/joyent/containerbuddy/issues/109

To improve the clarity of configuration, we're changing the `onStart` option to `preStart`.

For purposes of backwards compatibility the deprecation behavior is as follows. If `onStart` is provided without `preStart`, warn the user and alias the `onStart` to a `preStart`. If both `onStart` and `preStart` are provided, warn the user and use only the `preStart`.

#109 proposed logging at WARN but this section of the code is prior to the logging configuration being setup so it will simply send to stdout like our other pre-logging-config messages.


misterbisson:
:house_with_garden: :walking:

PR #118, opened 2016-03-25
Refactor of Pollable for modularization effort

tgross:
Making PollAction a method of Pollable will allow us to avoid a circular import when we split out packages from the main containerbuddy package.

@justenwalker this follows from the design I was describing in https://github.com/joyent/containerbuddy/issues/83#issuecomment-201479922, which I think will make building the metrics and tasks packages easier and make our life easier when we split up the fat containerbuddy package into smaller bits. (Note that this only required changing one test, which relied on binding variables from the enclosing scope in `poll`, so I'm reasonably confident this is the minimal intervention we can make.)

cc @misterbisson


tgross:
I've also moved some of the parsing functions into their own package so that I can share them with packages that implement Pollable but want to validate their inputs (which I suspect all our Pollable-implementing packages will want to do).

tgross:
This is safe to merge and release as a point release, as we've added new public APIs to our package (not that anyone is using the `containerbuddy` or `utils` package except us) but haven't eliminated anything and it's feature-identical. Did you have any thoughts on the design here, @justenwalker ?

justenwalker:
@tgross I haven't been able to pull this branch and play with it yet, however extracting out the utils package seems like a good idea.  Also the PollAction seems like an appropriate abstraction - so from that standpoint is looks good.

I'm not sure how I'll be using it in the scheduled tasks, so I'd like to see if I can rebase on it and adapt what I have.

tgross:
> I'm not sure how I'll be using it in the scheduled tasks, so I'd like to see if I can rebase on it and adapt what I have.

Sounds good. Let's hold off merging this until we've both had a chance to try it out in the features we're working on.

tgross:
Alternately, if we decide we're happy with https://github.com/joyent/containerbuddy/pull/123 that'll merge this as well.

tgross:
I'm going to close this PR because it's covered by #123

Issue #27, opened 2015-12-03
Proposal: add user defined application telemetry


misterbisson:
This is poorly thought through so far, but I wanted to open up discussion for it.

Health checking gives us a binary way to determine if the app is working or not. If the app is not healthy, we should not send any requests to it and we most likely need to spawn a replacement. But scaling depends on more than binary app health. Every app has one or more performance indicators that can be used to determine if the app is nearing overload and should be scaled up or is too lightly loaded and should be scaled down. In the spirit of what Containerbuddy does to make applications container-native, I think it might also make sense to add an awareness of those performance indicators.

Some performance indicators can be read from the system. The [five minute load average](https://en.wikipedia.org/wiki/Load_%28computing%29) reported by the kernel may work well for many apps, but many of the most interesting indicators come from the app itself.

In MySQL, the number of `Query` entries from [`SHOW PROCESSLIST`](http://dev.mysql.com/doc/refman/5.7/en/show-processlist.html) that are in any [`Waiting`](http://dev.mysql.com/doc/refman/5.7/en/general-thread-states.html) state can be a very significant performance indicator, and it's one that is best retrieved from within the container.

In Nginx, the average [request processing time](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_time) is useful info, but that's only output in the logs, which are hopefully not inside the container (if they were, [ngxtop](https://github.com/lebinh/ngxtop) would be a nice tool to help understand them). But we do expect [`http_stub_status_module`](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html) in our [triton-nginx image](https://github.com/tgross/triton-nginx), so we can look at that instead. `Active connections` vs. the [`worker_connections`](http://nginx.org/en/docs/ngx_core_module.html#worker_connections) limit is a hugely important number there. Any delta between `accepts` and `handled` is a huge red flag. Or, perhaps, the `Waiting` number is an inverse indicator (high numbers indicate low activity, low numbers are high activity, zero could be critical). (See https://github.com/tgross/triton-nginx/issues/3 for further musing on this.)

Here's what I'm imagining it would look like in the config (though it would be obviously unlikely to actually mix Nginx and MySQL in a single image):

```json
  "kpis": [
    {
      "name": "system-loadavg",
      "poll": 103,
      "kpi": "/opt/containerbuddy/system-loadavg.sh"
    },
    {
      "name": "mysql-queries-waiting",
      "poll": 31,
[...]

PR #119, opened 2016-03-28
[no merge] Initial implementation of telemetry

tgross:
This is my initial pass at #27. ~~Currently integration tests will pass but the metrics feature has no tests itself so those tests just verify that the whole thing builds and won't break if we're not including metrics.~~ (*update: has tests now!*) So this isn't ready to merge but it gives an idea of the approach.

- The Telemetry server will create itself a new ServiceConfig and register with discovery that way.
- Sensors expect only a return value of a string that can be parsed as a float64. Anything else on stdout will cause the metric to be rejected. Stderr will get passed to the container's stderr as per usual
- This is the out-of-the-box Prometheus client config, which includes all the metrics of Containerbuddy itself and is quite a lot of data. We may want to look into that.
- wrote lots and lots of unit tests (and fixed whatever results from that)

Next items TODO:
- ~~maybe some integration tests?~~ going to work up a prometheus blueprint first so I'm ok with leaving that off from this PR.

cc @justenwalker and @misterbisson 
Note that this PR includes all of https://github.com/joyent/containerbuddy/pull/118 currently, so when that's merged this PR should be cleaner.


tgross:
Awesome (er, "awesome")... our zombie reaping code is breaking in the integration test. But not on my machine even after `make clean`.

justenwalker:
Seems like a race.  could be that the threshold needs to be increased or the test needs to be tightened up some more.  The assumption is that the zombies will be reaped in a timely manner - which may not be the case on travis (and maynot be deterministic)

tgross:
> The additional zombie hasn't been re-parented, so of course it won't get reaped.

Good catch, and I think your suggested fix for checking zombies owned only by PID1 makes sense. I'll open a new PR for that and rebase this PR on that so we can review it cleanly.

My next question is how the zombie wasn't reparented, but that's probably out-of-scope for us.

justenwalker:
> My next question is how the zombie wasn't reparented, but that's probably out-of-scope for us (this is a kernel mechanism as far as I know).

I think it hasn't been re-parented because the parent already exists - still sleeping

```
S        1    17 /bin/sleep 2
Z       17    19 [sleep]
```

tgross:
I've added a whole mess of unit tests, and fixed two minor bugs along the way with making sure metrics check commands reset and not noticing some odd prometheus API choices with their collectors. Next big step is to add a README.

tgross:
@misterbisson I've added README updates in this commit https://github.com/joyent/containerbuddy/pull/119/commits/f7c613d4e13b10e44516a65837e2c587831b5447

justenwalker:
@tgross Haven't looked yet. 

But to the README point - it seems like that is getting quite big and will continue to grow.  Is there call for generating some kind of webpage and splitting out the concepts into a more organized format?

tgross:
Opened https://github.com/joyent/containerbuddy/issues/124 to discuss the documentation issue.

tgross:
Had a sidebar conversation with @misterbisson and we're going to rename this feature to "telemetry." Will have a commit to update this PR today.

tgross:
We're going to ship this branch as a release candidate for 1.4.0, so that we can use the implementation in an autopilot pattern we're planning on publishing this week. I'd like to get this and our refactor merged in prior to the actual release.

justenwalker:
@tgross I only was able to spare a couple minutes today, but I did a look over of the code and It all looked good. If you're waiting for my feedback; I'd say its a good RC just to get something out the door and find the actual bugs in the implementation.

In that respect, do you have an example `containerbuddy.json` that I could use to tweak and test locally?

tgross:
> I'd say it's a good RC just to get something out there and find the actual bugs in the implementation - if any.

What I'd like to do then is wait until I've got at least one known-working example (see below) and then merge this to master.

> In that respect, do you have an example containerbuddy.json that I could use to tweak and test locally? With the example shell scripts and program - I guess a docker container on hub would cover it :)

That's what I'm working on now, so I should definitely have something for you tomorrow. I've got https://github.com/autopilotpattern/prometheus as a collector and I'm adding the use case to https://github.com/autopilotpattern/touchbase now.

tgross:
> I'd say it's a good RC just to get something out there and find the actual bugs in the implementation - if any.

What I'd like to do then is wait until I've got at least one known-working example (see below) and then merge this to master.

> In that respect, do you have an example containerbuddy.json that I could use to tweak and test locally? With the example shell scripts and program - I guess a docker container on hub would cover it :)

That's what I'm working on now, so I should definitely have something for you tomorrow. I've got https://github.com/autopilotpattern/prometheus as a collector and I'm adding the use case to https://github.com/autopilotpattern/touchbase now.

tgross:
I've pushed up an integration test that exercises a couple of bugs I found.

tgross:
Comments from @misterbisson in https://github.com/autopilotpattern/touchbase/pull/18:

> Should the URL and name be user-configured options? My reason for asking is how discovery would work in an environment where there may be different for each application.

> Proposal: remove `name`, `url`, `poll`, and `ttl` as user-configurable options, so that the `telemetry` stanza is just the following:

> ```json
  "telemetry": {
    "port": 9090,
    "sensors": [
      {...}
    ]
  }
```

Just so that it's clear, the `name`, `poll`, and `ttl` fields are necessary to advertise telemetry as a service. But I think you're asking whether we should define these values and not allow the user to configure them. While appreciate a desire to simplify the configuration syntax, forcing a particular choice on the user is going to limit the usefulness of the feature. The major use case I'm thinking of here is having separate collectors for separate dev/prod/business-unit stacks (Prometheus won't listen for Consul "tags"). Polling requirements are likely to be application specific. Philosophically speaking, it's the responsibility of the application development team to determine the appropriate values for configuring the telemetry service for their application.

That being said, having sane defaults couldn't hurt. In the common case someone can just default to advertising a `telemetry` service on the URL `/telemetry` on port 9090 and having it heartbeat every 10 seconds for 30 sec TTL.

---

Along these same lines, we've mirrored the Prometheus golang API in terms of the sensor naming definitions. They give you the option to have `namespace`, `subsystem`, and `name` for each metric so we copied that dutifully:

```json
    "sensors": [
       {
         "namespace": "tb",
         "subsystem": "nginx",
         "name": "connections_active",
         "help": "Current number of active client connections",
         "type": "gauge",
         "poll": 5,
         "check": ["/opt/containerbuddy/sensor.sh", "connections_active"]
       },
```

But the Prometheus API doesn't actually care about that, and this would be the same thing as leaving out the `namespace` and `subsystem` fields and just naming it `tb_nginx_connections_active`. The golang client's [BuildFQName](https://github.com/prometheus/client_golang/blob/90c15b5efa0dc32a7d259234e02ac9a99e6d3b82/prometheus/metric.go#L99-L119) just concatenates the values with `_` anyways. Presumably they include this to encourage good practices and to make it easier when you're calling the code as a library. But our end users aren't calling into the golang API as a library, they're just giving us strings.

It looks like it's entirely safe with our existing implementation here for an end user to leave off `namespace` and `subsystem`. No code changes required. Maybe we encourage that in the docs (and [demonstration application](https://github.com/autopilotpattern/touchbase/pull/18))

tgross:
> It looks like it's entirely safe with our existing implementation here for an end user to leave off namespace and subsystem. No code changes required.

I've verified that was the case in this new unit test: https://github.com/joyent/containerbuddy/pull/119/commits/168df0dbab2dc40341856e85d168d3c6728be405

misterbisson:
Regarding configuration, we started that discussion in https://github.com/joyent/containerbuddy/issues/27#issuecomment-198631596, where I proposed the following in the context of our typical service definition:

```json
  "services": [
    {
      "name": "containerbuddy-telemetry",
      "port": <user defined>,
      "health": <not a user-configurable value>,
      "interfaces": <user defined>,
      "poll": <not a user-configurable value>,
      "ttl": <not a user-configurable value>,
      "tags": <maybe user defined>
    }
  ],
```

The reason I I feel strongly that `name` should not be user configurable (and why I didn't even suggest `url`) is that having this be identical across all applications is critical for anybody's ability to reasonably build monitoring tooling. @tgross outlines the argument against this:

> The major use case I'm thinking of here is having separate collectors for separate dev/prod/business-unit stacks

However, I believe that argument is inconsistent with what we're arguing elsewhere, which is to isolate dev from prod, etc, using lower level solutions and _not_ sharing the discovery catalog between environments.

tgross:
@misterbisson I've updated per your comments. Do me a favor and review the two READMEs. I'm going to fold these changes back into #123 as well.

tgross:
Marking this `[no merge]`. Ref https://github.com/joyent/containerbuddy/pull/123#issuecomment-207516382

misterbisson:
🏡 🚶

PR #120, opened 2016-03-29
Avoid racing reparent in zombie-reaping integration test

tgross:
At any given time we may have up to 1 zombie parented to PID1 (it has been reparented but not yet reaped) and 1 zombie not yet parented to PID1. We can't test any more precisely than this without racing the kernel reparenting mechanism (`do_exit()`).

Also updated the test to capture the ps tree only once so that we don't log something different from what we tested.

cc @justenwalker per our discussion in https://github.com/joyent/containerbuddy/pull/119


justenwalker:
LGTM

PR #129, opened 2016-04-12
Rename Containerbuddy to ContainerPilot

tgross:
We're renaming Containerbuddy to ContainerPilot to simplify and clarify the relationship between autopilot pattern—the approach to automating our applications—and ContainerPilot—the shared library that makes it easy to build autopilot pattern applications.

@misterbisson this PR renames Containerbuddy to be ContainerPilot.

```
$ grep -rni buddy *
$ echo $?
1
```


misterbisson:
🏡 🚶

PR #132, opened 2016-04-14
Use a pointer receiver on Backend.PollAction

tgross:
If the PollAction does not use a pointer receiver (as we're doing on Service and Sensor), then the Backend's healthCheckCmd will never be reset and the next onChange event will fail with the error "already running".

cc @justenwalker @misterbisson


misterbisson:
:pickles:

Issue #87, opened 2016-02-26
Proposal: Periodic tasks


justenwalker:
Currently, we have `preStart`, `preStop` and `postStart` events; As the container is running though, it may be useful to have periodic tasks execute to report on status to external systems separate from the health checks that report to the service discovery backend.

The primary use-case would be a logical extension point for push-style metrics without having to build in any backends into Containerbuddy directly. (See #27 for discussion)

Configuration may look something like:
```json
{
  "onScheduled": [
    { "frequency": "1s", "command": [ "/bin/push_metrics.sh" ] },
    { "frequency": "10s", "command": [ "/bin/push_other_metrics.sh" ] }
  ]
}
```

Some other things to consider:
- Should we be supporting a more flexible duration syntax? For example: [ISO8601 Durations](https://en.wikipedia.org/wiki/ISO_8601#Durations). For the metrics use-case, the frequency of events will be very short duration. Backups could potentially be longer (hours). Certainly I don't think there would be multi-day pauses between execution.
- Should we build in the idea of exponential back-off to deal with back-pressure? If a command results in a specific non-zero exit code, we reduce the frequency to give the target system time to recover. Note: This may be useless or harmful for larger frequencies (Hours), so we might need to also specify the retry frequency as well in order to fix that.

tgross:
> Should we be supporting a more flexible duration syntax? For example: ISO8601 Durations. For the metrics use-case, the frequency of events will be very short duration. Backups could potentially be longer (hours). Certainly I don't think there would be multi-day pauses between execution.

The architecture we have for Containerbuddy hasn't been particularly optimized for high-performance and forking off lots of processes at subsecond intervals might prove to be costly. I would suggest that if we want to allow for subsecond that we do some performance testing of that design.

> Should we build in the idea of exponential back-off to deal with back-pressure? If a command results in a specific non-zero exit code, we reduce the frequency to give the target system time to recover. Note: This may be useless or harmful for larger frequencies (Hours), so we might need to also specify the retry frequency as well in order to fix that.

This also brings up the question of semantics of the frequency. Under the current design of [`poll`](https://github.com/joyent/containerbuddy/blob/17641310368c320cb798f5665c862542b538ebbf/containerbuddy/main.go#L72-L90) the polling goroutine is blocked by the execution of the `pollingFunc` we give it. This is by design because we don't want (for example) multiple health checks running simultaneously -- if a health check can't return within the TTL then the node should be marked unhealthy anyways.

[...]

PR #134, opened 2016-04-15
Implement Periodic Tasks

justenwalker:
For #87 (Periodic Tasks)

## Periodic Tasks

This feature adds periodic tasks which are defined by a new `tasks` section which is array of objects such as:

```json
"tasks": [
    {
      "name": "task",
      "command": ["/bin/task.sh","arg1"],
      "frequency": "1500ms",
      "timeout": "100ms"
    }
  ]
```

More information on usage is in the README

## Integration Tests - Single Test

I found it useful to run a single integration test at a time. I added that capability to the Makefile.

```
$ make integration TEST=test_tasks
```

## utils.NewLogWriter

One thing of note is that there are now a lot of commands mapping standard out/standard error directly to the main stderr/stdout. This is problematic if I like to keep my shimmed application logging to stdout and ContainerPilot to stderr.

I put in a utility function `utils. NewLogWriter ` which can be used as a pipe to `logrus`. This basically makes these commands log using the preferred logging framework rather than raw standard out/standard err. It might be generally useful for the other ancillary commands like `preStart` `preStop` `postStop` and anything else we run that is not the shimmed application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,15 @@ The format of the JSON file configuration is as follows:
         "check": ["/bin/sensor.sh"]
       }
     ]
-  }
+  },
+  "tasks": [
+    {
+      "name": "task",
+      "command": ["/bin/task.sh","arg1"],
+      "frequency": "1500ms",
+      "timeout": "100ms"
+    }
+  ]
 }
 ```
 
@@ -203,6 +211,16 @@ If a `telemetry` option is provided, ContainerPilot will expose a [Prometheus](h
 
 Details of how to configure the telemetry endpoint and how the telemetry endpoint works can be found in the [telemetry README](https://github.com/joyent/containerpilot/blob/master/telemetry/README.md).
 
+#### Tasks (Optional):
+
+Tasks are commands that are run periodically. They are typically used to perform housekeeping such as incremental back-ups, or pushing metrics to systems that cannot collect metrics through service discovery like Prometheus.
+
+Tasks are defined the following properties:
+
+- `command` is the executable (and its arguments) that will run when the task executes.
+- `frequency` is the time between executions of the task. Supports milliseconds, seconds, minutes. The frequency must be a positive non-zero duration with a time unit suffix. (Example: `60s`) Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`.
+- `timeout` is the amount of time to wait before forcibly killing the task.  Tasks killed in this way are terminated immediately (`SIGKILL`) without an opportunity to clean up their state. This value is optional and defaults to the `frequency`
+- `name` is a friendly name given to the task for logging purposes - this has no effect on the task execution. This value is optional, and defaults to the `command` if not given.
 
 #### Other fields:
 
--- a/README.md
+++ b/README.md
@@ -218,8 +218,8 @@ Tasks are commands that are run periodically. They are typically used to perform
 Tasks are defined the following properties:
 
 - `command` is the executable (and its arguments) that will run when the task executes.
-- `frequency` is the time between executions of the task. Supports milliseconds, seconds, minutes. The frequency must be a positive non-zero duration with a time unit suffix. (Example: `60s`) Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`.
-- `timeout` is the amount of time to wait before forcibly killing the task.  Tasks killed in this way are terminated immediately (`SIGKILL`) without an opportunity to clean up their state. This value is optional and defaults to the `frequency`
+- `frequency` is the time between executions of the task. Supports milliseconds, seconds, minutes. The frequency must be a positive non-zero duration with a time unit suffix. (Example: `60s`) Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`. The minimum frequency is `1ms`
+- `timeout` is the amount of time to wait before forcibly killing the task.  Tasks killed in this way are terminated immediately (`SIGKILL`) without an opportunity to clean up their state. This value is optional and defaults to the `frequency`. The minimum timeout is `1ms`
 - `name` is a friendly name given to the task for logging purposes - this has no effect on the task execution. This value is optional, and defaults to the `command` if not given.
 
 #### Other fields:
--- a/README.md
+++ b/README.md
@@ -222,6 +222,8 @@ Tasks are defined the following properties:
 - `timeout` is the amount of time to wait before forcibly killing the task.  Tasks killed in this way are terminated immediately (`SIGKILL`) without an opportunity to clean up their state. This value is optional and defaults to the `frequency`. The minimum timeout is `1ms`
 - `name` is a friendly name given to the task for logging purposes - this has no effect on the task execution. This value is optional, and defaults to the `command` if not given.
 
+*Note on task frequency:* **pick a frequency greater than 50ms**. Although the task configuration permits frequencies as fast a 1ms, the overhead of spawning a process and its lifecycle is likely to be anywhere from 2ms to 25ms. Your task may not be able to run at all, or it might always be killed before it gets any useful work done.
+
 #### Other fields:
 
 - `onStart` is the executable (and its arguments) that will be called immediately prior to starting the shimmed application. This field is optional. If the `onStart` handler returns a non-zero exit code, ContainerPilot will exit.
--- a/services/services.go
+++ b/services/services.go
@@ -120,6 +120,11 @@ func (s *Service) PollAction() {
 	}
 }
 
+// PollStop does nothing in a Service
+func (s *Service) PollStop() {
+	// Nothing to do
+}
+
 // SendHeartbeat sends a heartbeat for this service
 func (s *Service) SendHeartbeat() {
 	s.discoveryService.SendHeartbeat(s.definition)
--- a/services/services.go
+++ b/services/services.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"time"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/joyent/containerpilot/discovery"
 	"github.com/joyent/containerpilot/utils"
 )
@@ -155,6 +156,6 @@ func (s *Service) CheckHealth() (int, error) {
 	if s.healthCheckCmd == nil {
 		return 0, nil
 	}
-	exitCode, err := utils.Run(s.healthCheckCmd)
+	exitCode, err := utils.RunWithFields(s.healthCheckCmd, log.Fields{"process": "health", "serviceName": s.Name, "serviceID": s.ID})
 	return exitCode, err
 }
--- a/services/services.go
+++ b/services/services.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"time"
 
 	"github.com/joyent/containerpilot/discovery"
 	"github.com/joyent/containerpilot/utils"
@@ -106,8 +107,8 @@ func parseService(s *Service, disc discovery.DiscoveryService) error {
 
 // PollTime implements Pollable for Service
 // It returns the service's poll interval.
-func (s Service) PollTime() int {
-	return s.Poll
+func (s Service) PollTime() time.Duration {
+	return time.Duration(s.Poll) * time.Second
 }
 
 // PollAction implements Pollable for Service.
--- a/tasks/tasks.go
+++ b/tasks/tasks.go
@@ -1,0 +1,177 @@
+package tasks
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os/exec"
+	"strings"
+	"sync"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/joyent/containerpilot/utils"
+)
+
+// TaskConfig configures tasks that run periodically
+type TaskConfig struct {
+	Name      string   `json:"name"`
+	Args      []string `json:"command"`
+	Frequency string   `json:"frequency"`
+	Timeout   string   `json:"timeout,omitempty"`
+
+	freqDuration    time.Duration
+	timeoutDuration time.Duration
+	cmd             *exec.Cmd
+	quitChannel     chan int
+	lock            *sync.Mutex
+	ticker          *time.Ticker
+	logWriters      []io.WriteCloser
+}
+
+// Task a task that runs periodically until killed
+type Task interface {
+	Start() error
+	Stop() error
+}
+
+func (t *TaskConfig) closeLogs() {
+	if t.logWriters == nil {
+		return
+	}
+	for _, w := range t.logWriters {
+		if err := w.Close(); err != nil {
+			log.Errorf("Unable to close log writer : %v", err)
+		}
+	}
+}
+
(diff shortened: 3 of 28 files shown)


justenwalker:
Weird, but it appears to have a problem now with `test_sighup_deadlock` that i didn't encounter locally or when I pushed my branch to TravisCI. Need more logs to determine why this happens.

justenwalker:
Now one of my unit tests are failing - I'll have to play whack-a-mole later when I have more time.  But feel free to review what I have so far @tgross

tgross:
> One thing of note is that there are now a lot of commands mapping standard out/standard error directly to the main stderr/stdout. This is problematic if I like to keep my shimmed application logging to stdout and ContainerPilot to stderr.

I like this idea. But I'm not sure I'm understanding the implementation. Would we update the ExecuteAndWait functions to use this? Can you walk me thru how you'd configure the example where you've got the shimmed app going to stdout and ContainerPilot to stderr?

justenwalker:
> Would we update the ExecuteAndWait functions to use this?

Yes. Instead of

```go
// Executes the given command and blocks until completed
func ExecuteAndWait(cmd *exec.Cmd) (int, error) {
	if cmd == nil {
		return 0, nil
	}
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
	if err := cmd.Run(); err != nil {
		if exiterr, ok := err.(*exec.ExitError); ok {
			if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
				return status.ExitStatus(), err
			}
		}
		log.Errorln(err)
		return 1, err
	}
	return 0, nil
}
```

It would be more like:

```go
func ExecuteAndWait(cmd *exec.Cmd) (int, error) {
	if cmd == nil {
		return 0, nil
	}
	fields := make(map[string]interface{})
	fields["command"] = cmd.Args[0]
	stdout := utils.NewLogWriter(fields, log.DebugLevel)
	stderr := utils.NewLogWriter(fields, log.WarnLevel)
	defer stdout.Close()
	defer stderr.Close()
	cmd.Stdout = stdout
	cmd.Stderr = stderr
	if err := cmd.Run(); err != nil {
		if exiterr, ok := err.(*exec.ExitError); ok {
			if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
				return status.ExitStatus(), err
			}
		}
		log.Errorln(err)
		return 1, err
	}
	return 0, nil
}
```

justenwalker:
@tgross I think all the tests are finally passing now - Plus it appears that the JSON panic fix induced some other bug in TestTerminateSignal which I've attempted to "fix".

tasks/tasks.go
```diff
@@ -0,0 +1,194 @@
...
+type TaskConfig struct {
+	Name      string   `json:"name"`
+	Args      []string `json:"command"`
+	Frequency string   `json:"frequency"`
+	Timeout   string   `json:"timeout,omitempty"`
+
+	freqDuration    time.Duration
+	timeoutDuration time.Duration
+	cmd             *exec.Cmd
+	quitChannel     chan int
```
tgross: `Task`/`TaskConfig` has a very different approach to polling than all our other pollable components. Can you walk me thru why we're making this change? I can see that the package API is a little cleaner but it results in the polling logic being repeated (and if we were to refactor the other polling APIs in the same way that multiplies).

utils/log_writer.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"bufio"
+	"io"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// NewLogWriter pipes stdout/err logs to logrus
+func NewLogWriter(fields log.Fields, level log.Level) io.WriteCloser {
```
tgross: I'm not sure I understand what this looks like for the end-user. Can you walk me thru how you'd configure the example where you've got the shimmed app going to stdout and ContainerPilot to stderr? (That was your original suggested use case).

tasks/tasks.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (t *TaskConfig) execute() {
+	log.Debugf("task[%s].execute init", t.Name)
+	cmd := utils.ArgsToCmd(t.Args)
+	t.cmd = cmd
+	fields := make(map[string]interface{})
+	fields["task"] = t.Name
+	t.logWriters = []io.WriteCloser{
+		utils.NewLogWriter(fields, 5),
+		utils.NewLogWriter(fields, 4),
```
tgross: This looks like we're hard-coding log levels into the logger? (And using magic numbers in the process?)

tasks/tasks.go
```diff
@@ -0,0 +1,194 @@
...
+		case <-t.quitChannel:
+			log.Debugf("task[%s].Start#gofunc <-quitChannel", t.Name)
+			t.ticker.Stop()
+			quit <- 0
+			t.kill(t.cmd)
+			return
+		}
+	}()
+
+	// Run command
```
tgross: I *think* this is in its own goroutine because `t.execute()` should be blocking on wait, but `run` includes a handler for this same channel. Am I missing something here?

tasks/tasks.go
justenwalker: The magic numbers were to get around a problem I had when we used multiple Godeps. I couldn't passe log.DebugLevel since it was the wrong dependency. This can be changed back now that we no longer have multiple godeps.

tasks/tasks.go
justenwalker: `run` creates a new channel and a new ticker for killing the running task after the `timeout` period.

utils/log_writer.go
justenwalker: I didn't actually plan on this being user-configurable. The task always logs to the `logging` config.  Stream maps to a particular LogLevel: `stdout => InfoLevel` and `stderr => DebugLevel`

It can be configured if we want to add that option. I could put it in as part of the `LogConfig` and optionally choose some default values if it is not provided.

tasks/tasks.go
justenwalker: I think some of it can be adapted to use the Pollable interface - but the PollTime is an integer - not a duration - so it assumes Seconds.

Can we alter the API to supply a Duration instead?

utils/log_writer.go
tgross: This would send all `stderr` from our shimmed application as `DebugLevel` and all `stdout` from our shimmed application as `Info`? That means that applications would effectively drop everything coming out of `stderr` in typical production environments, which seems unwise. If we're going to assign log-levels at all, shouldn't `stderr` be at a higher-priority level?

tasks/tasks.go
tgross: Yeah, I like that approach.

And, for a bonus, it lets us start a future bikeshed about whether polling for Service and Backend [should take fractional seconds](http://bikeshed.com/). 😀

utils/log_writer.go
justenwalker: Fair point - I could swap the log levels. I was coming from the point of view that, Despite the name containing `err`, `stderr` is for debug statements (the ones you get when you `set -x`) and `stdout` is for stuff you care about generally.

services/services.go
```diff
@@ -119,6 +120,11 @@ func (s *Service) PollAction() {
 	}
 }
 
+// PollStop does nothing in a Service
+func (s *Service) PollStop() {
```
tgross: I really like that we added this to the interface... I think we can leverage it down the road to maybe simplify the quit logic we have.

utils/log_writer.go
justenwalker: Maybe the default is to put them both at InfoLevel?

tgross: Ok, those levels make sense then... I think what I might be getting distracted by is that whatever's coming out of stdout/stderr is either unstructured or structured by the application. If the application is sending unstructured logs then this feature lets us provide structure to them, which is nice. But if the application is sending structured logs then we have opportunities to drop logs unintentionally.

Suppose our application is doing structured logging at the DEBUG level but ContainerPilot is configured for INFO. When the logs get shipped off to a log collector, all the DEBUG logs are going to be logged at INFO in the collector. Maybe it's just a matter of telling end-users they shouldn't use the `text` option if the application sends structured logs? Kinda sucks that we then drop the structure from ContainerPilot logs in that scenario though.

README.md
```diff
@@ -203,6 +211,16 @@ If a `telemetry` option is provided, ContainerPilot will expose a [Prometheus](h
...
 Details of how to configure the telemetry endpoint and how the telemetry endpoint works can be found in the [telemetry README](https://github.com/joyent/containerpilot/blob/master/telemetry/README.md).
 
+#### Tasks (Optional):
+
+Tasks are commands that are run periodically. They are typically used to perform housekeeping such as incremental back-ups, or pushing metrics to systems that cannot collect metrics through service discovery like Prometheus.
+
+Tasks are defined the following properties:
+
+- `command` is the executable (and its arguments) that will run when the task executes.
+- `frequency` is the time between executions of the task. Supports milliseconds, seconds, minutes. The frequency must be a positive non-zero duration with a time unit suffix. (Example: `60s`) Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`.
```
misterbisson: Should frequency and timeout have minimum values? A `1ns` frequency could DOS the container, for example.

You can imagine me saying "one minute is good enough for cron, so it’s good enough for me, ayup" in a folksy tone, if you want ;-). I'm not saying the minimum has to be one minute, but even once a second seems rather frequent for utility functions.

Maybe I'm wrong, though.

utils/log_writer.go
justenwalker: I'm not sure I fully grasp the issue since the terms we are using are confusing. I'll define what I mean and then try to state what the code is currently doing. If I need to make any changes, let me know using the terms below.

-  **application**: the shimmed application run by ContainerPilot
-  **process**: a process being run by ContainerPilot as part of it's normal operation: health check, task, sensor, etc...
- **log**: refers to the `logrus` logging backend
- **output**: refers to the two native output streams: (`os.Stdout` and `os.Stderr`)

ContainerPilot will always allow the application to stream its logs directly to `stdout` and `stderr` since it attaches the streams directly in. Currently this is done by default in `util.ExecuteAndWait`.

I'm thinking that the behavior for **app** remains the same. In essence, an **app** never writes output to the **log** and thus is unaffected by the log levels in the logging config.

```go
cmd.Stdout = os.Stdout
cmd.Stderr = os.Stderr
``` 

But for a **process** we take a different approach and instead:

```go
stdout := utils.NewLogWriter(fields, log.InfoLevel)
stderr := utils.NewLogWriter(fields, log.DebugLevel)
cmd.Stdout = stdout
cmd.Stderr = stderr
```

This means that all **process** output is shipped to the **log** which can be adjusted with the log config.

I think at this point we just need to agree on which `logrus.Level` **process** output streams should map to by default, and if those streams should be configurable.

README.md
justenwalker: I'd opt for more configurability over less, with some sane default unit. If you put `60` it's implied `60s`. If people want to blow their own foot off - they are free to do so.

The reasons why the frequency should be adjustable is that - I may have a local statsd agent that I'm using these tasks to ship metrics to at some reasonably high frequency - and it does some  aggregations to some longer time-period - I don't want to push that logic into every container.

utils/log_writer.go
tgross: > ContainerPilot will always allow the application to stream its logs directly to `stdout` and `stderr` since it attaches the streams directly in

Ah, I think I thought you were proposing we use the same mechanism for the application! And sure, because *process* output streams are most likely unstructured I think that's a good approach and not likely to be problematic. Given that, I'm good-to-go with the log levels you had originally. Thanks for getting this thru my noggin. 😀

README.md
tgross: I don't think @misterbisson argues that it shouldn't be adjustable, but that if we fire the task at ns or µs resolution we're not going to have good results. If for no other reason than that it's going to take that long to launch the task executable and wire it up.

misterbisson: > some reasonably high frequency

@justenwalker what frequency do you have in mind? I definitely want configurability (@tgross is correct in interpreting my earlier comment), but I'd also be cautious of attempting to support frequencies that are best measured in sub-second values.

justenwalker: For my specific use-case it probably will be on the order of milliseconds, but I didn't want to discount other use-cases which might need finer-grained resolution. 

An argument to your point however; we can't guarantee the overhead of spawning a process and managing its lifecycle will allow for nano-second precision frequency, so perhaps milliseconds are a reasonable floor.

justenwalker:
@misterbisson I added benchmarks on the `Run` function and it seems pretty stable around:

```
BenchmarkRunSuccess-2	    1000	   2038736 ns/op
```

So roughly 2ms/op ... Based on this test I've put a minimum timeout/frequency of 1ms

utils/log_writer.go
justenwalker: @tgross 077b74c2c19416139718ad1c818d7ebe8cdbb04d adds support for this as we've discussed

README.md
tgross: Hm... is golang's runtime really going to be able to give us precision at this level? Between the userland threads and the stop-the-world GC with a worst-case pause time of 10ms (even in the new go1.5+), I worry that we're going to have folks filing bugs on the tasks feature because of this. Usually you'd use counter or timestamps k/v for the sort of things that require higher precision.

Does this deserve a documentation warning maybe?

justenwalker: Should we increase the floor to 10ms? Or just document that any frequency below 50ms (some headroom for us) is not guaranteed to work.

tgross: 10ms seems like an arbitrary so I'd rather break it on a "fault line" of a unit. Let's just document it as you've proposed.

justenwalker: @tgross added note about task frequency to the README

README.md
```diff
@@ -203,6 +211,18 @@ If a `telemetry` option is provided, ContainerPilot will expose a [Prometheus](h
...
+
+- `command` is the executable (and its arguments) that will run when the task executes.
+- `frequency` is the time between executions of the task. Supports milliseconds, seconds, minutes. The frequency must be a positive non-zero duration with a time unit suffix. (Example: `60s`) Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`. The minimum frequency is `1ms`
+- `timeout` is the amount of time to wait before forcibly killing the task.  Tasks killed in this way are terminated immediately (`SIGKILL`) without an opportunity to clean up their state. This value is optional and defaults to the `frequency`. The minimum timeout is `1ms`
+- `name` is a friendly name given to the task for logging purposes - this has no effect on the task execution. This value is optional, and defaults to the `command` if not given.
+
+*Note on task frequency:* **pick a frequency greater than 50ms**. Although the task configuration permits frequencies as fast a 1ms, the overhead of spawning a process and it's lifecycle is likely to be anywhere from 2ms to 25ms. Your task may not be able to run at all, or it might always be killed before it gets any useful work done.
```
tgross: grammatical nit: "its lifecycle"

tgross:
Once last grammar fix for the README and I think this is good-to-go.

justenwalker:
> Once last grammar fix for the README

fixed @tgross

tgross:
Awesome. LGTM. Thanks for all the effort, @justenwalker

Issue #128, opened 2016-04-12
Split application state from config struct


tgross:
@justenwalker proposed the following in #123 and I want to make sure we come back to it as a future refactoring.

---

I can see how this would be painful.  I still think we are complicating the idea of external configuration (ie from a file) with our application state (internal configuration) - I believe the fact that the structures are very similar is incidental.

Perhaps this isn't the time to undertake such a refactor, but does the idea of separating these concepts from each other seem like a good goal? Maybe we can create another issue for that. 

Here are some of my thoughts:

instead of

```go
// Config is the top-level Containerbuddy Configuration
type Config struct {
	Consul       string                    `json:"consul,omitempty"`
	Etcd         json.RawMessage           `json:"etcd,omitempty"`
	LogConfig    *LogConfig                `json:"logging,omitempty"`
	OnStart      json.RawMessage           `json:"onStart,omitempty"`
	PreStart     json.RawMessage           `json:"preStart,omitempty"`
	PreStop      json.RawMessage           `json:"preStop,omitempty"`
	PostStop     json.RawMessage           `json:"postStop,omitempty"`
	StopTimeout  int                       `json:"stopTimeout"`
	Services     []*services.ServiceConfig `json:"services"`
	Backends     []*backends.BackendConfig `json:"backends"`
	PreStartCmd  *exec.Cmd
	PreStopCmd   *exec.Cmd
	PostStopCmd  *exec.Cmd
	Command      *exec.Cmd
	QuitChannels []chan bool
	ConfigFlag   string
}
```

something like

```go
type ApplicationState struct {
	PreStartCmd  *exec.Cmd
	PreStopCmd   *exec.Cmd
	PostStopCmd  *exec.Cmd
	Command      *exec.Cmd
	QuitChannels []chan bool
	Backends []backends.Backend
	Service  []services.Service
	ConfigFlag   string
}
```

Inside `loadConfig` some logic happens to parse the map[string]interface{} into several config structs
which will be passed as arguments to the `NewX` functions in packages which are responsible for converting them to the appropriate interfaces ([github.com/mitchellh/mapstructure](https://github.com/mitchellh/mapstructure) might help). Those config structs live in the package itself.  

For instance, after loading up the `var backendConfig backends.BackendConfig` we get `backend := backends.NewBackend(backendConfig)` which is put into the `ApplicationState` struct

More concrete steps:

- `loadConfig(config map[string]interface{})` will be a pure function which takes a raw config and produces a `ApplicationState` 
[...]

Issue #122, opened 2016-03-30
Allow numeric configuration values to be passed as strings


justenwalker:
Accept a string in place of a number for things like `port` and `ttl` - The change should be backwards compatible - it would accept both numbers and strings which contain a number.

**Use Case**

I have a process that I'm working on that generates a containerbuddy config file. It would be much more useful and maintainable to use the build-in serialization for JSON objects - I can put template definitions into strings, but it's impossible to do the same for numbers. A work around would be to concatenate strings together to form a JSON document instead of using a serializer - which is not ideal.

tgross:
Sounds good to me.

tgross:
> I have a process that I'm working on that generates a containerbuddy config file. It would be much more useful and maintainable to use the build-in serialization for JSON objects

Just out of curiosity, what language/library is this? JSON serialization without support numeric types is surprising (most of the stdlibs I've worked with have it).

justenwalker:
I'm using go.  It has support for numbers - the problem is this is not valid JSON so I can't deserialize it using the normal method:

```json
{
  "key": {{ .ENV_PORT }}
}
```

misterbisson:
@justenwalker I see your use case, and now I'm curious about what's behind it ;-)

As @tgross said, this sounds good.

tgross:
Was covered in https://github.com/joyent/containerpilot/issues/138

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/2.1.1

PR #138, opened 2016-04-23
Refactor: Config to Application State

justenwalker:
For #128

This is an initial stab at factoring out the application state from configuration.  One cool thing about this is that many (all?) of the global variables go away.

I also added support for telemetry reload (hopefully, I have not tested this extensively)

I thought I'd get this out in the open so that everyone can poke some holes in it.

cc: @tgross 

## Work in progress

One thing I'm not too happy about is the proliferation of methods that I implemented on the `Config` struct. I think the next step would probably be to push that logic into the packages themselves.  I've had some success implementing configs as some high-level structure containing mostly maps and using [github.com/mitchellh/mapstructure](https://github.com/mitchellh/mapstructure) - since it supports weak typing which would be good for things like #122

### DIFF
--- a/config/commands.go
+++ b/config/commands.go
@@ -1,0 +1,42 @@
+package config
+
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/joyent/containerpilot/utils"
+	"github.com/prometheus/common/log"
+)
+
+// ParsePreStart ...
+func (cfg *Config) ParsePreStart() (*exec.Cmd, error) {
+	// onStart has been deprecated for preStart. Remove in 2.0
+	if cfg.PreStart != nil && cfg.OnStart != nil {
+		log.Warnf("The onStart option has been deprecated in favor of preStart. ContainerPilot will use only the preStart option provided")
+	}
+
+	// alias the onStart behavior to preStart
+	if cfg.PreStart == nil && cfg.OnStart != nil {
+		log.Warnf("The onStart option has been deprecated in favor of preStart. ContainerPilot will use the onStart option as a preStart")
+		cfg.PreStart = cfg.OnStart
+	}
+	return parseCommand("preStart", cfg.PreStart)
+}
+
+// ParsePreStop ...
+func (cfg *Config) ParsePreStop() (*exec.Cmd, error) {
+	return parseCommand("preStop", cfg.PreStop)
+}
+
+// ParsePostStop ...
+func (cfg *Config) ParsePostStop() (*exec.Cmd, error) {
+	return parseCommand("postStop", cfg.PostStop)
+}
+
+func parseCommand(name string, args interface{}) (*exec.Cmd, error) {
+	cmd, err := utils.ParseCommandArgs(args)
+	if err != nil {
+		return nil, fmt.Errorf("Could not parse `%s`: %s", name, err)
+	}
+	return cmd, nil
+}
--- a/config/commands.go
+++ b/config/commands.go
@@ -4,8 +4,8 @@ import (
 	"fmt"
 	"os/exec"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/joyent/containerpilot/utils"
-	"github.com/prometheus/common/log"
 )
 
 // ParsePreStart ...
--- a/config/config.go
+++ b/config/config.go
@@ -22,7 +22,7 @@ import (
 
 // Config is the top-level ContainerPilot Configuration
 type Config struct {
-	Consul          string          `json:"consul"`
+	Consul          interface{}     `json:"consul"`
 	Etcd            json.RawMessage `json:"etcd"`
 	LogConfig       *LogConfig      `json:"logging"`
 	OnStart         json.RawMessage `json:"onStart"`
@@ -46,12 +46,16 @@ const (
 // ParseDiscoveryService ...
 func (cfg *Config) ParseDiscoveryService() (discovery.DiscoveryService, error) {
 	var discoveryService discovery.DiscoveryService
+	var err error
 	discoveryCount := 0
 	for _, discoveryBackend := range []string{"Consul", "Etcd"} {
 		switch discoveryBackend {
 		case "Consul":
-			if cfg.Consul != "" {
-				discoveryService = discovery.NewConsulConfig(cfg.Consul)
+			if cfg.Consul != nil {
+				discoveryService, err = discovery.NewConsulConfig(cfg.Consul)
+				if err != nil {
+					return nil, err
+				}
 				discoveryCount++
 			}
 		case "Etcd":
--- a/config/config.go
+++ b/config/config.go
@@ -22,18 +22,18 @@ import (
 
 // Config is the top-level ContainerPilot Configuration
 type Config struct {
-	Consul          string          `json:"consul,omitempty"`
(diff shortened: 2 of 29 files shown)


tgross:
I'm still digging into details as there's a lot here, but just want to say this is some damn nice work @justenwalker

tgross:
> One thing I'm not too happy about is the proliferation of methods that I implemented on the Config struct. I think the next step would probably be to push that logic into the packages themselves.

What I've noticed is that there are broadly two categories of methods there. There are some that are thin wrappers on `somepackage.NewSomeStruct` that check for `nil` and that's about it. Those `NewSomeStruct` all take `json.RawMessage`, so maybe there's a nice way to define a function type for those and then push them into the packages as you've suggested.

The second category are those that are building `exec.Cmd` and stuff like that. I suspect we don't want to try to move those out of the `core` package because it's going to cause us heartburn around imports. But moving them into functions that can be tested in isolation (in `core/commands.go` lets say) could be useful.

config/config.go
```diff
@@ -58,90 +40,19 @@ type Config struct {
 	BackendsConfig  json.RawMessage `json:"backends,omitempty"`
 	TasksConfig     json.RawMessage `json:"tasks,omitempty"`
 	TelemetryConfig json.RawMessage `json:"telemetry,omitempty"`
```
tgross: Not necessarily in scope for this PR, but `omitempty` is really only used when we serialize a struct, right? Maybe we don't really need this annotation.

config/config.go
```diff
@@ -5,44 +5,26 @@ import (
...
 
-var (
-	// Version is the version for this build, set at build time via LDFLAGS
-	Version string
-	// GitHash is the short-form commit hash of this build, set at build time
-	GitHash string
+	log "github.com/Sirupsen/logrus"
 )
 
 // Passing around config as a context to functions would be the ideomatic way.
```
tgross: I know you're WIP on this, but this whole comment will need reworking.

core/app.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+// Reload ...
+func (a *App) Reload() error {
+	a.signalLock.Lock()
+	defer a.signalLock.Unlock()
+	log.Infof("Reloading configuration.")
+
+	newCfg, err := config.ParseConfig(a.ConfigFlag)
+	if err != nil {
```
tgross: We never log this error, which means we fail to reload silently. We should consider doing `log.Errorf` here.

core/app_test.go
```diff
@@ -229,16 +232,16 @@ func argTestCleanup(oldArgs []string) {
...
 	os.Args = []string{"this", "-config", testJSON, "/testdata/test.sh", "test", "--debug"}
-	if _, err := LoadConfig(); err != nil && !strings.Contains(err.Error(), expected) {
+	if _, err := LoadApp(); err != nil && !strings.Contains(err.Error(), expected) {
 		t.Errorf("Expected %s but got %s", expected, err)
 	}
 }
 
 func validateParseError(t *testing.T, input []byte, matchStrings []string) {
-	if cfg, err := unmarshalConfig([]byte(input)); err != nil {
+	if cfg, err := config.UnmarshalConfig([]byte(input)); err != nil {
```
tgross: That we had to export this method from the `config` package makes me question whether this test is at the right level of abstraction.

config/config.go
justenwalker: oops!

config/config.go
justenwalker: Yeah, we can remove it - we really don't re-serialize the config anyway, so the extra tag options are mostly useless.

core/app_test.go
justenwalker: You are correct. I did the minimal work in order to make the tests pass (and some import acrobatics were necessary).
Some of this needs to be re-thought as it is getting hard to test some of this in isolation due to transitive dependencies causing cycles.

core/app.go
```diff
@@ -0,0 +1,342 @@
...
+	Command          *exec.Cmd
+	StopTimeout      int
+	QuitChannels     []chan bool
+	maintModeLock    *sync.RWMutex
+	signalLock       *sync.RWMutex
+	paused           bool
+	ConfigFlag       string
+}
+
+// NewApp ...
```
tgross: In `Reload()` it looks like we're copying all the content of the newly initialized app on top of the existing `App` struct so that our locks still work? That LGTM as a process but maybe `InitializeApp` and `LoadApp` (and probably `HandlePolling` and `InMaintenanceMode`?) should be non-exported and so that the only public interfaces are:

- `NewApp` (i.e. create from scratch only)
- `Reload`
- `Run`
- `ToggleMaintenanceMode`
- `Terminate`

core/signals_test.go
```diff
@@ -63,42 +62,49 @@ func TestMaintenanceSignal(t *testing.T) {
...
+	}
+	app.ConfigFlag = `{ "consul": "newconsul:8500" }`
+	err = app.Reload()
+	if err != nil {
+		t.Errorf("Valid configuration returned error: %v", err)
 	}
-	oldConfig.ConfigFlag = `{ "consul": "newconsul:8500" }`
-	if newConfig := reloadConfig(oldConfig); newConfig == nil || newConfig.Consul != "newconsul:8500" {
+	discSvc := app.DiscoveryService
+	if discSvc == nil {
```
tgross: Shouldn't we also check for `NoopDiscoveryService`, rather than just that it was cleared to be `nil`?

telemetry/telemetry.go
```diff
@@ -20,62 +20,77 @@ type Telemetry struct {
...
 	return t, nil
 }
 
-var server *http.Server
-var serverLock = sync.RWMutex{}
-
+// Serve starts serving the telemetry service
 func (t *Telemetry) Serve() {
-	serverLock.Lock()
-	serverLock.Unlock()
```
tgross: oops! That's an embarrassing bug!

core/app.go
justenwalker: That sounds good.  I can lock down those other calls

justenwalker:
@tgross There is a lot here. I broke it out into individual commits that should be hopefully self-explanitory, but I'll summarize:

- I took care of most of the comments that you made in the previous commit
- I did some refactoring of the implementation classes to use `interface{}` instead of `json.RawMessage` for reasons that become a little clearer after you read the source code.
- As a result of using `mapstructure` for decoding configs, I've completed #122 (tested in 9776ebf)
- I put in the ordering fix from #140 (see 2500db4) so that reaping happens first before everything, and signal handling starts before preStart runs.
- Fixed up the `add-dep` target in the makefile, it didn't seem to work when I added `mapstructure`

config/commands.go
```diff
@@ -0,0 +1,42 @@
+package config
+
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/joyent/containerpilot/utils"
+	"github.com/prometheus/common/log"
```
tgross: prometheus/common/log? What happened to logrus here?

tgross:
One last little comment about the log import in `config/commands.go` but otherwise this LGTM.

config/commands.go
justenwalker: Ah, my auto-importer often gets the common names wrong :(

tgross:
LGTM.

PR #141, opened 2016-04-29
changing onStart to preStart in README

dsapala:
onStart was deprecated in favor of preStart but the README still refers to onStart in many places.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ The format of the JSON file configuration is as follows:
 ```json
 {
   "consul": "consul:8500",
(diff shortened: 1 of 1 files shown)


justenwalker:
Nice catch @dsapala! 

In the future, could you elaborate why the change was made in the issue description?

Something like:
> `onStart` was deprecated in favor of `preStart` but the README still refers to `onStart` in many places.

Additionally, maybe we should should mention somewhere that `onStart` is supported, but deprecated somewhere in the README. What do you think @tgross?

dsapala:
Sorry, @justenwalker. Can do.

tgross:
Thanks @dsapala!

Issue #143, opened 2016-05-03
Refactor Discovery Backend Registration


justenwalker:
As we are going to support multiple backends, it would be good if we didn't have to alter global config parser so much.

I propose that we make the config more free-form (`map[string]interface{}`) and support hooking into keys using a Registration function. Discovery backends could call this registration in their `func init() { ... }` to register a handler for their key ("consul", "etcd", "zookeeper", ...)

I'll try an initial PR to get the idea out in a more concrete form. I think this will significantly improve how we parse the configuration file. 

As a part of this enhancement, it would be good to push the discovery backends into their own package.

tgross:
Merged in https://github.com/joyent/containerpilot/issues/144. I'm building a release for this now.

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/2.1.1

PR #144, opened 2016-05-04
Refactor of config and discovery registration

justenwalker:
For #143 

- Decoupled the service discovery backends from `config.go`
- Added README for creating new discovery backends
- Refactored config parser to allow for pluggable config sections (for now just discovery backends)
- Fixed the return code in `lint.sh` script
- Fixed some lint warnings/errors

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -14,6 +14,7 @@ import (
 	"github.com/joyent/containerpilot/services"
 	"github.com/joyent/containerpilot/tasks"
 	"github.com/joyent/containerpilot/telemetry"
+	"github.com/prometheus/common/log"
 )
 
 // Config is the top-level ContainerPilot Configuration
@@ -31,6 +32,7 @@ type Config struct {
 	TasksConfig     []interface{} `json:"tasks"`
 	TelemetryConfig interface{}   `json:"telemetry"`
 
+	RawConfig  map[string]interface{}
 	ConfigFlag string
 }
 
@@ -44,22 +46,15 @@ func (cfg *Config) ParseDiscoveryService() (discovery.DiscoveryService, error) {
 	var discoveryService discovery.DiscoveryService
 	var err error
 	discoveryCount := 0
-	for _, discoveryBackend := range []string{"Consul", "Etcd"} {
-		switch discoveryBackend {
-		case "Consul":
-			if cfg.Consul != nil {
-				discoveryService, err = discovery.NewConsulConfig(cfg.Consul)
+	for _, key := range discovery.GetBackends() {
+		handler := discovery.GetConfigHook(key)
+		if handler != nil {
+			if rawCfg, ok := cfg.RawConfig[key]; ok {
+				discoveryService, err = handler(rawCfg)
 				if err != nil {
 					return nil, err
 				}
-				discoveryCount++
-			}
-		case "Etcd":
-			if cfg.Etcd != nil {
-				discoveryService, err = discovery.NewEtcdConfig(cfg.Etcd)
-				if err != nil {
-					return nil, err
-				}
+				log.Debugf("parsed service discovery backend: %s", key)
 				discoveryCount++
 			}
 		}
@@ -189,6 +184,11 @@ func UnmarshalConfig(data []byte) (*Config, error) {
 		}
 		return nil, newJSONParseError(data, syntax)
 	}
+	var rawCfg map[string]interface{}
+	if err := json.Unmarshal(data, &rawCfg); err != nil {
+		return nil, err
+	}
+	config.RawConfig = rawCfg
 	return config, nil
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -7,33 +7,44 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"os/exec"
+	"reflect"
 	"strings"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/joyent/containerpilot/backends"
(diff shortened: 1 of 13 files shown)


discovery/README.md
```diff
@@ -0,0 +1,71 @@
+# Service Discovery
+
+ContainerPilot can support multiple, "pluggable" backends for service discovery.
```
tgross: Don't hard-wrap text in the READMEs, please.

main.go
```diff
@@ -6,6 +6,10 @@ import (
 	"github.com/joyent/containerpilot/core"
 
 	log "github.com/Sirupsen/logrus"
+
+	// Import backends so that they initialize
```
tgross: There's a bit of a code smell to me with this -- the individual backends are getting imported all the way at the top of our application which is leaking a lot of implementation detail upwards. Is it possible to push this import down into the `discovery` package?

discovery/discovery.go
```diff
@@ -1,19 +1,49 @@
...
+}
+
+// ServiceDiscoveryHandler parses a raw service discovery config
+type ServiceDiscoveryHandler func(interface{}) (Service, error)
+
+var backends = []string{}
+var discoveryHandlers = map[string]ServiceDiscoveryHandler{}
+
+// RegisterHandler registers a service discovery handler for a config key
+func RegisterHandler(name string, handler ServiceDiscoveryHandler) {
```
tgross: This is going to sound nitpicky, but is "handler" really the terminology we want to use here? A handler sounds like it's something we're going to be dealing with on an ongoing basis at runtime (like the poll handlers). I'll admit I'm having trouble coming up with a better name (`Configurator` is a little silly. 😀  )

main.go
justenwalker: I tried this, but it seems like blank imports are suppose to be imported in main package. (The linter complained). Also would cause a dependency loop: `discovery` -> `etcd` -> `discovery`

tgross:
I'm certainly guilty of this myself, but it might be helpful in the future when we do bigger PRs like this if we can avoid doing the lint work at the same time. It makes the "significant" updates a bit harder to follow when we do review.

discovery/discovery.go
justenwalker: Maybe "ServiceDiscoveryConfigurationHook" ? I bit more verbose but perhaps better?

justenwalker:
I can leave the lint work off this PR since it is a separate commit

main.go
tgross: Maybe being Devil's advocate here: why push them into their own package at all?

justenwalker: Seems like they are logically their own package. I feel that `discovery` should be implementation agnostic. Im sort of following the pattern that heka used for plugins.

Also, main.go needs to at least import `discovery` in order for the init chain to work. And then discovery must import `etcd` and `consul` and other backends even though it doesn't care about the implementation classes.

I can see arguments for both, but there doesn't appear to be a clear winner for me, so I'd opt to follow the existing standard.

justenwalker:
@tgross I fixed your comment about the README, the name "Handler", and removed the linting fixes. We can revisit lint errors in a separate PR.

discovery/etcd/etcd.go
```diff
@@ -52,6 +58,11 @@ func parseEndpoints(endpoints interface{}) []string {
 	return nil
 }
 
+// HandleEtcd is the handler function for creating the etcd backend
+func HandleEtcd(raw interface{}) (discovery.DiscoveryService, error) {
```
tgross: Can this be `EtcdConfigHook` too?

discovery/consul/consul.go
```diff
@@ -8,12 +8,22 @@ import (
...
 
+func init() {
+	discovery.RegisterBackend("consul", HandleConsul)
+}
+
 // Consul is a service discovery backend for Hashicorp Consul
 type Consul struct{ consul.Client }
 
+// HandleConsul is the handler function for creating the consul backend
+func HandleConsul(raw interface{}) (discovery.DiscoveryService, error) {
```
tgross: Can this be `ConsulConfigHook` too?

config/config.go
```diff
@@ -227,3 +301,64 @@ func highlightError(data []byte, pos int64) (int, int, string) {
...
+				arr = append(arr, v.Interface())
+			}
+		}
+		return arr
+	}
+	return nil
+}
+
+// We can't use mapstructure to decode our config map since we want the values
+// to also be raw interface{} types. mapstructure can only decode
```
tgross: Isn't that what we do in `utils.DecodeRaw`?

tgross:
Looking pretty good. Just a last couple comments and questions.

config/config.go
justenwalker: `utils.DecodeRaw` is a helper function for using mapstructure to decode a raw `interface{}` value to a target `struct` type. Unfortunately, this doesn't seem to recurse. In my testing, it doesn't work well if the members of the target struct are of []interface{} and interface{} type -- they always come back nil.

tgross: Oh, I see. I was missing where we'd of course be passing in an actual implementation of `interface{}` to `DecodeRaw`.  Makes sense.

justenwalker:
Seems like there is some timing bug in the tasks test. I pushed again to see if it clears up.

tgross:
Ok, this LGTM. We should try to figure out what's up with the tasks test, but it's unrelated so let's do that under a separate issue (I'll open one).

Issue #149, opened 2016-05-11
stopTimeout in config throws "Unknown config keys"


tgross:
@mterron reports:
> Hey, just updated to 2.1.1 and my containers are showing 2016/05/11 15:36:11 Unknown config keys: [stopTimeout]
I checked the code and the setting exists there, as far as I can understand go, at core/app.go and several other places, but sometimes spelled StopTimeout and some other times stopTimeout
I tried changing the config file to both cases but it does the same thing with both
Going back to 2.1.0 solves it

It looks to me like we're missing a line in [`config.decodeConfig`] (https://github.com/joyent/containerpilot/blob/8e8583c0119e7804f871288d0a08ffe7f8cc85cf/config/config.go#L347) that should read `delete(configMap, "stopTimeout")`. I'll work up a PR shortly.

tgross:
@mterron this was released in 2.1.2: https://github.com/joyent/containerpilot/releases/tag/2.1.2
Feel free to reopen if you still run into trouble.

PR #150, opened 2016-05-11
Remove stopTimeout from configMap before checking unknown keys

tgross:
For https://github.com/joyent/containerpilot/issues/149. When we parse the configuration we remove all the expected keys from the `map[string]interface{}` and then verify that there are no remaining keys. Any remaining keys will fire an error `Unknown config keys`. The `stopTimeout` configuration was accidentally left off when we did the refactor in https://github.com/joyent/containerpilot/pull/144, so this fixes that.

I can confirm that if I make only the change to the test config that the test fails as expected:
```
=== RUN   TestValidConfigParse
--- FAIL: TestValidConfigParse (0.00s)
        app_test.go:62: Unexpected error in LoadApp: Unknown config keys: [stopTimeout]
```

cc @mterron @justenwalker @misterbisson

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -349,6 +349,7 @@ func decodeConfig(configMap map[string]interface{}, result *rawConfig) error {
 	delete(configMap, "preStart")
 	delete(configMap, "preStop")
 	delete(configMap, "postStop")
+	delete(configMap, "stopTimeout")
 	delete(configMap, "services")
 	delete(configMap, "backends")
 	delete(configMap, "tasks")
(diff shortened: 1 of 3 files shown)


misterbisson:
Lgtm

misterbisson:
...though it appears the tests failed

tgross:
Hashicorp appears to have removed their binaries from the consul-template repo.

> The command '/bin/sh -c curl -Lo /tmp/consul_template_0.11.0_linux_amd64.zip https://github.com/hashicorp/consul-template/releases/download/v0.11.0/consul_template_0.11.0_linux_amd64.zip &&     unzip /tmp/consul_template_0.11.0_linux_amd64.zip &&     mv consul-template /bin' returned a non-zero code: 9

This will break many our autopilotpattern applications on their next build.  Lemme dig up where they moved them to.

misterbisson:
They're in https://releases.hashicorp.com/consul-template/

justenwalker:
Ah my builds were failing too this morning - good catch (for both)

Issue #146, opened 2016-05-09
Investigate timing issue in tasks test


tgross:
@justenwalker wrote in https://github.com/joyent/containerpilot/pull/144#issuecomment-217302376:

> Seems like there is some timing bug in the tasks test.

The build in question is https://travis-ci.org/joyent/containerpilot/builds/128126162 which passed when we re-pushed it. Here's the failing test:

> 2016/05/05 19:28:30 TEST: test_tasks
Creating testtasks_consul_1
testtasks_consul_1 is up-to-date
Creating testtasks_app_1
Expected task1 to have 7 executions: got 8
Expected task3 to have 2 executions: got 3
No stopped containers
2016/05/05 19:28:39 FAIL: test_tasks

justenwalker:
I have a feeling it is something to do with the timing of the measurement, somehow it got delayed and these tasks were allowed an extra execution.

I think we have to 
- loosen the timings a bit more to account for variance
- make the expected result depend on the time measured
- find another way to test this which doesn't have this variance

tgross:
Closed in https://github.com/joyent/containerpilot/pull/152

PR #152, opened 2016-05-11
integration_tests: loosen requirements for passing test_tasks

justenwalker:
For #146 

It seems that between running the app and detecting the output of the tasks, there could be a sizable delay. Rather than increasing the amount of time the test takes to run, I've just loosened the threshold for passing the test a little bit to account for variance in the time it takes to run and then measure output.


tgross:
This LGTM.

Issue #135, opened 2016-04-15
Telemetry health checks


adamlc:
Its great to see theres now telemetry. I think a couple of small updates need to be made. I see that the telemetry service registers a health check in consul, which is fine but it doesn't seem like the name is configurable and it seems the healthcheck doesn't get deleted when the container stops.

Should hopefully be changed easily enough :)

tgross:
> it seems the healthcheck doesn't get deleted when the container stops

That's definitely a bug... I'll see what I can do about tracking it down.

misterbisson:
> it doesn't seem like the [telemetry service] name is configurable

True. The telemetry service name is not configurable, as it represents ContainerPilot itself. The roadmap for the telemetry service includes [presenting additional metadata about the container](https://github.com/joyent/containerpilot/issues/27#issuecomment-207462572). Predictability in the service name is intended as a way to simplify the development of consumers and new thinking about how this data can be used.

The service ​_does_​ support filtering using tags. You can add tags to the ContainerPilot config and look for them using [the consul_sd_config options](https://prometheus.io/docs/operating/configuration/#<consul_sd_config>), even though that's not demonstrated in https://github.com/autopilotpattern/prometheus.

> the healthcheck doesn't get deleted when the container stops

We agree, that's a bug we need to fix.

adamlc:
> The telemetry service name is not configurable, as it represents ContainerPilot itself. The roadmap for the telemetry service includes presenting additional metadata about the container.

That makes sense, I guess a consumer would just query all of the ContainerPilot services 😸

tgross:
I need to get back to this one so I'm going to take a look at it this morning.

tgross:
I've opened https://github.com/joyent/containerpilot/pull/155 with an integration test exercising the registration and deregistration of the telemetry endpoint in Consul. I can't reproduce this bug with this test or with "by hand" tests, at least with the latest version of ContainerPilot.

Can you confirm this is still a problem @adamlc and maybe provide a Compose file and ContainerPilot config that reproduces the issue if so?

adamlc:
I'll take another look in the next few days @tgross 👍

justenwalker:
[...]

PR #155, opened 2016-05-12
Add integration test to verify telemetry service (de)registration

tgross:
For https://github.com/joyent/containerpilot/issues/135

This integration test verifies that we're registering the telemetry service (`containerpilot`) in the service catalog and that we're deregistering it when we stop the container.

cc @misterbisson @adamlc


misterbisson:
:pickles:

Issue #153, opened 2016-05-11
using consul ACL with containerpilot


ric03uec:
We (at shippable) are experimenting with `containerpilot` as an implementation for `autopilot` pattern. The problem we're running into is when using `consul` ACL with `containerpilot`. The ACL works well with `consul-template` that supports it out of box so all our use cases are supported by default.
- Does `containerpilot` have support for `consul` ACL? I went through the docs but could'nt find any way to do this. 
- Is there any way to make authorization work with `containerpilot`.
We cannot give client `containerpilot` instances the ability to access all of consul data and would ideally want to restrict one instance to a predefined namespace  
Thanks

misterbisson:
@ric03uec ContainerPilot doesn't support [Consul ACLs](https://www.consul.io/docs/internals/acl.html) _today_, but it's a feature I'd love to support. 

Can you start us off by sharing some use-cases and how users might configure them on the ContainerPilot side?

ric03uec:
@misterbisson thanks for the quick response.

We're trying to support the following use-case'

1. Set the `acl_default_policy` of the consul cluster to `deny`
2. Admin navigates to `consul`, ACL page and creates a new ACL(say `serviceACL) with following policy
```
service "foo-" {
  policy: "write"
}
```
This generates a token, say `serviceFooPolicyToken` that can now manipulate any services starting with `foo-` 
3. boot a service `foo-api` using `containerpilot` that uses the token ` serviceFooPolicyToken`. This service should now be registered into `consul` cluster
4. boot a service `foo-ui` using `containerpilot` that uses the token `serviceFooPolicyToken`. This should have `foo-api` as its dependency. `containerpilot` should be able to register this service and get the status of `foo-api` service also because of the correct token
5. boot a service `bar-ui` using `containerpilot` that uses the token `someRandomToken`. `bar-ui` should **not** be able to get the status or even list `foo-api` and `foo-ui`. registration of `bar-ui` should also fail if `someRandomToken` is not a valid token for any ACL

The basic idea to enable namespace based boundaries for resources. enabling ACL will help us get that working. **THE** best solution would be to have ACL and an explicit field called `namespace` in the config file, something like
```
{
  "consul" : "localhost:8500",
  "token" : "sometoken",
  "namespace": "foo",
  ... other configs
}
```
But i think this will require changes in `consul` itself.

tgross:
[...]

PR #156, opened 2016-05-12
Add note to README calling out CONSUL_HTTP_TOKEN

tgross:
For https://github.com/joyent/containerpilot/issues/153. This PR calls out that users can configure the `CONSUL_HTTP_TOKEN` environment variable in order to use Consul's ACL system.

cc @misterbisson @ric03uec


misterbisson:
:pickles:

PR #159, opened 2016-05-18
Not registerd App.ConfigFlag value for reload

fastpopo:
SIGHUP signal doesn't work because the app instance doesn't have ConfigFlag.

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -102,6 +102,7 @@ func NewApp(configFlag string) (*App, error) {
 	a.Backends = cfg.Backends
 	a.Tasks = cfg.Tasks
 	a.Telemetry = cfg.Telemetry
+	a.ConfigFlag = configFlag
 	return a, nil
 }
 


tgross:
@fastpopo this breaks the SIGHUP deadlock integration test: https://travis-ci.org/joyent/containerpilot/builds/131124923. Are you sure this is the source of whatever problem you're seeing? The integration tests for SIGHUP were passing and I can't reproduce this myself. Can you provide a reproduction and/or failing test?

justenwalker:
@tgross I'm actually trying to do this as well to cary over the work @wleese did.

The actual error is

```
level=error msg="Error serving telemetry on 172.17.0.38:9090: listen tcp 172.17.0.38:9090: bind: address already in use" 
```

So it looks like the telemetry listener is having a problem shutting down before it is started again.

fastpopo:
previous, the SIGHUP signal doesn't work correctly. so, deadlock test passed without any reload configurations. 

And "bind: address already in use" is different issue. 
This issue caused by underlying socket. 
the socket didn't closed until rebind the Telemetry Listener.

I added new code for prevent this issue and I faced new issue with this problem. 
```
time="2016-05-19T06:08:43Z" level=fatal msg="accept tcp 172.17.0.3:9090: use of closed network connection"
```

I need more time for deep-dive about this issue.

tgross:
ref to https://github.com/joyent/containerpilot/issues/163 too. Going to dive into this later today.

tgross:
Closing in lieu of https://github.com/joyent/containerpilot/pull/165

PR #165, opened 2016-05-26
Fix SIGHUP

tgross:
Fix the config reload problem described in https://github.com/joyent/containerpilot/pull/140 and https://github.com/joyent/containerpilot/pull/159

Includes:
- I've squashed @wleese's commits into a cherry-picked commit a26fa2dfe94dd32363233968b9c6895d255fbaae because the changeset was getting unworkable (and the commit messages could use some work). This subsumes @fastpopo's PR #159. @wleese's commit 27c4f36ba25c57172cfd8e56af03289d591294d3 was already merged into master in @justenwalker's work in 2500db4ed67d7ffa2590647addd81ae62f1dd323
- I've corrected in that same commit an error in the integration test that caused the test runner to race ahead of Consul being started (this is what we see in https://travis-ci.org/joyent/containerpilot/builds/127509059 and that's obscuring the real problem).
- Removed the `signal.Reset` and `handleSignals` calls from the config reload and we have a passing build (ref https://github.com/joyent/containerpilot/pull/165#issuecomment-222688453)

cc @misterbisson as an FYI

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"os/signal"
 	"sync"
 	"syscall"
 	"time"
@@ -229,7 +228,6 @@ func (a *App) Reload() error {
 
 	a.stopPolling()
 	a.forAllServices(deregisterService)
-	signal.Reset()
 
 	a.load(newApp)
 	return nil
@@ -247,7 +245,6 @@ func (a *App) load(newApp *App) {
 	}
 	a.Telemetry = newApp.Telemetry
 	a.Tasks = newApp.Tasks
-	a.handleSignals()
 	a.handlePolling()
 }
 
--- a/core/app.go
+++ b/core/app.go
@@ -32,20 +32,20 @@ var (
 // after it is run, it can be reloaded and paused with signals.
 type App struct {
 	ServiceBackend discovery.ServiceBackend
-	Services         []*services.Service
-	Backends         []*backends.Backend
-	Tasks            []*tasks.Task
-	Telemetry        *telemetry.Telemetry
-	PreStartCmd      *exec.Cmd
-	PreStopCmd       *exec.Cmd
-	PostStopCmd      *exec.Cmd
-	Command          *exec.Cmd
-	StopTimeout      int
-	QuitChannels     []chan bool
-	maintModeLock    *sync.RWMutex
-	signalLock       *sync.RWMutex
-	paused           bool
-	ConfigFlag       string
+	Services       []*services.Service
+	Backends       []*backends.Backend
+	Tasks          []*tasks.Task
+	Telemetry      *telemetry.Telemetry
+	PreStartCmd    *exec.Cmd
+	PreStopCmd     *exec.Cmd
+	PostStopCmd    *exec.Cmd
+	Command        *exec.Cmd
+	StopTimeout    int
+	QuitChannels   []chan bool
+	maintModeLock  *sync.RWMutex
+	signalLock     *sync.RWMutex
+	paused         bool
+	ConfigFlag     string
 }
 
 // EmptyApp creates an empty application
@@ -102,6 +102,7 @@ func NewApp(configFlag string) (*App, error) {
 	a.Backends = cfg.Backends
 	a.Tasks = cfg.Tasks
 	a.Telemetry = cfg.Telemetry
+	a.ConfigFlag = configFlag
 	return a, nil
 }
 
--- a/integration_tests/fixtures/app/Dockerfile
+++ b/integration_tests/fixtures/app/Dockerfile
@@ -9,9 +9,12 @@ RUN npm install -g json http-server
 
 COPY build/containerpilot /bin/containerpilot
 COPY app-with-consul.json /app-with-consul.json
+COPY app-with-consul-prestart-sigusr1.json /app-with-consul-prestart-sigusr1.json
+COPY app-with-consul-prestart-sighup.json /app-with-consul-prestart-sighup.json
 COPY app-with-etcd.json /app-with-etcd.json
 COPY reload-app.sh /reload-app.sh
 COPY reload-app-etcd.sh /reload-app-etcd.sh
+COPY reload-app-prestart.sh /reload-app-prestart.sh
 COPY sensor.sh /sensor.sh
 COPY task.sh /task.sh
 
--- a/integration_tests/fixtures/app/app-with-consul-prestart-sighup.json
+++ b/integration_tests/fixtures/app/app-with-consul-prestart-sighup.json
@@ -1,0 +1,45 @@
+{
+  "consul": "consul:8500",
+  "preStart": ["/reload-app-prestart.sh", "HUP"],
+  "logging": {
+    "level": "DEBUG",
+    "format": "text"
+  },
+  "services": [
+    {
+      "name": "app",
+      "port": 8000,
+      "health": "/usr/bin/curl --fail -s -o /dev/null http://localhost:8888",
+      "poll": 1,
+      "ttl": 5,
+      "tags": ["application"]
+    }
+  ],
+  "backends": [
+    {
+      "name": "nginx",
+      "poll": 7,
+      "onChange": "/reload-app.sh"
+    },
+    {
+      "name": "app",
+      "poll": 5,
+      "onChange": "/reload-app.sh",
+      "tag": "application"
+    }
+  ],
+  "telemetry": {
+    "port": 9090,
+    "sensors": [
+       {
+        "namespace": "containerpilot",
+        "subsystem": "app",
+        "name": "some_counter",
+        "help": "help text",
+        "type": "counter",
+        "poll": 1,
+        "check": ["/sensor.sh", "count"]
+       }
+    ]
+  }
+}
--- a/integration_tests/fixtures/app/app-with-consul-prestart-sigusr1.json
+++ b/integration_tests/fixtures/app/app-with-consul-prestart-sigusr1.json
@@ -1,0 +1,45 @@
+{
+  "consul": "consul:8500",
+  "preStart": ["/reload-app-prestart.sh", "USR1"],
+  "logging": {
+    "level": "DEBUG",
+    "format": "text"
+  },
+  "services": [
+    {
+      "name": "app",
+      "port": 8000,
+      "health": "/usr/bin/curl --fail -s -o /dev/null http://localhost:8000",
+      "poll": 1,
+      "ttl": 5,
+      "tags": ["application"]
+    }
+  ],
+  "backends": [
+    {
+      "name": "nginx",
+      "poll": 7,
+      "onChange": "/reload-app.sh"
+    },
+    {
+      "name": "app",
+      "poll": 5,
+      "onChange": "/reload-app.sh",
+      "tag": "application"
+    }
+  ],
+  "telemetry": {
+    "port": 9090,
+    "sensors": [
+       {
+        "namespace": "containerpilot",
+        "subsystem": "app",
+        "name": "some_counter",
+        "help": "help text",
+        "type": "counter",
+        "poll": 1,
+        "check": ["/sensor.sh", "count"]
+       }
+    ]
+  }
+}
--- a/integration_tests/fixtures/app/reload-app-prestart.sh
+++ b/integration_tests/fixtures/app/reload-app-prestart.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# Sends a signal to ContainerPilot during the preStart
+
+# wait a few seconds for the Consul container to become available
+n=0
+while true
+do
+    if [ n == 10 ]; then
+        echo "Timed out waiting for Consul"
+        exit 1;
+    fi
+    curl -Ls --fail http://consul:8500/v1/status/leader | grep 8300 && break
+    n=$((n+1))
+    sleep 1
+done
+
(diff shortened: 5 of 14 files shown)


fastpopo:
I check this issue, I don't know why but closing HTTP Handler may need more time.

in Telemetry.go: 

type Telemetry struct {
	Port          int           `mapstructure:"port"`
	Interfaces    []interface{} `mapstructure:"interfaces"` // optional override
	Tags          []string      `mapstructure:"tags"`
	SensorConfigs []interface{} `mapstructure:"sensors"`
	Sensors       []*Sensor
	ServiceName   string
	URL           string
	TTL           int
	Poll          int
	mux           *http.ServeMux
	lock          sync.RWMutex
	listen        net.Listener
	addr          net.TCPAddr
	listening     bool
+      waitForServerClose sync.WaitGroup
}

func (t *Telemetry) Shutdown() {
	t.lock.Lock()
	defer t.lock.Unlock()
	if t.listening {
+              t.waitForServerClose.Add()
		t.listen.Close()
		t.listening = false
	}
}

func (t *Telemetry) Serve() {
	t.lock.Lock()
	defer t.lock.Unlock()
	if t.listening {
		return
	}
	ln, err := net.Listen(t.addr.Network(), t.addr.String())
	if err != nil {
		log.Errorf("Error serving telemetry on %s: %v", t.addr.String(), err)
	}
	t.listen = ln
	t.listening = true
	go func() {
		log.Debugf("telemetry: Listening on %s\n", t.addr.String())
+		log.Error(http.Serve(t.listen, t.mux))
+              t.waitForServerClose.Done()
		log.Debugf("telemetry: Stopped listening on %s\n", t.addr.String())
	}()
}

"log.Fatal(http.Serve(t.listen, t.mux))" will terminate containerpilot when I close the underlying socket. So, please use "log.Error()" instead of "log.Fatal()"

Please check this code. it will fix the "already bind address" issue. 

And I faced new issue with this test sighup_deadlock  test that have some issue with signal.
you can confirm that if you apply the above codes. 

+
app.go - Reload() :
signal.Reset(), it will raise hang with many singals. 
So, please use signal.Ignore() instead of signal.Reset(), and then it works fine.

after that, I faced new issue.
the task and register service is not work correctly.

now days, i check this issue for fix SIGHUP signal scenario.

tgross:
> So I check this issue, I don't know why but closing HTTP Handler may need more time.

Yeah, this does seem to be a race -- I have a unit test built on this branch that hits `.Serve()` then `.Shutdown()` and then `.Serve()` again and there's no panic. Which suggests there's something we're missing.

> this code will terminate containerpilot when I close the underlying socket. 
So, please use "log.Error()" instead of "log.Fatal()"

I doubt that we want to just ignore the error and continue. Would the server be working correctly at that point?

tgross:
I see what my unit test was getting wrong there -- shutting down and restarting the _same_ server works fine. We want to create a new server:

```go
func TestTelemetryServerRestart(t *testing.T) {
	if telem, err := NewTelemetry(decodeJSONRawTelemetry(t, jsonFragment)); err != nil {
		t.Fatalf("Could not parse telemetry JSON: %s", err)
	} else {
		telem.Serve()
		telem.Shutdown()
		if telem, err := NewTelemetry(decodeJSONRawTelemetry(t, jsonFragment)); err != nil {
			t.Fatalf("Could not parse telemetry JSON: %s", err)
		} else {
			telem.Serve()
		}
	}
}
```

This will currently fail:
```
time="2016-05-27T13:20:33Z" level=fatal msg="accept tcp 172.17.0.4:8000: use of closed network connection"
exit status 1
```

which reproduces our `test_sighup_deadlock` failure as a unit test. I want to create a new unit test that replicates the `test_sighup_prestart` panic next.

tgross:
> We want to create a new server

To clarify; we "want" to do so in the test because that's what we're currently doing in the code. But the original implementation of telemetry _didn't_ recreate the server. We changed that in https://github.com/joyent/containerpilot/pull/138/commits/680a2266b4dd9c9a59547e1e97c79947e165a1f7 and I think may be where we broke things.

fastpopo:
I checked the issue with my custom test code, the error raised by http.Serve.
When I close the underlying socket by listener.close, the promethus handler will access closed socket.

So, I suggest to you please use the log.error not log.fatal

I checked with golang community and some other communities, I realize that golang not support any graceful shutdown mechanism for closing http.serve

tgross:
> I checked the issue with my custom test code, the error raised by http.Serve.
> When I close the underlying socket by listener.close, the promethus handler will access closed socket.
> So, I suggest to you please use the log.error not log.fatal

If you want me to consider this idea then you need to show me the test code that proves that we have a working server when we do so.

fastpopo:
please check this code.

```
package main

import (
  "net"
	"net/http"
  "fmt"
  "time"
  "log"
	"github.com/prometheus/client_golang/prometheus"
)

func main() {
  ip := net.ParseIP("192.168.0.76")
  addr := net.TCPAddr{IP: ip, Port: 8080}
  
  mux := http.NewServeMux()
  mux.Handle("/metrics", prometheus.Handler())
  
  listener, err := net.Listen(addr.Network(), addr.String())

  go func() {
    fmt.Printf("Listening\n")
    log.Fatal(http.Serve(listener, mux))
    fmt.Printf("Terminating\n")
  }()

  time.Sleep(time.Second * 3)
	collector := prometheus.NewCounter(prometheus.CounterOpts{
				Name:      "test_metric",
				Help:      "Test Metric",
	})

	prometheus.Register(collector)
	collector.Add(100)
	collector.Add(10)

  listener.Close()
  time.Sleep(time.Second * 3)

  mux = http.NewServeMux()
  mux.Handle("/metrics", prometheus.Handler())
  
  listener, err = net.Listen(addr.Network(), addr.String())
  if err != nil {
    log.Fatal(err)
  }

  go func() {
    fmt.Printf("Listening\n")
    http.Serve(listener, mux)
    fmt.Printf("Terminating\n")
  }()

  time.Sleep(time.Second * 5)
  fmt.Printf("Terminated!!\n")
}
```

it will produce "use of closed network connection" issue too.

fastpopo:
@tgross please check this issue.
https://github.com/golang/go/issues/11219

tgross:
@fastpopo you're not helping here. I'm already aware of that and I _pointed out that problem in the original implementation_.

tgross:
https://github.com/joyent/containerpilot/pull/165/commits/e340da80093941763dc3a8c23f96efa4af7442f1 is a passing unit test w/ the fix suggested, but the integration tests still fail with a panic:

```
2016/05/27 10:27:05 TEST: test_sighup_prestart
Creating testsighupprestart_consul_1
Creating testsighupprestart_app_1
==== TEST LOGS ====
2016/05/27 14:27:37 Expected app to be healthy after SIGHUP: Service app (tag:) count != 1
==== APP LOGS ====
time="2016-05-27T14:27:08Z" level=info msg="\"172.17.0.2:8300\"" process=PreStart
time="2016-05-27T14:27:08Z" level=info msg="Reloading configuration."
time="2016-05-27T14:27:08Z" level=debug msg="parsed service discovery backend: consul"
time="2016-05-27T14:27:08Z" level=info msg="Deregistering service: app"
time="2016-05-27T14:27:08Z" level=info msg="telemetry: Listening on 172.17.0.3:9090"
time="2016-05-27T14:27:08Z" level=info msg="Deregistering service: containerpilot"
time="2016-05-27T14:27:08Z" level=info msg="telemetry: Closing listner"
time="2016-05-27T14:27:08Z" level=error msg="Error serving telemetry on 172.17.0.3:9090: listen tcp 172.17.0.3:9090: bind: address already in use"
time="2016-05-27T14:27:08Z" level=info msg="telemetry: Listening on 172.17.0.3:9090"
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x6077cd]

goroutine 47 [running]:
net/http.(*Server).Serve(0xc82013af00, 0x0, 0x0, 0x0, 0x0)
        /usr/local/go/src/net/http/server.go:1884 +0x40d
net/http.Serve(0x0, 0x0, 0x7fefadf5c7f0, 0xc8201f60c0, 0x0, 0x0)
        /usr/local/go/src/net/http/server.go:1765 +0x85
github.com/joyent/containerpilot/telemetry.(*Telemetry).Serve.func1(0xc820134d00)
        /cp/src/github.com/joyent/containerpilot/telemetry/telemetry.go:83 +0x191
created by github.com/joyent/containerpilot/telemetry.(*Telemetry).Serve
        /cp/src/github.com/joyent/containerpilot/telemetry/telemetry.go:85 +0x333

```

tgross:
After a bit more debugging into `http.Serve` the panic comes down to `net.Listen` calling `ListenTCP` and then giving us a `nil` listener. Because we only log the expected [`OpError`](https://golang.org/src/net/tcpsock_posix.go?s=10482:10546#L310) we end up panicking. If I change this to `log.Fatalf` then we exit in the expected spot instead of panicking. This doesn't solve the problem of course but at least makes it looks like we knew what we were doing. 😀 

Bottom line here comes back to the original design decision which was that we can't update the telemetry port (ref https://github.com/joyent/containerpilot/pull/123#issuecomment-208455770). I'd like to examine adding graceful reload but this should have been done as a separate feature and tested independently rather than having been folded into the big config refactor.

I'm going to back out the changes made in 680a2266b4dd9c9a59547e1e97c79947e165a1f7 until we can invest the time into finding a correct solution for it. The libraries I've seen all are problematic from a packaging standpoint (ex. using gopkg.in).

tgross:
Current status is that the deadlock integration test is failing. I see deregistration happening and us hitting the no-op for server shutdown that I introduced in 8863602b51469643b708c6f178d6d3a783325eb9. Might just be that we're not giving the app enough time to reregister but looking into it next.

justenwalker:
> After a bit more debugging into http.Serve the panic comes down to net.Listen calling ListenTCP and then giving us a nil listener. Because we only log the expected OpError we end up panicking. If I change this to log.Fatalf then we exit in the expected spot instead of panicking. This doesn't solve the problem of course but at least makes it looks like we knew what we were doing.

One observation is that if the underlying listener is closed, then the HTTP server does through an error with a specific message.  We can ignore that error. (See https://github.com/justenwalker/containerpilot/commit/e85e61ebc0c80ccf7df86129c9cf4d84f9838d58)

justenwalker:
As a point of comparison. Consul doesn't care if the http.Serve throws an error (ttps://github.com/hashicorp/consul/blob/master/command/agent/http.go#L165)

tgross:
> One observation is that if the underlying listener is closed, then the HTTP server does through an error with a specific message. We can ignore that error.

My tests showed we can't, or at least we can't if we want to reload the config later. Let's take the discussion of trying this over to https://github.com/joyent/containerpilot/issues/166

tgross:
As far as the deadlock test situation goes, it's passing intermittently for me.

We're definitely sending more signals than can be handled by the signal handler in the time that occurs. Typically I'll see anywhere between 50-100 handler events instead of the 300 we send. This is the expected and correct behavior, because we lock the signal handler during handling.

The expected order of operations here is:
- lock signalLock
- reset signals
- start telemetry server (no-op)
- handle signals
- handle polling
  - start backend polling
  - start service polling
  - start telemetry
  - start task polling
- unlock signalLock

If I add logging so that every loop thru the SIGHUP sender is timestamped, I find that every time the test fails the handler is locking at the same moment that we *stop* sending signals. Then our 6 second wait for the TTL to expires passes and the test then fails because we haven't restarted the pollables. We are reaching the telemetry server (which is a no-op) but then apparently stopping there. I'm trying to instrument the code non-destructively to see if I can suss out the exact spot we're hanging.

tgross:
>  I find that every time the test fails the handler is locking at the same moment that we stop sending signals.

This turns out just be coincidence. We're fully passing thru the `handlePolling` and `handleSignals` code and getting all the quit channels we expect, but we just stop doing any work on the pollables!

tgross:
Got it! I'd instrumented the code path thru reload and discovered that we're hanging during `signal.Reset`. This calls [this bit](https://golang.org/src/os/signal/signal.go?s=3176:3204#L126) in the signal package. This is probably what @fastpopo ran into earlier but swapping it out for `signal.Ignore` does nothing to improve the situation. This is going to be very hairy to debug.

tgross:
Manage to get this stack trace out via SIGABRT but it's not clear to me where the problem is or whether this SIGABRT is landing during the lock https://gist.github.com/tgross/e9a9ea62e33b033c2115e318a767ca62

tgross:
Ok, so sometimes the way to work thru a solution is to realize you're solving the wrong problem! I've removed the `signal.Reset` and `handleSignals` calls from the config reload and we have a passing build.

Resetting the signal handlers creates a race that we are having trouble resolving with creating deadlocks. Fortunately it turns out that it's unnecessary. When the config is reloaded, we keep the existing App instance and replace its values. This means we can safely keep the signal-handling channel around and just reuse it!

justenwalker:
> This means we can safely keep the signal-handling channel around and just reuse it!

Nice find!

tgross:
Any objections to merging this then, @justenwalker?

justenwalker:
LGTM @tgross 

As far as the telemetry reload stuff. I'll dig a bit deeper into it when I get time.  

As I said in my initial PR i didn't have a chance to test it exhaustively, so no surprise there are problems - my bad.  There are wrappers around the http package that support reload though we might consider using: (See [graceful.go](https://github.com/tylerb/graceful/blob/master/graceful.go))

Issue #145, opened 2016-05-08
Expose ContainerPilot details as env vars for other apps to consume


misterbisson:
There are times when it would be very useful to share ContainerPilot configuration details (and the details it resolves) with the "main" application. The ContainerPiliot-advertised IP addresses for the registered services are one example that would be useful for some apps, so we can make sure they listen on the same interface that ContainerPilot thinks they are.

This usage example addresses Consul startup when there are multiple interfaces with RFC1918 IP addresses:

```yaml
  command: >
    /opt/containerpilot/containerpilot
      /bin/consul agent -server
        -config-dir=/config
        -ui-dir /ui
        -advertise ${CONTAINERPILOT_SERVICE_CONSUL_IP} \
        -bootstrap-expect 3
```

The Consul issue is further described in https://github.com/autopilotpattern/consul/issues/17

tgross:
The particular syntax we have here won't work, because Compose would interpolate that variable before it ever makes it into the container. But I suspect we can get away with allowing golang template syntax (so long as Docker doesn't decide to use that in the future in lieu of shell env var style). We'd then want to create those environment variables by overriding [`os/exec.Cmd.Env`](https://golang.org/pkg/os/exec/#Cmd) when we fork the process.

Design questions/issues:
- What variables do we want to expose?
- If the variables are named dynamically, which this proposal seems to suggest, what naming convention should those variables use?
- I don't know of any mechanism to update the environment variables of the child process after we fork (outside of non-portable `mdb` wizardry), which means if we update the ContainerPilot config then the environment variables in the child process won't be updated.
- We should not replace the `Env` slice entirely, but instead append the new environment variables to the parent's (i.e. ContainerPilot's) environment variables. Otherwise env vars set via `-e` flags on `docker run` and the Compose equivalent will be erased, which would be unexpected behavior.

misterbisson:
Good questions @tgross 

> We should not replace the Env slice entirely, but instead append the new environment variables

Agreed.

> I don't know of any mechanism to update the environment variables of the child process after we fork 

Very fair point, and an acceptable limitation, I think.

> The particular syntax we have here won't work... But I suspect we can get away with allowing golang template syntax

Very good point, and one of the frustrations at the heart of the underlying issue. Golang template syntax should be a fair and workable alternative, however.

> What variables do we want to expose... what naming convention?

This needs more thought.

I need to cast about for more use cases so we can better understand the problem before diving in.

justenwalker:
[...]

PR #170, opened 2016-06-01
Interpolate variables in command args

tgross:
For https://github.com/joyent/containerpilot/issues/145

Use the same template rendering process for the command line as we do in the config file. This allows us to pass environment variables at runtime into the shimmed application's own flags.

I've hopped on this sooner rather than later because I'm working on https://github.com/autopilotpattern/etcd and finding that the etcd configuration simply can't be updated in the `preStart` hooks because they support *only* config flags or env vars, which means we can't do things like set the etcd node name to the hostname otherwise.

cc @misterbisson @justenwalker @dekobon

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -111,7 +111,8 @@ func (a *App) Run() {
 	if 1 == os.Getpid() {
 		reapChildren()
 	}
-	command, err := utils.ParseCommandArgs(flag.Args())
+	args := getArgs(flag.Args())
+	command, err := utils.ParseCommandArgs(args)
 	if err != nil {
 		log.Errorf("Unable to parse command arguments: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


misterbisson:
This is a smaller change than I expected. It looks pretty straightforward.

tgross:
> This is a smaller change than I expected. 

@justenwalker did all the heavy lifting back when we added environment variable interpolation to the config file. 😀 

This just borrows the existing mechanism.

tgross:
Added documentation and rebased.

misterbisson:
This looks clean to me.

justenwalker:
Looks solid - good work 👍

misterbisson:
🏡 🚶

Issue #171, opened 2016-06-01
Extend template rendering for configuration


tgross:
In https://github.com/joyent/containerpilot/issues/145#issuecomment-218928756 @justenwalker proposed a feature like `consul-template` or `confd` where the environment interpolation we do could have extended functionality.

> With golang templates, I think we can provide a mechanism for looking up app-state information. With a bit of tweaking we could do something like [consul-template](https://github.com/hashicorp/consul-template)

> Possibly we could expose an "env" section in the containerpilot config which is loaded just before running any commands

> Limitation is still that reload will not change the environment or the command.

> I'm unsure if we want to pre-process these template env variables along with the rest of the golang templates we currently support, or after the config is loaded (multi-pass). The latter means they'd have to be escaped or otherwise not the same delimiter that golang uses by default.

tgross:
One of the examples we want here is https://github.com/autopilotpattern/consul/issues/17, which I've run into with etcd as well. I'm wondering whether we could accomplish all the known use cases by exposing some new environment variables based on the container configuration automatically. Similar to how Docker exposes information about `--link` as env vars, we could expose something like `CONTAINERPILOT_<servicename>_IP`

tgross:
An implementation path that might work here is to call [os.Setenv](https://golang.org/pkg/os/#Setenv) right [here where we get the service IP](https://github.com/joyent/containerpilot/blob/85596c4fe9b51bc37c778db6e3a7717282881a7e/services/services.go#L98). I _think_ that'd get the variables exposed everywhere we need them.

We'd just need to agree on a naming convention for those env vars.

tgross:
This specific fix would also eliminate some brittle code in user-defined hooks in [autopilotpattern/couchbase](https://github.com/autopilotpattern/couchbase/blob/f428ed912d9d39142f905b306020fb9a5de3c3f4/bin/manage.sh#L15) and [autopilotpattern/elasticsearch](https://github.com/autopilotpattern/elasticsearch/blob/00abe9e52b3602dc9b7c27cf24f92a412665181c/bin/manage.sh#L62)

justenwalker:
> An implementation path that might work here is to call os.Setenv right [here where we get the service IP](https://github.com/joyent/containerpilot/blob/85596c4fe9b51bc37c778db6e3a7717282881a7e/services/services.go#L98). I think that'd get the variables exposed everywhere we need them.

[...]

PR #172, opened 2016-06-03
Expose service IP address as environment variable

tgross:
For https://github.com/joyent/containerpilot/issues/171

For each service, create an environment variable that has the name `CONTAINERPILOT_<SERVICE>_IP` and the value of the IP address where we're advertising that service.

@justenwalker @dekobon I'd like to do an integration test before this gets merged just to make sure it works out ok in practice, but this is good for review of the approach I think.

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -102,7 +103,23 @@ func NewApp(configFlag string) (*App, error) {
 	a.Tasks = cfg.Tasks
 	a.Telemetry = cfg.Telemetry
 	a.ConfigFlag = configFlag
+
+	// set an environment variable for each service IP address so that
+	// forked processes have access to this information
+	for _, service := range a.Services {
+		envKey := getEnvVarNameFromService(service.Name)
+		os.Setenv(envKey, service.IPAddress)
+	}
+
 	return a, nil
+}
+
+// Normalize the validated service name as an environment variable
+func getEnvVarNameFromService(service string) string {
+	envKey := strings.ToUpper(service)
+	envKey = strings.Replace(envKey, "-", "_", -1)
+	envKey = fmt.Sprintf("CONTAINERPILOT_%v_IP", envKey)
+	return envKey
 }
 
 // Run starts the application and blocks until finished
--- a/backends/backends.go
+++ b/backends/backends.go
@@ -31,8 +31,8 @@ func NewBackends(raw []interface{}, disc discovery.ServiceBackend) ([]*Backend, 
 		return nil, fmt.Errorf("Backend configuration error: %v", err)
 	}
 	for _, b := range backends {
-		if b.Name == "" {
-			return nil, fmt.Errorf("backend must have a `name`")
+		if err := utils.ValidateServiceName(b.Name); err != nil {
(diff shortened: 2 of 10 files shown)


tgross:
This looks good at first pass but with an important limitation that we should note: the environment variables for service IPs can't be used in the ContainerPilot config file, because we need to render the config file template beforehand to create the env vars in the first place. I got caught by this while making an integration test.

core/app.go
```diff
@@ -102,6 +103,15 @@ func NewApp(configFlag string) (*App, error) {
 	a.Tasks = cfg.Tasks
 	a.Telemetry = cfg.Telemetry
 	a.ConfigFlag = configFlag
+
+	// set an environment variable for each service IP address so that
+	// forked processes have access to this information
+	for _, service := range a.Services {
+		envKey := fmt.Sprintf("CONTAINERPILOT_%v_IP",
+			strings.ToUpper(service.Name))
```
justenwalker: Unsure, but do we have to sanitize this? ie: converting dashes to underscores, removing characters that shouldn't be part of an environment variable.

dekobon: What is a valid service name? Any UTF-8 string?

justenwalker: According to the [documentation](https://www.consul.io/docs/agent/services.html) 

> DNS-compliant service and tag names may contain any alpha-numeric characters, as well as dashes. Dots are not supported because Consul internally uses them to delimit service tags.

I think we should adopt the same convention. service names should be alpha-numeric with dashes.  We'll normalize to uppercase and convert dashes to underscores to comply with environment variables.

If the service name does not conform, then it doesn't get environment variables.

tgross: Aye, sounds like a sane approach. Will do.

tgross: > I think we should adopt the same convention. service names should be alpha-numeric with dashes. 

Looking at it again, this suggests to me that we should do this in the `config` package and validate the service names there (I'd say do it in the `services` package but we need to do it for `backends` as well). I'm against silently converting/canonicalizing and think we should return an error if the name is bad. This will cause ContainerPilot to halt on initial start, but I'd be willing to discuss it if people have objections.

i.e. the behavior I'm proposing is:
- `my-service` => `my-service` service name and `CONTAINERPILOT_MY_SERVICE_IP` env var
- `my.service` => return error
- `my%service` => return error

tgross:
I've added an integration test and validation to the service names as part of our configuration loading. Next bit is to update the documentation.

tgross:
Gah, curse the linter! Should be fixed now and will rebase so we have a nice clean history when we're done here.

justenwalker:
> i.e. the behavior I'm proposing is:
> 
> my-service => my-service service name and CONTAINERPILOT_MY_SERVICE_IP env var
> my.service => return error
> my%service => return error

Possibly backwards-incompatible since we never validated service names before.  all of our examples have been compliant though.  

If we're strictly SemVer, this suggests we update the major version.  If we make it return a warning instead then we at least allow non-compliant service/backend names while still informing them that, because they are non-compliant, thins won't work as the expect for this new feature. Then in 3.x it becomes an error.

tgross:
> If we make it return a warning instead then we at least allow non-compliant service/backend names while still informing them that, because they are non-compliant, things won't work as expect for this new feature.

Sounds like a good idea.

tgross:
How's that?

justenwalker:
LGTM 👍

Issue #175, opened 2016-06-08
Co-process hooks


tgross:
A frequent request we've received is "how do I run multiple processes with ContainerPilot?" (ex https://github.com/joyent/containerpilot/issues/157 but in other places too). The primary reason we've balked at having ContainerPilot act as a supervisor is because it makes the semantics of restarts and signal handling difficult with relationship to some higher-level scheduler. In other words, if one of the applications stops do we stop the whole container or restart the application?

The topology problems with Consul deployments on Triton and PaaS-like environment described in https://github.com/joyent/containerpilot/issues/162 demonstrates a use case for co-processes where the main application is monitored by ContainerPilot but we want a persistent (rather than periodic) process to be run alongside it. This would also match nicely with running `consul-template` persistently as in #157. Additionally, in these scenarios it's clear that the extra process is intended to be a co-process -- it's not the main application. This means we don't have to worry about passing along signals, etc. other than reaping children.

The proposal here is this:
- allow for a `coprocess` field in the configuration which takes an array of commands
- start each coprocess immediately after the `preStart` but before starting our main application
- SIGTERM ourselves if any of the co-processes dies

cc @misterbisson @justenwalker @dekobon

justenwalker:
Need to discuss more about exit codes, stderr/out. Also if it expected by
behavior that when the coprocess dies, the container dies - I think most
people want it to be restarted.

tgross:
> Need to discuss more about exit codes, stderr/out.
> Also if it expected by behavior that when the coprocess dies, the container dies - I think most people want it to be restarted.

All our lifecycle hooks feed stdout/stderr thru the logger and we lose their exit codes.  As far as restarts, the co-process is secondary but what if the co-process can't be started because it's been misconfigured? We'd just end up restarting continuously and the main process might be in a bad state as a result.

justenwalker:
I wonder how the log will look with coprocessor stdout interleaved with the main app

tgross:
Funky, I imagine... the chance that the coprocess and the main app have similar logging formats is unfortunately pretty low. This might make it a pain for log parsing downstream.

tgross:
> but what if the co-process can't be started because it's been misconfigured? We'd just end up restarting continuously and the main process might be in a bad state as a result.

A possible option to solve this dilemma could be to have a configurable number of restarts on the coprocess:

```json
{"coprocess":
  [
    {
      "cmd": "consul-template",
      "restarts": 5
    }
  ]
}
```

[...]

Issue #162, opened 2016-05-20
Registering containers as separate nodes in consul


jgillis01:
Is there a way for containerpilot to register a container as its own node in consul? Currently, containerpilot registers a service against a particular node in consul (given a clustered setting). If that consul node becomes unavailable, we lose service health visibility until that consul node is restored.

nmarshall-cst:
+1 For me this occurs when Triton containers are removed or fail. I will also see multiple containers on the same node within the consul UI.

fastpopo:
Few days ago, I made some comtainer that include both the containerpilot and consul.

If you want register the container as a consul node, please try it.

In the dockerfile, entrypoint should contain some shell script that include running consul node for containerpilot's argument.

And the consul will execute itself as a foreground, so you can't execute some apllication what you want. So, you can use nohup for consul.

tgross:
@jgillis01 someone at a asked me about this offline recently and I'm going to admit I'm a little unclear as to what we're doing wrong here. But obviously the behavior you're describing isn't ideal. We're currently using [`Agent.ServiceRegister`](https://github.com/joyent/containerpilot/blob/3f7f6b83f9128b2a7b05117edebf69e4574b2efb/discovery/consul/consul.go#L120-L130) with the Consul API client, which corresponds to [`/v1/service/agent/register`](https://www.consul.io/docs/agent/http/agent.html#agent_service_register).

Should we be using [`/v1/catalog/register`](https://www.consul.io/docs/agent/http/catalog.html#catalog_register) (or in other words [`Catalog.Register`](https://godoc.org/github.com/hashicorp/consul/api#Catalog.Register))? The Consul documentation suggests otherwise but maybe I'm misreading it. Do you have a minimal test case I could use to exercise the bad behavior so we can verify this?

jgillis01:
@tgross,

The documentation seems to suggest that the agent endpoint is used to interact with a consul agent running locally on the node:
> The Agent endpoints are used to interact with the local Consul agent.

I agree the catalog endpoint should not be used to register services due to the health check restrictions:
>If the Check key is provided, a health check will also be registered. Note: this register API manipulates the health check entry in the Catalog, but it does not setup the script, TTL, or HTTP check to monitor the node's health. To truly enable a new health check, the check must either be provided in agent configuration or set via the agent endpoint.

The following steps can be used to reproduce the behavior I am experiencing:

1. Create a consul cluster with 4 nodes (-bootstrap-expect 3).
[...]

PR #176, opened 2016-06-08
Implementation of co-process hooks

tgross:
For https://github.com/joyent/containerpilot/issues/175, inspired by discussions in https://github.com/joyent/containerpilot/issues/162.

This PR implements user-defined hooks for coprocesses: processes that are started but remain resident rather than exiting immediately. These can be used to run local Consul agents, daemons like confd or consul-template, log shippers in non-Docker environments, etc.


coprocesses/coprocess.go
```diff
@@ -0,0 +1,105 @@
...
+		coprocess.Name = strings.Join(coprocess.Args, " ")
+	}
+	return nil
+}
+
+// Start runs the coprocess
+func (coprocess *Coprocess) Start() {
+	log.Debugf("coprocess[%s].Start", coprocess.Name)
+	fields := log.Fields{"process": "coprocess", "coprocess": coprocess.Name}
+	stdout := utils.NewLogWriter(fields, log.InfoLevel)
```
justenwalker: Using LogWriter will take care of that weird interleaving problem I mentioned before.  Good catch.

coprocesses/coprocess.go
```diff
@@ -0,0 +1,105 @@
...
+	stderr := utils.NewLogWriter(fields, log.DebugLevel)
+	coprocess.logWriters = []io.WriteCloser{stdout, stderr}
+	defer coprocess.closeLogs()
+
+	for {
+		if coprocess.Restarts < 0 {
```
justenwalker: This logic might have a problem.  If the Restarts counter is not reset after a successful start, then a process that sporadically dies and is restarted successfully may eventually fail to restart in the future.  

Defining success here could be tricky though... Some options:

1. coprocesses have health checks. success means that the health check passed.
2. success means the coprocess is still running after some delay

tgross: > then a process that sporadically dies and is restarted successfully may eventually fail to restart in the future.

If you've got something that flaky then I you probably want it to always restart, I guess? Maybe we should make the Restarts optional? How would you recommend we handle that from a configuration standpoint. My understanding is that we end up with a value of 0 if it's omitted, which we could special case but then we wouldn't have a way to say "never restart"

justenwalker: In the first pass you mentioned that a dying coprocess kills the container - Which is probably preferable to restarting it a finite number of times. The traditional way is to treat `-1` special like "always restart" and then "0" means what it says on the tin. 

However, I hate doing that. I'd try to avoid doing the whole special values case if we can just put another config in.  I'd take a page out of the [upstart](http://upstart.ubuntu.com/cookbook/#respawn-limit) cook-book and make two configs:

- restart: true / fase
- restartLimit: 1+ | "unlimited"

Since we are doing raw configs, we can detect the type of `restartLimit` and actually get a nice self-documenting value.

justenwalker: We could also make the restartLimit more human readable and just always accept a string argument

- `restartLimit: "3"`
- `restartLimit: "3 wait 5s"`
- `restartLimit: "unlimited"`

tgross: If we're going to accept a multi-type value for `restartLimit`, we could just default to no restart and avoid having the extra `restart: true|false`

tgross: Ok, I've taken another pass at that.

coprocesses/coprocess.go
```diff
@@ -0,0 +1,148 @@
...
+			return fmt.Errorf("`restartLimit` field only takes positive integers or `unlimited`")
+		}
+	case int:
+		if t > 0 {
+			coprocess.restart = true
+			coprocess.restartLimit = t
+			coprocess.restartsRemain = t
+			break
+		} else {
+			// otherwise ignore it; can't fallthrough on type switch
```
tgross: Actually these are the zero-values for these struct fields... I kinda like that it's explicit here but maybe we should omit it and then have the `default` type case raise a parse error?

coprocesses/coprocess.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	switch t := coprocess.RestartLimit.(type) {
+	case string:
+		if t == "unlimited" {
+			coprocess.restart = true
+			coprocess.restartLimit = unlimitedRestarts
+			coprocess.restartsRemain = unlimitedRestarts
+		} else {
+			return fmt.Errorf("`restartLimit` field only takes positive integers or `unlimited`")
```
justenwalker: I'd also put a case in for "never" to make it easy to generate/template the config. Omission as the **only way** to get the desired behavior is inconvenient.

coprocesses/coprocess.go
justenwalker: Should probably check for field omission before type asserting - and then raise an error in the default case as you said.

coprocesses/coprocess.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	case int:
```
justenwalker: Also, for consistency - we should support strings with integer contents like we do for stopTimeout, etc...

tgross: Oh, right. I'd forgotten about that.

coprocesses/coprocess.go
tgross: Sounds good on all the above.

tgross:
I've updated the config parser to accept `"never"` as well as strings-that-parse-as-ints. And I've added a bunch of table-based tests for the restart parsing and behavior.

justenwalker:
Another nitpick - probably my fault too:

```
restartLimit: never
restartLimit: unlimited
restartLimit: 4
```

Doesn't read well to me.  especially never (even though we should have it). Not really a huge deal, but maybe there is a better name for `never` or for `restartLimit`

tgross:
Maybe calling the field `restarts` would be better? That way it's grammatical either way.

```
restarts: never
restarts: unlimited
restarts: 4
```

justenwalker:
Yeah that works for me

tgross:
Done on the field rename. That reads a lot nicer in the code too.

Next on the list are some integration tests and documentation updates.

tgross:
I've worked up an integration test that exercises the restart behavior as well as ensures that the restart limit is being reset after a SIGHUP to reload the configuration. I've also added some more testing of the raw config load for this package, because mapstructure had a couple little surprises for me as described in https://github.com/joyent/containerpilot/pull/176/commits/771f25c6cf2cfb2a58c000f5e0e0a1cdb850831d

Next up is to document it all.

tgross:
@misterbisson I've pushed up a bunch of associated documentation changes into this PR now too.

tgross:
Just pushed a big ol' rebase so that I've got the implementation (+ unit tests), the integration tests, and the documentation all as clean commits.

tgross:
@justenwalker any last comments on this one?

justenwalker:
@tgross Just that you've done an excellent job. LGTM.

adamlc:
Massive thanks for this, very useful!

For those that have apps that need to wait for coprocesses to do something before startong I found this tool called [wait-on](https://github.com/jeffbski/wait-on) that makes this really easy. Not sure if it'll affect signals, but I'm using directly in my node app anyway :)

PR #183, opened 2016-07-06
Set GOEXPERIMENT for go1.6 framepointer support

tgross:
In the upcoming golang 1.7, framepointers will be [enabled by default](https://github.com/golang/go/issues/15840). But after reading that GH issue I've realized that in the meantime, we can get this behavior today by pinning our go toolchain version -- which we should have been doing all along -- and adding an environment variable to our build container.

cc @pfmooney @misterbisson

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
-FROM golang
+FROM golang:1.6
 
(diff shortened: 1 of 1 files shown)


misterbisson:
:hotdog:

Issue #139, opened 2016-04-25
Have health checks timeout


tgross:
I've run into an issue where a health check hangs because of something unrelated to ContainerPilot. The underlying problem seems to right itself, but because the health check hangs the container never becomes healthy.

If we port the [timeout functionality from the new tasks feature](https://github.com/joyent/containerpilot/blob/cea9070df36d4bd5795142af2d2dbcb4d17e4035/tasks/tasks.go#L128-L141) to health checks and other user-defined behaviors we could avoid situations where a machine that can become healthy if the health check is given a chance to re-run doesn't do so.

justenwalker:
@tgross Should that be a feature of all `Pollables` ?

tgross:
Yeah, I think that makes sense.

tgross:
Would be worth looking at https://github.com/docker/docker/pull/23218/files#diff-e9ee36fa14ba54512c7e7b0918c7b1d0R147 for what looks to be a pretty robust timeout on child processes done by the Docker team.

tgross:
I'm going to take this on and move some of the exec/wait stuff current in utils into a `commands` package that wraps the `exec.Cmd` interface with the timeout behaviors we want here.

javabean:
In the mean time, you can use the [timeout](http://linux.die.net/man/1/timeout) unix utility in your health check command.

tgross:
I've laid the groundwork for this in #184 so now I just need to add the configuration options for it.

tgross:
TravisCI is showing a failing integration test that works locally and worked in the PR but not the most recent master build: https://travis-ci.org/joyent/containerpilot/builds/149271893

```
2016/08/02 18:50:46 TEST: test_tasks
Creating testtasks_consul_1
testtasks_consul_1 is up-to-date
Creating testtasks_app_1
Expected task1 to never time out: got 1
No stopped containers
2016/08/02 18:50:54 FAIL: test_tasks
```

I kicked off the build again because I think the failure is transient but that's no good so I need to figure out what I broke.

tgross:
[...]

PR #184, opened 2016-07-13
Refactoring to allow for pollable timeout

tgross:
For https://github.com/joyent/containerpilot/issues/139

This PR lifts most of the existing logic from the work @justenwalker did in the `tasks` package so it can be reused to give us a single abstraction for running external commands for user-defined hooks for both blocking actions like the main app and co-processes (via `RunAndWait`), and non-blocking actions that should timeout (via `RunWithTimeout`).

So far I've ported the `services` package to use it. This is work in progress and I need to move some tests around and then continue the implementation into the other packages. From here it's mostly no new code though, just swapping out `exec.Cmd` for `command.Command` in places and deleting redundancies.


tgross:
One of the little bonuses of this is that we create the `exec.Cmd` fresh from the parsed args when we make the external call only, so we parse args once but there's no funky logic with trying to recreate `exec.Cmd` in the right spot, which has been a source of bugs in the past.

tgross:
Current build passes tests but:

> /cp/src/github.com/joyent/containerpilot/commands/args.go:48:1: comment on exported function ArgsToCommand should be of the form "ArgsToCommand ..."
/cp/src/github.com/joyent/containerpilot/commands/commands.go:13:6: exported type Command should have comment or be unexported
/cp/src/github.com/joyent/containerpilot/commands/commands.go:24:1: exported function NewCommand should have comment or be unexported
/cp/src/github.com/joyent/containerpilot/commands/commands.go:47:10: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
Found 4 lint suggestions; failing.
make: *** [lint] Error 1
The command "make lint build" exited with 2.

Curses, foiled by the linter again!

tgross:
This is in a working state now but it needs pass to clean up dead code and then some ergonomic tweaks in the `commands` package API.

There is one behavioral difference between this implementation and the existing, which is that debug logs are a lot more verbose. The verbosity we have for `tasks` will now be seen for all the other hooks, which should improve the supportability of ContainerPilot.

tgross:
> some ergonomic tweaks in the commands package API,

Particularly, as I'm doing code cleanup here, I'm thinking it may have been a mistake to make `RunAndWait`, `RunAndWaitForOutput` and `RunWithTimeout` methods rather than functions because it means we've now got nil checks all over the code base instead of neatly wrapping them in the functions. I may undo that little bit here.

commands/commands.go
```diff
@@ -0,0 +1,172 @@
...
+	"os/exec"
+	"syscall"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/joyent/containerpilot/utils"
+)
+
+// Command wraps an os/exec.Cmd with a timeout, logging, and arg parsing.
+type Command struct {
```
justenwalker: I'm really digging this new Command abstraction - good choice

justenwalker:
> I may undo that little bit here because that's the sort of thing we will definitely miss in the future and create a bug.

Upon further inspection, I agree.  Much easier to do the nil checks inside a function call like `Run(command)` rather than `command.Run()` since you don't have to worry if command is nil in the former case.

tgross:
> Much easier to do the nil checks inside a function call like Run(command) rather than command.Run() since you don't have to worry if command is nil in the former case every time you call it.

Yeah and I just found another place I missed it, too! Will do.

tgross:
@justenwalker I've made that change. See what you think.

tgross:
I've updated the PR title to reflect that this only lays the groundwork for us to have timeouts. It doesn't actually yet implement them except for the existing timeouts option in the `tasks` package. I'd like to do that work under separate PR.

tgross:
@justenwalker did you have any more thoughts on this PR?

justenwalker:
LGTM

PR #186, opened 2016-07-18
SIGINT Termination and Co-Process Restart Suspension

davefinster:
Not sure what the wider repercussions would be on enabling this so completely open to feedback. 

SIGINT Termination:
Because of the issues around service registration with multiple instances of Consul being available behind a single DNS hostname (where the same service could potentially become registered multiple times), we have installed Consul into our container image via co-processes. We use the same image for development purposes locally (typically through `docker run -it`) and one would typically Ctrl-C or otherwise issue a SIGINT to break out of this. Containerpilot at the moment will only listen for/forward on SIGTERM. This change causes it to perform the same action for SIGINT as well. 

If it isn't acceptable to make this a hard-coded setting, perhaps an option can be added into the configuration file such as `{"extraTermSigs": ["SIGINT"]}` to control the behaviour

Co-Process Restart Suspension:
One issue that we also observed given we were embedding Consul into the container is the rather large list of failed Consul nodes that would amass over time. In an attempt to have the Consul agent 'leave' the cluster, it would shut itself down only to be restarted by ContainerPilot and re-join the cluster, even while CP was in the process of terminating. I've simply added a method that suspends the automated restarting of co-processes if ContainerPilot is in the process of terminating.

### DIFF
--- a/coprocesses/coprocess.go
+++ b/coprocesses/coprocess.go
@@ -146,6 +146,10 @@ func (coprocess *Coprocess) Start() {
 	}
 }
 
+func (coprocess *Coprocess) SuspendRestart() {
+	coprocess.restart = false
+} 
+
 // Stop kills a running coprocess
 func (coprocess *Coprocess) Stop() {
 	log.Debugf("coprocess[%s].Stop", coprocess.Name)
--- a/coprocesses/coprocess.go
+++ b/coprocesses/coprocess.go
@@ -146,6 +146,7 @@ func (coprocess *Coprocess) Start() {
 	}
 }
 
+// Stops the coprocess from being restarted if it dies
 func (coprocess *Coprocess) SuspendRestart() {
 	coprocess.restart = false
 } 
--- a/coprocesses/coprocess.go
+++ b/coprocesses/coprocess.go
@@ -146,7 +146,7 @@ func (coprocess *Coprocess) Start() {
 	}
 }
 
-// Sets whether a process from being restarted if it dies ignoring restart count
+// ShouldRestart sets whether a process from being restarted if it dies ignoring restart count
 func (coprocess *Coprocess) ShouldRestart(shouldRestart bool) {
 	coprocess.restart = shouldRestart
 } 
--- a/coprocesses/coprocess.go
+++ b/coprocesses/coprocess.go
@@ -146,9 +146,9 @@ func (coprocess *Coprocess) Start() {
 	}
 }
 
-// Stops the coprocess from being restarted if it dies
-func (coprocess *Coprocess) SuspendRestart() {
-	coprocess.restart = false
+// Sets whether a process from being restarted if it dies ignoring restart count
+func (coprocess *Coprocess) ShouldRestart(shouldRestart bool) {
+	coprocess.restart = shouldRestart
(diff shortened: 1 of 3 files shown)


nmarshall-cst:
_+1_

tgross:
@davefinster I definitely agree with the addition of a SIGINT handler. Some comments on implementation details:
- Why add the `ShouldRestart` method rather than make a call to the `Stop` method instead? That should prevent it from restarting. Maybe I'm missing something though.
- We've found that any time we change anything in the signal handlers that we *definitely* need thorough unit and integration testing, because it's super easy to run into corner cases. Can you flesh out tests now that the overall idea has been given a :+1:?

tgross:
@nmarshall-cst I hate to sound like a jerk but you would mind using the GitHub reactions rather than adding a +1 comment in the future? This is a small project now but if we set the expectation we can avoid having giant emoji/gif-filled +1 threads down the road.

davefinster:
@tgross Sure, I'll look into fleshing out some testing for the SIGINT handler. I'm probably leaning more towards adding a configuration parameter that triggers whether CP listens for it, just to limit the potential flow on effects of another signal suddenly being captured. 

As for the ShouldRestart method vs Stop call, my thinking is that the co-process may actually be required during shutdown for some reason or the user might want to utilise a custom shutdown method rather then outright killing the co-process. 

In the case of a Consul co-process, it is used by ContainerPilot to de-register the service. The user might also desire to have the Consul agent gracefully leave the cluster so it is marked as 'left' rather than failed. I've achieved this by adding ```"preStop":"consul leave"``` to the configuration file. In this instance, as part of the leave Consul will shut itself down. Without ```ShouldRestart```, ContainerPilot will observe the Consul exit and restart it causing Consul to re-join the cluster (potentially) just in time for the container to exit.

tgross:
> I'm probably leaning more towards adding a configuration parameter that triggers whether CP listens for it, just to limit the potential flow on effects of another signal suddenly being captured.

I'd rather we didn't have a flag for it just because it doubles the "test space" of configuration. I did some digging into SIGINT and can't really think of any example where it'd be the wrong thing to handle. The only other typical case where you'd see it is in a debugger, where the use case is similar.

> As for the ShouldRestart method vs Stop call, my thinking is that the co-process may actually be required during shutdown for some reason or the user might want to utilise a custom shutdown method rather then outright killing the co-process.

That's a really good point. I hadn't thought of that use case.

tgross:
> Sure, I'll look into fleshing out some testing for the SIGINT handler.

@davefinster any follow up on this?

tgross:
We still have no tests on this PR. I think the fix is important so I'll carry this PR once I have some time in the next week or two.

tgross:
We're going to keep this SIGINT capture behavior in v3 but eliminate the coprocess behaviors outright. I'll try to carry this PR once we get to that point of the roadmap so we can keep @davefinster's contributions if possible.

tgross:
The SIGINT handler was added in https://github.com/joyent/containerpilot/pull/339/commits/8d62ae0fff97acde7feb7b64f7cf4854f00af65a as part of #339. Going to close this PR.

Issue #185, opened 2016-07-14
Makefile vendor task fires inappropriately


tgross:
During `make test` we're still hitting the `vendor` task even though the `vendor/` directory and its contents have been written. This doesn't effect the TravisCI runs but adds maybe 5-10 sec to a local test run so it's annoying.

Totally non-critical but just dropping this issue here as a minor TODO at some point.

tgross:
Ah, easy enough. `vendor` is in the `.PHONY` block!

tgross:
This was merged.

PR #188, opened 2016-07-18
Remove vendor from .PHONY block in makefile

tgross:
For https://github.com/joyent/containerpilot/issues/185

This makes sure that we don't re-run the `vendor` target with its expensive and Internet-connecting `godep restore` on every unit test run under local development. Updating the vendor directory should be done by `make clean vendor`.


justenwalker:
LGTM

PR #191, opened 2016-07-23
Discussion : Convert from Godep to Glide?

justenwalker:
(Made this a PR so it would be easier to demo locally)

I know early on we chose `godep` as the way to manage dependencies, but [glide](https://glide.sh) seems to have implemented some pretty slick support for both resolving versions (Semver) and reproducibility (via lock files) - plus it supports importing out dependencies from godep - so the transition should be painless.

One thing that's pretty cool is that glide separates the resolved dependencies `glide.lock` from the desired state `glide.yaml` so we can specify dependencies with semver tags (if the project supports it) in `glide.yaml` and this translates to an actual commit ID in `glide.lock`

(below copied from the documentation)

# Glide

Glide is a package manager for Go that is conceptually similar to package managers for other languages such as Cargo for Rust, NPM for Node.js, Pip for Python, Bundler for Ruby, and so forth.

## Glide provides the following functionality:

- Records dependency information in a glide.yaml file. This includes a name, version or version range, version control information for private repos or when the type cannot be detected, and more.
- Tracks the specific revision each package is locked to in a glide.lock file. This enables reproducibly fetching the dependency tree.
- Works with Semantic Versions and Semantic Version ranges.
- Supports Git, Bzr, HG, and SVN. These are the same version control systems supported by go get.
- Utilizes vendor/ directories, known as the Vendor Experiment, so that different projects can have differing versions of the same dependencies.
- Allows for aliasing packages which is useful for working with forks.
- Import configuration from Godep, GPM, Gom, and GB.

## Example of the conversion process:

```
$ glide init
[INFO]  Generating a YAML configuration file and guessing the dependencies
[INFO]  Attempting to import from other package managers (use --skip-import to skip)
[INFO]  Found Godeps.json file in /Users/justen/go/src/github.com/joyent/containerpilot
[INFO]  --> Parsing Godeps metadata...
[INFO]  Importing Godep configuration
[WARN]  Godep uses commit id versions. Consider using Semantic Versions with Glide
[INFO]  --> Found imported reference to github.com/Sirupsen/logrus at revision be52937128b38f1d99787bb476c789e2af1147f1
[INFO]  --> Found imported reference to github.com/beorn7/perks at revision 3ac7bf7a47d159a033b107610db8a1b6575507a4
[INFO]  --> Found imported reference to github.com/coreos/etcd at revision 6bc9b6e4edf44b59bb99b3761e4461a1ba608b0b
[INFO]  --> Found imported reference to github.com/golang/protobuf at revision dda510ac0fd43b39770f22ac6260eb91d377bce3
[INFO]  --> Found imported reference to github.com/hashicorp/consul at revision 158eabdd6f2408067c1d7656fa10e49434f96480
[INFO]  --> Found imported reference to github.com/matttproud/golang_protobuf_extensions at revision d0c3fe89de86839aecf2e0579c40ba3bb336a453
[INFO]  --> Found imported reference to github.com/mitchellh/mapstructure at revision d2dd0262208475919e1a362f675cfc0e7c10e905
[INFO]  --> Found imported reference to github.com/prometheus/client_golang at revision 90c15b5efa0dc32a7d259234e02ac9a99e6d3b82
[INFO]  --> Found imported reference to github.com/prometheus/client_model at revision fa8ad6fec33561be4280a8f0514318c79d7f6cb6
[INFO]  --> Found imported reference to github.com/prometheus/common at revision 40456948a47496dc22168e6af39297a2f8fbf38c
[INFO]  --> Found imported reference to github.com/prometheus/procfs at revision 406e5b7bfd8201a36e2bb5f7bdae0b03380c2ce8
[INFO]  --> Found imported reference to golang.org/x/net at revision 7f88271ea9913b72aca44fa7fc8af919eacc17ce
[INFO]  --> Found imported reference to golang.org/x/sys at revision 50c6bc5e4292a1d4e65c6e9be5f53be28bcbe28e
[INFO]  Scanning code to look for dependencies not found in import
[INFO]  --> Found reference to github.com/samalba/dockerclient
[INFO]  Writing configuration file (glide.yaml)
[INFO]  Would you like Glide to help you find ways to improve your glide.yaml configuration?
[INFO]  If you want to revisit this step you can use the config-wizard command at any time.
[INFO]  Yes (Y) or No (N)?
Y
[INFO]  Looking for dependencies to make suggestions on
[INFO]  --> Scanning for dependencies not using version ranges
[INFO]  --> Scanning for dependencies using commit ids
[INFO]  Gathering information on each dependency
[INFO]  --> This may take a moment. Especially on a codebase with many dependencies
[INFO]  --> Gathering release information for dependencies
[INFO]  --> Looking for dependency imports where versions are commit ids
[INFO]  Here are some suggestions...
[INFO]  The package github.com/Sirupsen/logrus is currently set to use the version be52937128b38f1d99787bb476c789e2af1147f1.
[INFO]  There is an equivalent semantic version (http://semver.org) release of v0.9.0. Would
[INFO]  you like to use that instead? Yes (Y) or No (N)
Y
[INFO]  Would you like to remember the previous decision and apply it to future
[INFO]  dependencies? Yes (Y) or No (N)
Y
[INFO]  Updating github.com/Sirupsen/logrus to use the tag v0.9.0 instead of commit id be52937128b38f1d99787bb476c789e2af1147f1
[INFO]  The package github.com/Sirupsen/logrus appears to use semantic versions (http://semver.org).
[INFO]  Would you like to track the latest minor or patch releases (major.minor.patch)?
[INFO]  Tracking minor version releases would use '>= 0.9.0, < 1.0.0' ('^0.9.0'). Tracking patch version
[INFO]  releases would use '>= 0.9.0, < 0.10.0' ('~0.9.0'). For more information on Glide versions
[INFO]  and ranges see https://glide.sh/docs/versions
[INFO]  Minor (M), Patch (P), or Skip Ranges (S)?
P
[INFO]  Would you like to remember the previous decision and apply it to future
[INFO]  dependencies? Yes (Y) or No (N)
Y
[INFO]  Updating github.com/Sirupsen/logrus to use the range ~0.9.0 instead of commit id v0.9.0
[INFO]  Configuration changes have been made. Would you like to write these
[INFO]  changes to your configuration file? Yes (Y) or No (N)
Y
[INFO]  Writing updates to configuration file (glide.yaml)
[INFO]  You can now edit the glide.yaml file.:
[INFO]  --> For more information on versions and ranges see https://glide.sh/docs/versions/
[INFO]  --> For details on additional metadata see https://glide.sh/docs/glide.yaml/
```


tgross:
Ok, so what's the big advantage of moving to this over godeps?

Also what's the deal with the `subpackages` lines? They don't seem to be pinned in either the `glide.yaml` or `glide.lock`?

justenwalker:
> Ok, so what's the big advantage of moving to this over godeps?

Godeps is pretty much *just* the lock file.  

Some advantages:

1. Includes an actual dependency solver
2. Will notify you of conflicts (We need `A` and `B`, but `A` -> `C@aaaabbbb` and `B` -> `C@ccccdddd`) so you can make informed decisions about which overrides you want to make.
3. Support for semantic versioning - useful for repositories that are tagged appropriately and which we may want to auto upgrade minor versions - although `glide update`  is still required to change the `glide.lock` file. Just makes it easier to keep in-sync with upstream dependencies if we need to - while still allowing us to create reproducible builds.
4. Allows using forks of dependencies with `repo` , which may be quite useful if used in moderation.

(See [glide.yaml docs](https://glide.readthedocs.io/en/latest/glide.yaml/))


> Also what's the deal with the subpackages lines? They don't seem to be pinned in either the glide.yaml or glide.lock?

`subpackages` :  A record of packages being used within a repository. This does not include all packages within a repository but rather those being used.

Basically, when `glide install` is run, it doesn't pull entire repodown, but only the subpackages that are in-use.

justenwalker:
Also, a fallacious appeal to authority: https://github.com/uber/tchannel-go/commit/200e957b4c1f2c2ae6cc66e9783a097fc6f49b31

tgross:
I'm mostly convinced.

> subpackages : A record of packages being used within a repository. This does not include all packages within a repository but rather those being used.

I'm still concerned about this though. None of these subpackages are pinned as far as I can tell. How do we avoid breakage here?

justenwalker:
> I'm still concerned about this though. None of these subpackages are pinned as far as I can tell. How do we avoid breakage here?

They are pinned by the repo under which they are defined.  This is not a subpackage like "go get this other repo" - its more like a sub-folder within that pinned repository.

tgross:
> They are pinned by the repo under which they are defined.

But doesn't that assume that the top-level package has pinned their dependencies as well? Libraries generally aren't supposed to lock their dependencies under the Cargo/Ruby/Python model, only give minimum versions (which is why "commit your vendor folder" is such a horrible practice for libraries).

If we can't lock the dependencies of dependencies then we don't have a reproducible build for the binary, right?

justenwalker:
> But doesn't that assume that the top-level package has pinned their dependencies as well? Libraries generally aren't supposed to lock their dependencies under the Cargo/Ruby/Python model, only give minimum versions (which is why "commit your vendor folder" is such a horrible practice for libraries).


At this point though, I'm not sure we're talking about the same thing anymore. I'll try to clarify my understanding:

Subpackages are a list of which packages you are using within a dependency - not transient dependencies.

For instance, some project importing `github.com/joyent/containerpilot` may have:

```yaml
import:
- package: github.com/joyent/containerpilot
  version: ~1.0.0
  subpackages:
  - config
  - backends
```


Caveat - if a project *does* check in it's vendor dependencies - then those may also be considered as well.

tgross:
> Subpackages are a list of which packages you are using within a dependency - not transient dependencies.

_Oh_. Well now I get it! 😊

Ok I'm sold. I've also taken this branch for a spin myself and it looks like we don't have any spurious rebuilds from the Makefile targets. TravisCI is happy... let's merge it!

PR #198, opened 2016-08-05
Prevent polling and coprocesses from being started multiple times

ddunkin:
Prevent polling and coprocesses from being started multiple times if a reload is triggered during preStart.

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -47,6 +47,8 @@ type App struct {
 	signalLock     *sync.RWMutex
 	paused         bool
 	ConfigFlag     string
+	coworkLock     *sync.RWMutex
+	coworkStarted  bool
 }
 
 // EmptyApp creates an empty application
@@ -54,6 +56,7 @@ func EmptyApp() *App {
 	app := &App{}
 	app.maintModeLock = &sync.RWMutex{}
 	app.signalLock = &sync.RWMutex{}
+	app.coworkLock = &sync.RWMutex{}
 	return app
 }
 
@@ -146,8 +149,7 @@ func (a *App) Run() {
 			os.Exit(code)
 		}
 	}
-	a.handleCoprocesses()
-	a.handlePolling()
+	a.startCowork()
 
 	if a.Command != nil {
 		// Run our main application and capture its stdout/stderr.
@@ -270,12 +272,20 @@ func (a *App) Reload() error {
 		return err
 	}
 
-	a.stopPolling()
-	a.forAllServices(deregisterService)
-	a.stopCoprocesses()
+	a.stopCoworkAndDeregister()
 
 	a.load(newApp)
 	return nil
+}
+
+func (a *App) stopCoworkAndDeregister() {
+	a.coworkLock.Lock()
+	defer a.coworkLock.Unlock()
+
+	a.stopPolling()
+	a.forAllServices(deregisterService)
+	a.stopCoprocesses()
+	a.coworkStarted = false
 }
 
 func (a *App) load(newApp *App) {
@@ -291,8 +301,7 @@ func (a *App) load(newApp *App) {
 	a.Telemetry = newApp.Telemetry
 	a.Tasks = newApp.Tasks
 	a.Coprocesses = newApp.Coprocesses
-	a.handlePolling()
-	a.handleCoprocesses()
+	a.startCowork()
 }
 
 type serviceFunc func(service *services.Service)
@@ -300,6 +309,16 @@ type serviceFunc func(service *services.Service)
 func (a *App) forAllServices(fn serviceFunc) {
 	for _, service := range a.Services {
 		fn(service)
+	}
+}
+
+func (a *App) startCowork() {
+	a.coworkLock.Lock()
+	defer a.coworkLock.Unlock()
+	if !a.coworkStarted {
+		a.handleCoprocesses()
+		a.handlePolling()
+		a.coworkStarted = true
 	}
 }
 


tgross:
> Prevent polling and coprocesses from being started multiple times if a reload is triggered during preStart.

@ddunkin I just want to make sure I understand the problem as your describing it here:

The polling and coprocesses aren't started until after the `preStart` has exited (we block on the `preStart`). But if we receive a reload then the polling/coprocesses are started in `App.load`, which means when we exit `preStart` we load them again. This means we end up with 2 goroutines running for each pollable and two copies of each coprocess. Right?

If so, I think the approach you have here seems right. An alternate approach would be to start the signal handler after the `preStart` exits but then we'd lose the ability for the `preStart` to reconfigure ContainerPilot.

This fix needs extensive testing though. There are a *lot* of possible races and interleaving of events that we need to test to ensure we have no deadlocks and that we understand the secondary effects on heartbeat TTLs. Here's your happy path scenario where we start the co-work once and only once:

- App.Run()
  - start `preStart`
    - `preStart` fires SIGHUP
  - exit `preStart`
- signal handler->App.Reload()
    - App.Reload->a.stopCoworkAndDeregister() [obtain lock] [release lock]
    - App.Reload->a.load()
      - a.load()->a.startCowork() [obtain lock]
      - (all co-processes started)
- Service.SendHeartbeat()
- App.Run->a.startCowork() [block]
      - App.Reload->a.load()->a.startCowork exit [release lock]
- App.Run->a.startCowork() [obtains lock]
- App.Run->a.startCowork() exits w/ no-op


Here's a less happy path where we start and stop the co-work in a way that causes the application to become "unhealthy" after it's already been registered:

- App.Run()
  - start `preStart`
    - `preStart` fires SIGHUP
  - exit `preStart`
- App.Run->a.startCowork() [obtain lock]
      - (all co-processes started)
- Service.SendHeartbeat()
- signal handler->App.Reload()
    - App.Reload->a.stopCoworkAndDeregister() [blocks] [locks all signal handling!]
- App.Run->a.startCowork() exit [release lock]
- App.Reload->a.stopCoworkAndDeregister() [obtains lock]
  - (all co-processes stopped)
  - (service is deregistered)
  - App.Reload->a.load()
  - a.load()->a.startCowork() [obtain lock]
    - (all co-processes started)
  - App.Reload->a.load()->a.startCowork exit [release lock]
- Service.SendHeartbeat()

This looks to be *correct* behavior, to be sure. But if we're going to have this kind of behavior we need to test and document the heck out of it.

ddunkin:
@tgross you understand correctly. I'll work on a test case today.

ddunkin:
What is the problem in the second scenario laid out above? I don't see the difference between a SIGHUP from preStart being delayed and a SIGHUP from a health check or from an external sender (e.g. `docker kill --signal=HUP`).

As it is now (with or without this fix), there doesn't appear to be a way to force a reconfiguration from preStart to happen before anything else is started. If CP should support that behavior, it could reload based on a specific exit code from preStart.

ddunkin:
Side note: I came across the duplicate health-check bug while updating https://github.com/Faithlife/redis-autopilot to the latest ContainerPilot. I want to run a Consul agent in container because we have problems with the Redis container registering with a different one of our three Consul server agents depending on the order of addresses returned by DNS. However I see now that coprocesses are not started until after preStart where I do the master/slave configuration of Redis, which requires Consul, so I don't think I'll be able to use preStart anyway.

tgross:
> What is the problem in the second scenario laid out above?

It's not a problem, just a less obvious code path that causes the co-process to be restarted. I just want to make sure we're considering these cases so that (for example) the co-process restarts in that scenario work as we expect with their `restart` config.

> As it is now (with or without this fix), there doesn't appear to be a way to force a reconfiguration from preStart to happen before anything else is started.

Yeah that's kind of inherent to signal handling being async. Fair enough.

> However I see now that coprocesses are not started until after preStart where I do the master/slave configuration of Redis, which requires Consul, so I don't think I'll be able to use preStart anyway.

Yeah in cases like that you're going to want to put that into a periodic task or the health check, as we did in autopilotpattern/mysql. Unfortunately for every case like this, there's one where we want to configure to coprocess in `preStart` instead, so it's a matter of picking one behavior or the other.

ddunkin:
> Unfortunately for every case like this, there's one where we want to configure to coprocess in preStart instead, so it's a matter of picking one behavior or the other.

`prePreStart`? ;)

There's a problem with putting the preStart checks in the health check. In the case of a redis master that has failed and is restarting, and another master has taken over while the original was down, I want to reconfigure the original to no longer be the master before it starts back up. There's no good place to hook in to make that check now.

tgross:
> In the case of a redis master that has failed and is restarting, and another master has taken over while the original was down, I want to reconfigure the original to no longer be the master before it starts back up. 

We've done this sort of thing in autopilotpattern/mysql, using session keys in Consul to create a lock to coordinate the efforts of the instances. I'd recommend checking that out, although it's admittedly not trivial.

tgross:
We still have no tests on this PR. I think the fix is important so I'll carry this PR once I have some time in the next week or two.

ddunkin:
Sorry, I haven't had a chance to get back to this as it's no longer on my critical path. I may be able to get to it next week.

tgross:
No worries @ddunkin I probably wouldn't be able to get to it until after then anyways.

tgross:
We're going to keep fix this behavior in v3 but eliminate the coprocess behaviors outright. I'll try to carry this PR once we get to that point of the roadmap so we can keep @ddunkin's contributions if possible.

tgross:
We can actually close this as the work is covered in #290

Issue #201, opened 2016-08-10
Coprocess restarts unintentionally on config reload


tgross:
If we SIGHUP ContainerPilot then it appears that the coprocesses get shut down but then immediately restart. This means that they restart twice -- once because of the restart and once because of the application reload. Although [`Coprocess.Stop()`](https://github.com/joyent/containerpilot/blob/5ba69b4e042a497a30912445161a553c4667c8a2/coprocesses/coprocess.go#L138-L143) is designed to avoid this clearly there's a bug.

Here's an example of autopilotpattern/nginx receiving SIGHIP to PID1:

```
2016/08/10 18:33:30 Reloading configuration.
2016/08/10 18:33:30 parsed service discovery backend: consul
2016/08/10 18:33:30 Deregistering service: nginx
2016/08/10 18:33:30     2016/08/10 18:33:30 [INFO] agent: Deregistered service 'nginx-f71cfe3bb188'
2016/08/10 18:33:30     2016/08/10 18:33:30 [INFO] agent: Deregistered check 'nginx-f71cfe3bb188'
2016/08/10 18:33:30 Deregistering service: containerpilot
2016/08/10 18:33:30     2016/08/10 18:33:30 [INFO] agent: Deregistered service 'containerpilot-f71cfe3bb188'
2016/08/10 18:33:30 telemetry: shutdown received but currently a no-op
2016/08/10 18:33:30 coprocess[/usr/local/bin/consul agent -data-dir=/data -config-dir=/config -rejoin -retry-join consul -retry-max 10 -retry-interval 10s].Start
2016/08/10 18:33:30 coprocess[/usr/local/bin/consul agent -data-dir=/data -config-dir=/config -rejoin -retry-join consul -retry-max 10 -retry-interval 10s].Stop
2016/08/10 18:33:30 Killing coprocess /usr/local/bin/consul agent -data-dir=/data -config-dir=/config -rejoin -retry-join consul -retry-max 10 -retry-interval 10s: 22
2016/08/10 18:33:30     2016/08/10 18:33:30 [INFO] agent: Deregistered check 'containerpilot-f71cfe3bb188'
2016/08/10 18:33:30 coprocess[/usr/local/bin/consul agent -data-dir=/data -config-dir=/config -rejoin -retry-join consul -retry-max 10 -retry-interval 10s] exited: signal: killed
[...]

PR #203, opened 2016-08-10
Ensure coprocesses do not restart once stopped

tgross:
Fixes https://github.com/joyent/containerpilot/issues/201

- send SIGKILL synchronously to coprocess
- set `restartLimit` and `restart` flag to prevent a Coprocess that's
  received `Stop()` from restarting.

I've run the unit and integration tests and given this a try on autopilotpattern/nginx and my work on https://github.com/autopilotpattern/mysql/issues/43 as well.


misterbisson:
🏡 🚶

Issue #207, opened 2016-08-13
Main app output is lost in v2.4.0?


fitz123:
Hi there!

In v2.3.0 I could see in docker logs whatever sent into /dev/stdout /dev/stderr - doesn't matter was it coprocess or main app, but in new version It seems I lost this ability.
Using the same Dockerfile, just updated version stops showing logs from main process (nginx):
```
RUN ln -sf /dev/stdout /var/log/nginx/access.log \
	&& ln -sf /dev/stderr /var/log/nginx/error.log
```
This still works for the nginx as a co-processor though (same 2 redirections). 

Is it intended to be like that, is it broken or we have now an options to controll app output via containerpilot?
Using official nginx:stable as a base image. I think https://github.com/autopilotpattern/nginx actually broken now (no logs) as well, since my Dockerimage almost equal

ffmpeg container (`ENTRYPOINT  ["/usr/local/bin/containerpilot","ffmpeg"]`) shows no process output too. With v2.3.0 all application output is available with `docker logs`, with v2.4.0 only coprocess'es output is available

p.s. is there any way to suppress so loud telemetry checks output in the log?

Thanks!

adamlc:
I can confirm this too!

tgross:
Well that's a problem! Looking into this.

tgross:
I've been able to partially reproduce the problem. In https://github.com/tgross/nginx-autopilotpattern/tree/cp207_check I've rebuilt autopilotpattern/nginx with ContainerPilot 2.4.0. When I run that stack and check the Docker logs for the Nginx instance I get the following:

```
$ docker logs -f nginx_nginx_1
2016/08/15 12:50:59 preStart.RunAndWait start
2016/08/15 12:50:59 preStart.Cmd.Run
2016/08/15 12:50:59 2016/08/15 12:50:59 [ERR] (view) "service(example)" health services: error fetching: Unexpected response code: 500 (No cluster leader)
2016/08/15 12:50:59 2016/08/15 12:50:59 [ERR] (runner) watcher reported error: health services: error fetching: Unexpected response code: 500 (No cluster leader)
2016/08/15 12:50:59 Consul Template returned errors:
2016/08/15 12:50:59 preStart.RunAndWait end
2016/08/15 12:50:59 nginx.RunAndWait start
2016/08/15 12:50:59 nginx.Cmd.Run
2016/08/15 12:50:59 telemetry: Listening on 172.17.0.4:9090
[...]

PR #209, opened 2016-08-15
Attach stdout/stderr in RunAndWait

tgross:
For https://github.com/joyent/containerpilot/issues/207.

#184 caused a bug where stdout/stderr of the main application were no longer being captured. We need to attach stdout/stderr for a command whenever we don't have logging fields set. But we can't do this in `setUpCmd` because it breaks the case of `RunAndWaitForOutput` so I've added it to `RunAndWait` directly.

I've run unit and integration tests on this PR as well as added the build output to a build of autopilotpattern/nginx to [verify the behavior for log capture of the Nginx logs is correct](https://gist.github.com/tgross/fb8e41d7b8f2a6cae135a7a90464fd45).

cc @fitz123 @misterbisson @justenwalker

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -66,6 +66,10 @@ func RunAndWait(c *Command, fields log.Fields) (int, error) {
 	}
 	log.Debugf("%s.RunAndWait start", c.Name)
 	c.setUpCmd(fields)
+	if fields == nil {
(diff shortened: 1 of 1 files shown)


misterbisson:
Lgtm.

Issue #142, opened 2016-04-30
Support for ZooKeeper


giorgio-v:
Are there any plans to support ZK as an alternative to Consul?

justenwalker:
Consul is preferred since it has a more rich service-discovery story than etcd/zookeeper. Services are first-class objects and it even allows you to query them via DNS.

Zookeeper is just a filesystem, the implementation (like etcd) would be very ContainerPilot specific - there is no real standard for service discovery on these distributed filesystems.  I don't know if it is on the roadmap, but it would be possible. Probably by re-using some of the code in the etcd implementation.

I'm sure @tgross / @misterbisson would welcome a PR if you have some time to develop one 😃

giorgio-v:
I think can work on this. No guarantees about timeframe since I’m just getting started, but if it’s something deemed useful I’d be glad to help.

tgross:
I'd love to see a ZK backend!

I think @justenwalker has some good points about ZK's scope, but there are plenty of projects that use ZK under the hood for service discovery so I'd hope it's not impossible.

justenwalker:
One thing that would be good to discuss before going on is the library for communicating with zookeeper.  With etcd it was a natural choice to use the official SDK since etcd is also a golang project.  

- https://github.com/samuel/go-zookeeper
- https://wiki.ubuntu.com/gozk
- Others?

I haven't looked into them. go-zookeeper is the first result in google for `golang zookeeper`. Canonical also made a library called `gozk` although I don't know if it is actively developed. There may be others?

justenwalker:
you may want to wait for  #144 since it will affect how service discovery backends are implements. Hopefully making it a bit more straight forward. At least it will include a README.

giorgio-v:
OK, I’ll take a look at that PR in the meantime.

tgross:
Following up on the license issue; let's use `go-zookeeper` for the client library.

giorgio-v:
👍

justenwalker:
One other comment.

I've been looking into Hashicorp Vault recently, and I saw a [GH Issue](https://github.com/hashicorp/vault/issues/1367) about problems with `go-zookeeper`. Apparently, the project is either dead, or at least very stagnant. It's pretty much the only game in town for a go client, so I'm not sure we have much of a choice. Someone will likely need to fork/maintain it in the future, but so far no one has risen to the task.

talbright:
I don't think it has been, see my comments on hashicorp/vault#1367

[...]

PR #214, opened 2016-08-24
[WIP] Zookeeper support

giorgio-v:
This is a first cut of the ZooKeeper discovery module. 
#### What’s in there

The core functionality, roughly modeled after Etcd, should be complete and ready for review. Unit tests should be reasonably complete, in fact, I decide to keep in this PR the characterization tests for ZK to better explain / show the rationale behind the implementation. I think they could be deleted once this part is reviewed and, hopefully, approved. 

Currently, these tests run against a ZKv3.4.8 instance launched manually. 
#### What needs to be done
- [ ] Code review. Everything’s new for me in this project, so I expect it to be somewhat lengthy.
- [ ] Integrate the module with the rest of the system. Here I’d need some info for the things still needed.
- [ ] Docs, I guess
#### Additional info
- See #142 for more context
- ZK doesn’t have a node TTL mechanism, so I had to emulate it. However, this functionality is slated for 3.6. See https://issues.apache.org/jira/browse/ZOOKEEPER-2169.

### DIFF
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -164,7 +164,11 @@ func (conn ZooKeeper) createParentPath(path string) error {
 	newPath := ""
 	for i := range pathElements {
 		newPath = strings.Join([]string{newPath, sep, pathElements[i]}, "")
-		if _, err := conn.Client.Create(newPath, nil, 0, zk.WorldACL(zk.PermAll)); err != nil && err != zk.ErrNodeExists {
+		if _, err := conn.Client.Create(
+			newPath,
+			nil,
+			0,
+			zk.WorldACL(zk.PermAll)); err != nil && err != zk.ErrNodeExists {
 			return err
 		}
 	}
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -174,7 +174,11 @@ func (conn ZooKeeper) registerService(service *discovery.ServiceDefinition) erro
 	if exists, _, err := conn.Client.Exists(key); err != nil {
 		return err
 	} else if !exists {
-		if _, err := conn.Client.Create(key, []byte(value), 0, zk.WorldACL(zk.PermAll)); err != nil {
+		if _, err := conn.Client.Create(
+			key,
+			[]byte(value),
+			zk.FlagEphemeral,
+			zk.WorldACL(zk.PermAll)); err != nil {
 			return err
 		}
 		_, _, ch, err := conn.Client.GetW(key)
@@ -187,6 +191,7 @@ func (conn ZooKeeper) registerService(service *discovery.ServiceDefinition) erro
 				_, _, ch, err = conn.Client.GetW(ev.Path)
 				if err != nil {
 					log.Warning(err)
+					conn.Deregister(service)
 				}
 			case <-time.After(time.Duration(service.TTL) * time.Second):
 				log.Warningf("TTL expired, deregistering %s", key)
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -171,15 +171,23 @@ func (conn ZooKeeper) registerService(service *discovery.ServiceDefinition) erro
 		return err
 	}
 	value := encodeZooKeeperNodeValue(service)
-	if exists, _, _ := conn.Client.Exists(key); !exists {
+	if exists, _, err := conn.Client.Exists(key); err != nil {
+		return err
+	} else if !exists {
 		if _, err := conn.Client.Create(key, []byte(value), 0, zk.WorldACL(zk.PermAll)); err != nil {
 			return err
 		}
-		_, _, ch, _ := conn.Client.GetW(key)
+		_, _, ch, err := conn.Client.GetW(key)
+		if err != nil {
+			return err
+		}
 		go func() {
 			select {
 			case ev := <-ch:
-				_, _, ch, _ = conn.Client.GetW(ev.Path)
+				_, _, ch, err = conn.Client.GetW(ev.Path)
+				if err != nil {
+					log.Warning(err)
+				}
 			case <-time.After(time.Duration(service.TTL) * time.Second):
 				log.Warningf("TTL expired, deregistering %s", key)
 				conn.Deregister(service)
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -197,19 +197,19 @@ func (conn ZooKeeper) registerService(service *discovery.ServiceDefinition) erro
 	return nil
 }
 
-func encodeZooKeeperNodeValue(service *discovery.ServiceDefinition) string {
+func encodeZooKeeperNodeValue(service *discovery.ServiceDefinition) []byte {
 	node := &ServiceNode{
 		ID:      service.ID,
 		Name:    service.Name,
 		Address: service.IPAddress,
 		Port:    service.Port,
 	}
-	json, err := json.Marshal(&node)
+	result, err := json.Marshal(&node)
 	if err != nil {
 		log.Warnf("Unable to encode service: %s", err)
-		return ""
+		return nil
 	}
-	return string(json)
+	return result
 }
 
 func decodeZooKeeperNodeValue(rawValue []byte) (ServiceNode, error) {
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -97,13 +97,9 @@ func NewZooKeeperConfig(raw interface{}) (*ZooKeeper, error) {
 // Deregister removes this instance from the registry
 func (conn *ZooKeeper) Deregister(service *discovery.ServiceDefinition) {
 	key := conn.getNodeKey(service)
-	exists, _, err := conn.Client.Exists(key)
-	if err != nil {
-		log.Warnf("Error on deregistering %s. Node should exists: %s", key, err)
-	}
-	if exists {
-		log.Warnf("Deregistering %s", key)
-		conn.Client.Delete(conn.getNodeKey(service), -1)
+	log.Warnf("Deregistering %s", key)
+	if err := conn.Client.Delete(key, -1); err != nil {
+		log.Errorf("Error on deregistering %s: %s", key, err)
 	}
 }
 
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -17,6 +17,8 @@ func init() {
 	discovery.RegisterBackend("zookeeper", ConfigHook)
 }
 
+// ZooKeeper is a wrapper ZooKeeper connection. It also stores the
+// prefix under which ContainerPilot nodes will be registered.
 type ZooKeeper struct {
 	Connection *zk.Conn
 	Prefix     string
@@ -31,7 +33,7 @@ type ServiceNode struct {
 	Tags    []string `json:"tags"`
 }
 
-type ZookeeperRawConfig struct {
+type zookeeperRawConfig struct {
 	Endpoints interface{} `mapstructure:"endpoints"`
 	Prefix    string      `mapstructure:"prefix"`
 }
@@ -171,18 +173,17 @@ func (c ZooKeeper) registerService(service *discovery.ServiceDefinition) error {
 	if exists, _, _ := c.Connection.Exists(k); !exists {
 		if _, err := c.Connection.Create(k, []byte(value), 0, zk.WorldACL(zk.PermAll)); err != nil {
 			return err
-		} else {
-			_, _, ch, _ := c.Connection.GetW(k)
-			go func() {
-				select {
-				case ev := <-ch:
-					_, _, ch, _ = c.Connection.GetW(ev.Path)
-				case <-time.After(time.Duration(service.TTL) * time.Second):
-					log.Warningf("TTL expired, deregistering %s", k)
-					c.Deregister(service)
-				}
-			}()
 		}
+		_, _, ch, _ := c.Connection.GetW(k)
+		go func() {
+			select {
+			case ev := <-ch:
+				_, _, ch, _ = c.Connection.GetW(ev.Path)
+			case <-time.After(time.Duration(service.TTL) * time.Second):
+				log.Warningf("TTL expired, deregistering %s", k)
+				c.Deregister(service)
+			}
+		}()
 	}
 	return nil
 }
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -3,13 +3,13 @@ package zookeeper
 import (
 	"encoding/json"
 	"fmt"
-	log "github.com/Sirupsen/logrus"
 	"github.com/joyent/containerpilot/utils"
 	"github.com/samuel/go-zookeeper/zk"
 	"sort"
 	"strings"
 	"time"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/joyent/containerpilot/discovery"
 )
 
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -1,216 +1,214 @@
 package zookeeper
 
 import (
-  "encoding/json"
-  "fmt"
-  "strings"
-  "sort"
-  "time"
-  log "github.com/Sirupsen/logrus"
-  "github.com/samuel/go-zookeeper/zk"
-  "github.com/joyent/containerpilot/utils"
+	"encoding/json"
+	"fmt"
+	log "github.com/Sirupsen/logrus"
+	"github.com/joyent/containerpilot/utils"
+	"github.com/samuel/go-zookeeper/zk"
+	"sort"
+	"strings"
+	"time"
 
-  "github.com/joyent/containerpilot/discovery"
+	"github.com/joyent/containerpilot/discovery"
 )
 
 func init() {
-  discovery.RegisterBackend("zookeeper", ConfigHook)
+	discovery.RegisterBackend("zookeeper", ConfigHook)
 }
 
 type ZooKeeper struct {
-  Connection *zk.Conn;
-  Prefix string;
+	Connection *zk.Conn
+	Prefix     string
 }
 
 // ServiceNode is the serializable form of a ZooKeeper service record
 type ServiceNode struct {
-  ID      string   `json:"id"`
-  Name    string   `json:"name"`
-  Address string   `json:"address"`
-  Port    int      `json:"port"`
-  Tags    []string `json:"tags"`
+	ID      string   `json:"id"`
+	Name    string   `json:"name"`
+	Address string   `json:"address"`
+	Port    int      `json:"port"`
+	Tags    []string `json:"tags"`
 }
 
 type ZookeeperRawConfig struct {
-  Endpoints interface{} `mapstructure:"endpoints"`
-  Prefix    string      `mapstructure:"prefix"`
+	Endpoints interface{} `mapstructure:"endpoints"`
+	Prefix    string      `mapstructure:"prefix"`
 }
 
 // ConfigHook is the hook to register with the ZooKeeper backend
 func ConfigHook(raw interface{}) (discovery.ServiceBackend, error) {
-   return NewZooKeeperConfig(raw)
+	return NewZooKeeperConfig(raw)
 }
 
 // NewZooKeeperConfig creates a new service discovery backend for zookeeper
 func NewZooKeeperConfig(raw interface{}) (*ZooKeeper, error) {
-  zookeeper :=&ZooKeeper { Prefix: "/containerpilot" }
-  config := &struct {
-    Address string `mapstructure:"address"`
-  }{}
-  if err := utils.DecodeRaw(raw, &config); err != nil {
-    return nil, err
-  }
-  c, _, err := zk.Connect([]string{config.Address}, time.Second)
-  if err != nil {
-    return nil, err
-  }
-  zookeeper.Connection = c
-  return zookeeper, nil
+	zookeeper := &ZooKeeper{Prefix: "/containerpilot"}
+	config := &struct {
+		Address string `mapstructure:"address"`
+	}{}
+	if err := utils.DecodeRaw(raw, &config); err != nil {
+		return nil, err
+	}
+	c, _, err := zk.Connect([]string{config.Address}, time.Second)
+	if err != nil {
+		return nil, err
+	}
+	zookeeper.Connection = c
+	return zookeeper, nil
 }
 
 // Deregister removes this instance from the registry
 func (c *ZooKeeper) Deregister(service *discovery.ServiceDefinition) {
-  c.Connection.Delete(c.getNodeKey(service), -1)
+	c.Connection.Delete(c.getNodeKey(service), -1)
 }
 
 // MarkForMaintenance removes this instance from the registry
 func (c *ZooKeeper) MarkForMaintenance(service *discovery.ServiceDefinition) {
-  c.Deregister(service)
+	c.Deregister(service)
 }
 
 // SendHeartbeat refreshes the TTL of this associated zookeeper node
 func (c *ZooKeeper) SendHeartbeat(service *discovery.ServiceDefinition) {
-  if err := c.registerService(service); err != nil {
-    log.Warnf("Error registering service %s: %s", service.Name, err)
-  }
+	if err := c.registerService(service); err != nil {
+		log.Warnf("Error registering service %s: %s", service.Name, err)
+	}
 }
 
 func (c *ZooKeeper) parentPath(service *discovery.ServiceDefinition) string {
-  return fmt.Sprintf("%s/%s", c.Prefix, service.Name)
+	return fmt.Sprintf("%s/%s", c.Prefix, service.Name)
 }
 
 func (c *ZooKeeper) getNodeKey(service *discovery.ServiceDefinition) string {
-  return fmt.Sprintf("%s/%s", c.parentPath(service), service.ID)
+	return fmt.Sprintf("%s/%s", c.parentPath(service), service.ID)
 }
 
 func (c *ZooKeeper) getAppKey(appName string) string {
-  return fmt.Sprintf("%s/%s", c.Prefix, appName)
+	return fmt.Sprintf("%s/%s", c.Prefix, appName)
 }
 
 var zookeeperUpstreams = make(map[string][]ServiceNode)
 
 // CheckForUpstreamChanges checks another zookeeper node for changes
 func (c *ZooKeeper) CheckForUpstreamChanges(backendName, backendTag string) bool {
-  // TODO: is there a way to filter by tag in zookeeper?
-  services, err := c.getServices(backendName)
-  if err != nil {
-     log.Errorf("Failed to query %v: %s", backendName, err)
-    return false
-  }
-  didChange := zookeeperCompareForChange(zookeeperUpstreams[backendName], services)
-  if didChange || len(services) == 0 {
-    // We don't want to cause an onChange event the first time we read-in
-    // but we do want to make sure we've written the key for this map
-    zookeeperUpstreams[backendName] = services
-  }
-  return didChange
+	// TODO: is there a way to filter by tag in zookeeper?
+	services, err := c.getServices(backendName)
+	if err != nil {
+		log.Errorf("Failed to query %v: %s", backendName, err)
+		return false
+	}
+	didChange := zookeeperCompareForChange(zookeeperUpstreams[backendName], services)
+	if didChange || len(services) == 0 {
+		// We don't want to cause an onChange event the first time we read-in
+		// but we do want to make sure we've written the key for this map
+		zookeeperUpstreams[backendName] = services
+	}
+	return didChange
 }
 
 func (c *ZooKeeper) getServices(appName string) ([]ServiceNode, error) {
-  var services []ServiceNode
+	var services []ServiceNode
 
-  key := c.getAppKey(appName)
-  children, _, error := c.Connection.Children(key)
-  if(error != nil) {
-    return services, error
-  }
-  for i := range children {
-    path := fmt.Sprintf("%s/%s", key, children[i])
-    data, _, error := c.Connection.Get(path)
-    if(error != nil) {
-      return services, error
-    }
-    srv, error := decodeZooKeeperNodeValue(data)
-    if(error != nil) {
-      return services, error
-    }
-    services = append(services, srv)
-  }
-  return services, nil
+	key := c.getAppKey(appName)
+	children, _, error := c.Connection.Children(key)
+	if error != nil {
+		return services, error
+	}
+	for i := range children {
+		path := fmt.Sprintf("%s/%s", key, children[i])
+		data, _, error := c.Connection.Get(path)
+		if error != nil {
+			return services, error
+		}
+		srv, error := decodeZooKeeperNodeValue(data)
+		if error != nil {
+			return services, error
+		}
+		services = append(services, srv)
+	}
+	return services, nil
 }
 
 // Compare the two arrays to see if the address or port has changed
 // or if we've added or removed entries.
 func zookeeperCompareForChange(existing, new []ServiceNode) (changed bool) {
-  if len(existing) != len(new) {
-    return true
-  }
+	if len(existing) != len(new) {
+		return true
+	}
 
-  sort.Sort(ByZooKeeperServiceID(existing))
-  sort.Sort(ByZooKeeperServiceID(new))
-  for i, ex := range existing {
-    if ex.Address != new[i].Address ||
-      ex.Port != new[i].Port {
-      return true
-    }
-  }
-  return false
+	sort.Sort(ByZooKeeperServiceID(existing))
+	sort.Sort(ByZooKeeperServiceID(new))
+	for i, ex := range existing {
+		if ex.Address != new[i].Address ||
+			ex.Port != new[i].Port {
+			return true
+		}
+	}
+	return false
 }
 
 func (c ZooKeeper) createParentPath(path string) error {
-  pathElements := strings.Split(path, "/")[1:]
-  sep := "/"
-  newPath := ""
-  for i := range pathElements {
-    newPath = strings.Join([]string{newPath, sep, pathElements[i]}, "")
-    if exists, _, _ := c.Connection.Exists(newPath); !exists {
-      if _, err := c.Connection.Create(newPath, nil, 0, zk.WorldACL(zk.PermAll));
-			err != nil {
-        return err
-      }
-    }
-  }
-  return nil
+	pathElements := strings.Split(path, "/")[1:]
+	sep := "/"
+	newPath := ""
+	for i := range pathElements {
+		newPath = strings.Join([]string{newPath, sep, pathElements[i]}, "")
+		if exists, _, _ := c.Connection.Exists(newPath); !exists {
+			if _, err := c.Connection.Create(newPath, nil, 0, zk.WorldACL(zk.PermAll)); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
 
 func (c ZooKeeper) registerService(service *discovery.ServiceDefinition) error {
-  k := c.getNodeKey(service)
-  if err := c.createParentPath(c.parentPath(service)); err != nil {
-    return err
-  }
-  value := encodeZooKeeperNodeValue(service)
-  if exists, _, _ := c.Connection.Exists(k); !exists {
-    if _, err := c.Connection.Create(k, []byte(value), 0, zk.WorldACL(zk.PermAll));
-    err != nil {
-      return err
-    } else {
-      _, _, ch, _ := c.Connection.GetW(k)
-      go func() {
-        select {
-        case ev := <-ch:
-          _, _, ch, _ = c.Connection.GetW(ev.Path)
-        case <- time.After(time.Duration(service.TTL) * time.Second):
-          log.Warningf("TTL expired, deregistering %s", k)
-          c.Deregister(service)
-        }
-      }()
-    }
-  }
-  return nil
+	k := c.getNodeKey(service)
+	if err := c.createParentPath(c.parentPath(service)); err != nil {
+		return err
+	}
+	value := encodeZooKeeperNodeValue(service)
+	if exists, _, _ := c.Connection.Exists(k); !exists {
+		if _, err := c.Connection.Create(k, []byte(value), 0, zk.WorldACL(zk.PermAll)); err != nil {
+			return err
+		} else {
+			_, _, ch, _ := c.Connection.GetW(k)
+			go func() {
+				select {
+				case ev := <-ch:
+					_, _, ch, _ = c.Connection.GetW(ev.Path)
+				case <-time.After(time.Duration(service.TTL) * time.Second):
+					log.Warningf("TTL expired, deregistering %s", k)
+					c.Deregister(service)
+				}
+			}()
+		}
+	}
+	return nil
 }
 
 func encodeZooKeeperNodeValue(service *discovery.ServiceDefinition) string {
-  node := &ServiceNode{
-    ID:      service.ID,
-    Name:    service.Name,
-    Address: service.IPAddress,
-    Port:    service.Port,
-  }
-  json, err := json.Marshal(&node)
-  if err != nil {
-    log.Warnf("Unable to encode service: %s", err)
-    return ""
-  }
-  return string(json)
+	node := &ServiceNode{
+		ID:      service.ID,
+		Name:    service.Name,
+		Address: service.IPAddress,
+		Port:    service.Port,
+	}
+	json, err := json.Marshal(&node)
+	if err != nil {
+		log.Warnf("Unable to encode service: %s", err)
+		return ""
+	}
+	return string(json)
 }
 
 func decodeZooKeeperNodeValue(rawValue []byte) (ServiceNode, error) {
-  service := ServiceNode{}
-  err := json.Unmarshal(rawValue, &service)
-  if err != nil {
-    return service, err
-  }
-  return service, nil
+	service := ServiceNode{}
+	err := json.Unmarshal(rawValue, &service)
+	if err != nil {
+		return service, err
+	}
+	return service, nil
 }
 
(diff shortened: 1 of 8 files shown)


giorgio-v:
Could someone explain how to fix the `make vendor` step? Sorry for the stupid question, but I don’t know how the build works. Thank you.

tgross:
Thanks for the contribution @giorgio-v! I'm taking a look thru this now and will have some detailed comments on the code shortly.

> ZK doesn’t have a node TTL mechanism, so I had to emulate it.

I'm a little concerned about this from a design standpoint. With the existing discovery backends we mark an instance as failed if:

- The health check fails
- The application crashes (this forces a ContainerPilot exit and therefore a TTL expiry)
- The underlying host crashes (TTL expires)
- There's a net split between this container and the discovery backend (TTL expires)

With the emulation here we only get the first of these 4 options. Do we know when ZK3.6 is supposed to land? Do we feel comfortable with having a backend that doesn't completely work as intended (cc @justenwalker @misterbisson)?

tgross:
> Could someone explain how to fix the make vendor step?

The error you're getting in the [build](https://travis-ci.org/joyent/containerpilot/builds/155849214) is:
> discovery/zookeeper/zookeeper.go:8:2: cannot find package "github.com/samuel/go-zookeeper/zk"

When you're specifying the package in the [glide.yaml](https://github.com/joyent/containerpilot/pull/214/files#diff-395f41b2a27b70ce44399c91c82158c2R58) file you need to specify the subpackages as well. So the section you've added should read:

```yaml
- package: github.com/samuel/go-zookeeper
  version: 8c6a1d79a44d2742da3660540d1fd43f4ddcfc8c
  subpackages:
  - zk
```

I can assure you I found this terminology [just as confusing myself](https://github.com/joyent/containerpilot/pull/191#issuecomment-236903098)! 😀

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,220 @@
+package zookeeper
+
+import (
+	"encoding/json"
+	"fmt"
+	log "github.com/Sirupsen/logrus"
```
tgross: Minor golang style thing: usually you put the third-party packages last with a line between them and the stdlib packages, so that gofmt sorts them separately.

justenwalker:
> Do we feel comfortable with having a backend that doesn't completely work as intended

My 2 cents is that we should try using Ephemeral ZKNodes instead of TTLs since that is the intended primitive supported by Zookeeper at the moment.

justenwalker:
> Do we feel comfortable with having a backend that doesn't completely work as intended

If it doesn't expire the service when the container dies unexpectedly - then that's a bug.  I think if we're using Ephemeral ZKNodes that won't be a problem though.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,220 @@
...
+func (c *ZooKeeper) Deregister(service *discovery.ServiceDefinition) {
+	c.Connection.Delete(c.getNodeKey(service), -1)
+}
+
+// MarkForMaintenance removes this instance from the registry
+func (c *ZooKeeper) MarkForMaintenance(service *discovery.ServiceDefinition) {
+	c.Deregister(service)
+}
+
+// SendHeartbeat refreshes the TTL of this associated zookeeper node
```
tgross: We don't have an actual TTL, so we should adjust this comment so that it isn't misleading.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func (c ZooKeeper) createParentPath(path string) error {
+	pathElements := strings.Split(path, "/")[1:]
+	sep := "/"
+	newPath := ""
+	for i := range pathElements {
+		newPath = strings.Join([]string{newPath, sep, pathElements[i]}, "")
+		if exists, _, _ := c.Connection.Exists(newPath); !exists {
+			if _, err := c.Connection.Create(newPath, nil, 0, zk.WorldACL(zk.PermAll)); err != nil {
```
tgross: Is this race a problem? We're checking whether the path exists and then creating it if it doesn't. If I have two instances of a service I can have a [TOCTOU](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use) error here and then we are going to get a noisy error in the caller.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,220 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (c ZooKeeper) registerService(service *discovery.ServiceDefinition) error {
+	k := c.getNodeKey(service)
```
tgross: I know the golang folks love their single-letter variables but I like to try to avoid them except in the case of the method receiver.

discovery/zookeeper/zookeeper.go
justenwalker: IIRC, zookeeper supports transactions - which, if implementer here, would make this atomic.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,220 @@
...
+	if err := c.createParentPath(c.parentPath(service)); err != nil {
+		return err
+	}
+	value := encodeZooKeeperNodeValue(service)
+	if exists, _, _ := c.Connection.Exists(k); !exists {
+		if _, err := c.Connection.Create(k, []byte(value), 0, zk.WorldACL(zk.PermAll)); err != nil {
+			return err
+		}
+		_, _, ch, _ := c.Connection.GetW(k)
```
tgross: I may be misunderstanding the code here -- can you explain the intent of what we're doing in this section? It looks like you're setting a watch on the key and then starting a new goroutine that creates a new watch on each event received. This seems like we're going to be firing up a new goroutine on every health check poll and then deregistering when the TTL expires but I can't figure out how we're passing that goroutine a message not to deregister.

I also find it troubling the amount of `_` we have here where we're throwing away `error` objects.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,220 @@
...
+		Name:    service.Name,
+		Address: service.IPAddress,
+		Port:    service.Port,
+	}
+	json, err := json.Marshal(&node)
+	if err != nil {
+		log.Warnf("Unable to encode service: %s", err)
+		return ""
+	}
+	return string(json)
```
tgross: You're coercing the `[]byte` to a string but then coercing it back to `[]byte` when you use it in `c.Connection.Create`, etc. You can just have this function return `[]byte` and be good-to-go.

discovery/zookeeper/zookeeper_test.go
```diff
@@ -0,0 +1,454 @@
...
+	return &discovery.ServiceDefinition{
+		ID:        id,
+		Name:      "my-service",
+		IPAddress: "192.168.1.1",
+		TTL:       1,
+		Port:      9000,
+	}
+}
+
+func sut() *ZooKeeper {
```
tgross: "sut"? What's this name mean?

discovery/zookeeper/zookeeper_test.go
```diff
@@ -0,0 +1,454 @@
...
+		"/node",                     // path
+		[]byte("value, can be nil"), // data
+		zk.FlagEphemeral,            // flags
+		zk.WorldACL(zk.PermAll))     // ACL
+	if err != nil {
+		t.Fatalf("Unable to create node")
+	}
+}
+
+func TestCrateIntermediateEphemeralNode(t *testing.T) {
```
tgross: > TestCreateIntermediateEphemeralNode

spelling error

discovery/zookeeper/zookeeper_test.go
```diff
@@ -0,0 +1,454 @@
...
+import (
+	"github.com/joyent/containerpilot/discovery"
+	"github.com/samuel/go-zookeeper/zk"
+	"testing"
+	"time"
+)
+
+// Factories, utilities
+func zkConnection() *zk.Conn {
+	c, _, _ := zk.Connect([]string{"127.0.0.1"}, time.Second)
```
tgross: Our Makefile will need a Zookeeper target container, won't it?

giorgio-v: Yes, I’ll work on this part once after sorting out the first phase of the review.

discovery/zookeeper/zookeeper_test.go
giorgio-v: System Under Test? Should I change it? No problem, of course.

discovery/zookeeper/zookeeper.go
giorgio-v: Agreed. I didn’t find a guideline about this or maybe I have missed it?

discovery/zookeeper/zookeeper.go
giorgio-v: I’ll add the missing error checks. As for the checks, the watcher is set only if `k` does not exist. Does this clarify your doubts a bit?

discovery/zookeeper/zookeeper.go
giorgio-v: I’ll check later and let you know.

giorgio-v:
@tgross @justenwalker Thanks for the great feedback. I’ll reply later with more details and then update the PR.

discovery/zookeeper/zookeeper.go
giorgio-v: As far as I can tell, there’s no transaction support in ZK out of the box. @justenwalker do you have any ref handy?

giorgio-v: Got it. You're talking about the Java library. I'll check what's available on the Go side. Thanks for the pointer.

giorgio-v:
I _think_ I have addressed all the observations, but I’ll the esteemed reviewers have their say. 😉 If that’s the case, I’ll work on the test integration, so that we can check build results.

giorgio-v:
Addendum. @justenwalker suggested using transactions (or rather, multi operations) to solve races during the path / node creation. I've looked into that and I think it doesn't fit the context because trying to create an existing node is an error and, as far as I know, operations are guaranteed to execute atomically, but they are not isolated. 

An alternative solution could be to encode the namespacing into the name of the registered service. I haven't explored potential ramifications but it should work.

glide.yaml
```diff
@@ -55,3 +55,7 @@ import:
   subpackages:
   - unix
 - package: github.com/samalba/dockerclient
+- package: github.com/samuel/go-zookeeper
+  version: 8c6a1d79a44d2742da3660540d1fd43f4ddcfc8c
+  subpackages:
+    - zk
```
tgross: TravisCI is still giving us this error:

```
docker run --rm --link containerpilot_consul:consul --link containerpilot_etcd:etcd -v /home/travis/gopath/src/github.com/joyent/containerpilot/vendor:/go/src -v /home/travis/gopath/src/github.com/joyent/containerpilot:/cp/src/github.com/joyent/containerpilot -w /cp/src/github.com/joyent/containerpilot containerpilot_build bash ./scripts/unit_test.sh
discovery/zookeeper/zookeeper.go:7:2: cannot find package "github.com/samuel/go-zookeeper/zk" in any of:
	/cp/src/github.com/joyent/containerpilot/vendor/github.com/samuel/go-zookeeper/zk (vendor tree)
	/usr/local/go/src/github.com/samuel/go-zookeeper/zk (from $GOROOT)
	/go/src/github.com/samuel/go-zookeeper/zk (from $GOPATH)
	/cp/src/github.com/samuel/go-zookeeper/zk
make: *** [test] Error 1
```

@justenwalker any thoughts? It looks like maybe we need to regenerate https://github.com/joyent/containerpilot/blob/master/glide.lock and commit it?

justenwalker: Yeah need to run `glide update`

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,228 @@
+package zookeeper
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/joyent/containerpilot/utils"
+	"github.com/samuel/go-zookeeper/zk"
```
tgross: I think we only moved some of the items I mentioned in https://github.com/joyent/containerpilot/pull/214#discussion_r76603448. Anything that isn't stdlib should be in the 2nd group below.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,228 @@
...
+	return false
+}
+
+func (conn ZooKeeper) createParentPath(path string) error {
+	pathElements := strings.Split(path, "/")[1:]
+	sep := "/"
+	newPath := ""
+	for i := range pathElements {
+		newPath = strings.Join([]string{newPath, sep, pathElements[i]}, "")
+		if _, err := conn.Client.Create(newPath, nil, 0, zk.WorldACL(zk.PermAll)); err != nil && err != zk.ErrNodeExists {
```
tgross: Good approach to fix that. Can you break this long line please?

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	value := encodeZooKeeperNodeValue(service)
+	if _, err := conn.Client.Create(
+		key,
+		[]byte(value),
+		zk.FlagEphemeral,
+		zk.WorldACL(zk.PermAll)); err != nil && err != zk.ErrNodeExists {
+		return err
+	}
+	_, _, ch, err := conn.Client.GetW(key)
```
tgross: The edit swallowed the rest of my commentary, so circling back to the design question here...

@tgross wrote:
> I may be misunderstanding the code here -- can you explain the intent of what we're doing in this section? It looks like you're setting a watch on the key and then starting a new goroutine that creates a new watch on each event received. This seems like we're going to be firing up a new goroutine on every health check poll and then deregistering when the TTL expires but I can't figure out how we're passing that goroutine a message not to deregister.

@giorgio-v wrote:
> As for the checks, the watcher is set only if `k` does not exist. Does this clarify your doubts a bit?

I understand that, but what is the purpose of the watcher and how are we communicating with it?

discovery/zookeeper/zookeeper_test.go
```diff
@@ -0,0 +1,492 @@
...
+	zookeeper.Deregister(service)
+	if err := zookeeper.Client.Delete("/containerpilot/my-service", -1); err != nil {
+		t.Fatalf("Unable to cancel parent node: %s", err)
+	}
+	if err := zookeeper.Client.Delete("/containerpilot", -1); err != nil {
+		t.Fatalf("Unable to cancel grand parent node: %s", err)
+	}
+}
+
+func TestDeregisteServiceIdempotency(t *testing.T) {
```
tgross: should be:

> TestDeregisterServiceIdempotency

discovery/zookeeper/zookeeper.go
giorgio-v: Yeah I hate them, but `go fmt` insists on keeping everything on the same line. Does anyone know how to deal with this? I’m using `go fmt -s -w .`

discovery/zookeeper/zookeeper.go
giorgio-v: The watcher is used to intercept changes and hence stop the timeout, but this approach is flawed anyway.

1. Now that we don’t conditionalize on the existence of the node, as you already pointed out, we are creating a channel at every call.
1. Moreover, there shouldn’t be any need to use the channel returned by `GetW` since the Go library lets you register a connection-level callback.

So, back to the drawing board, but the plan should be something like this:
- Define the callback at the `ZooKeeper` struct level. It should be an extraction from the current `select` inside `registerService`, roughly speaking.
- Correct `registerService` so that it sets the value regardless of the outcome of `Create` call and, in doing so, triggering the `EventNodeDataChanged` event every time it is called.
- Correct `TestZKWatchEventNodeDataChanged` to demonstrate:
-- How to use connection-level callback
-- Event firing _even with the same data_ written to the node

giorgio-v:
Don't know if get notified upon new commits. Anyway, I have updated the PR to address the remaining observations, especially regarding TTL and channels.

misterbisson:
@giorgio-v I'd love to talk with you and learn more about your work and plans for this. Would you mind emailing me? I'm casey dot bisson at joyent

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,266 @@
...
+		return nil, err
+	}
+	zookeeper.Client = conn
+	return zookeeper, nil
+}
+
+// Deregister removes this instance from the registry
+func (conn *ZooKeeper) Deregister(service *discovery.ServiceDefinition) {
+	key := conn.getNodeKey(service)
+	exists, _, err := conn.Client.Exists(key)
```
tgross: Why check `Client.Exists` here (and throwing away the channel) and making an extra ZK request instead of just doing `Client.Delete` and then handling the error? You're throwing away the error in `conn.Client.Delete` too which means if it fails the check for `Exists` didn't do you any good anyways.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,266 @@
...
+
+func (conn *ZooKeeper) getAppKey(appName string) string {
+	return fmt.Sprintf("%s/%s", conn.Prefix, appName)
+}
+
+var zookeeperUpstreams = make(map[string][]ServiceNode)
+
+// CheckForUpstreamChanges checks another zookeeper node for changes
+func (conn *ZooKeeper) CheckForUpstreamChanges(backendName, backendTag string) bool {
+	// TODO: is there a way to filter by tag in zookeeper?
```
tgross: It doesn't look like there's such a concept in ZK. We should remove this note and the `Tags` field of the results struct.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,266 @@
...
+func (conn ZooKeeper) createParentPath(path string) error {
+	pathElements := strings.Split(path, "/")[1:]
+	sep := "/"
+	newPath := ""
+	for i := range pathElements {
+		newPath = strings.Join([]string{newPath, sep, pathElements[i]}, "")
+		if _, err := conn.Client.Create(
+			newPath,
+			nil,
+			0,
```
tgross: We've got the `data` and `flags` fields here and they're both unnamed and totally undocumented in the ZK client docs. Can we at least put a trailing comment on these fields to note that these are undocumented ZK client fields?

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	return nil
+}
+
+func (conn ZooKeeper) registerService(service *discovery.ServiceDefinition) error {
+	key := conn.getNodeKey(service)
+	if err := conn.createParentPath(conn.parentPath(service)); err != nil {
+		return err
+	}
+	value := encodeZooKeeperNodeValue(service)
```
tgross: You're not handling `nil` here if there's an encoding error. The idiomatic approach you might want to consider is having `encodeZooKeeperNodeValue` return `([]byte, error)` instead of just `[]byte`. This will make it obvious when a `nil` check is being dropped.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,266 @@
...
+	if _, err := conn.Client.Create(
```
tgross: What's in the string we're throwing away here? Is it something that might be handy in `log.debug`?

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	return didChange
+}
+
+func (conn *ZooKeeper) getServices(appName string) ([]ServiceNode, error) {
+	var services []ServiceNode
+
+	key := conn.getAppKey(appName)
+	children, _, error := conn.Client.Children(key)
+	if error != nil {
```
tgross: `error` is a keyword, this should be `err`. (I'm _really_ surprised `go vet` doesn't call this out.)

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,266 @@
...
+	value := encodeZooKeeperNodeValue(service)
+	if _, err := conn.Client.Create(
+		key,
+		[]byte(value),
+		zk.FlagEphemeral,
+		zk.WorldACL(zk.PermAll)); err != nil && err != zk.ErrNodeExists {
+		return err
+	}
+	// Set the watcher and trigger the call via `Set`
+	_, _, _, err := conn.Client.GetW(key)
```
tgross: I'm still really unclear on what we're trying to do with this. It reads to me like:

- we create a connection with an event callback in `NewZooKeeperConfig`
- we create a watch on a key and throw away the channel here in `registerService`
- we set a value on that key here in `registerService`
- we start a goroutine here in `registerService`
- when an event callback that goroutine checks the service key, creates a new watch and throws away that new channel. If there's an error in doing so it deregisters.

What is sending messages to the event callback in the first place? Why are we creating watches and throwing them away by doing `GetW` in the `ttlHandler` instead of doing `Get`?

tgross: Also, as a style issue this function calls functions that are defined at the top of this page of code and have only this function as a caller. This means that in order to read the code you have to jump back and forth from the top and bottom of the page. If the called functions `ttlHandler` and `eventChanCallback` were to live down here this would be more legible.

discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,266 @@
...
+			nil,
+			0,
+			zk.WorldACL(zk.PermAll)); err != nil && err != zk.ErrNodeExists {
+			return err
+		}
+	}
+	return nil
+}
+
+func (conn ZooKeeper) registerService(service *discovery.ServiceDefinition) error {
```
tgross: Is there any shared state being created here under the hood in the `Zookeeper` struct? Does this need a pointer receiver? (i.e. `func (conn *Zookeeper) registerService...`)

glide.yaml
tgross: @giorgio-v you'll still need to do this to get Travis to pass.

tgross:
Thanks again for taking on this slog of a project @giorgio-v. Just a few more items to work out and then we'll need the integration tests and doc updates. This is looking pretty good!

giorgio-v:
No, thank _you_, Tim! 

Really, really appreciate your suggestions. I'm sorry you have to deal with some rookie mistakes. My bad.

glide.yaml
giorgio-v: Not clear what should I do here. IIRC, `glider update` doesn't change anything to commit on my repo. Or are you referring to the Make file?

discovery/zookeeper/zookeeper.go
giorgio-v: - With a connection-level callback handler, every "interesting" event is received, provided that you register a watch. Other events ending up here are session-related and we ignore them, at least at the moment. 
- Watches must be re-registered every time you receive a notification. That's why we call GetW in ttlHandler. 
- The channel is thrown away because we handle events at the connection level, otherwise it seems to me we go back to the first version, IIRC, where we used that channel with the select. Maybe I misunderstood your previous suggestions?

Do these address your questions?

discovery/zookeeper/zookeeper.go
giorgio-v: It returns the path of the node being created. Yes, I think it's a good idea to log it

discovery/zookeeper/zookeeper.go
giorgio-v: The intent was to avoid excessive noise during deregistration, when we try to delete an already deregistered node.

discovery/zookeeper/zookeeper.go
giorgio-v: Not sure I understand your question (sorry for being dense), but no, we don't create any additional state, I think.

tgross: If we don't have a pointer receiver then we're operating on a value and not a reference. So if this method changes anything in the `Zookeeper` struct it won't be changed in the struct that we store as our [`App.ServiceBackend`](https://github.com/joyent/containerpilot/blob/574570ab5fe608da0709f13d37b15f7707b1fbcc/core/app.go#L34)

discovery/zookeeper/zookeeper.go
tgross: Right but it's racy between the check and the `Delete`. If the error is meaningful anyway (which it must be if we're `log.Warnf` it), it's better to just try to `Delete` and then catch and log that error.

```go
func (conn *ZooKeeper) Deregister(service *discovery.ServiceDefinition) {
	key := conn.getNodeKey(service)
	log.Warnf("Deregistering %s", key)
	if err := onn.Client.Delete(key, -1); err != nil {
		log.Errorf("Error on deregistering %s: %s, key, err)
	}
```

Also, you may want to consider returning an error from this function and then having the caller log it. This would make this a bit more testable.

discovery/zookeeper/zookeeper.go
tgross: > With a connection-level callback handler, every "interesting" event is received, provided that you register a watch.

Note that this implies that we probably should have this method take a pointer receiver too.

> Maybe I misunderstood your previous suggestions?

I didn't really have suggestions, just a lot of questions because I don't understand the workflow at a higher-level. You're explaining to me the implementation details but I'm still stuck on the basics. 😊

Why are we using watches at all here? What are the events we're expecting to receive from ZK on that connection-level channel? It doesn't look like the `onChange` handler is using that to receive messages, which is what I would expect watches to be used for. But it's possible ZK uses the "watch" terminology very differently from Consul and etcd.

discovery/zookeeper/zookeeper.go
giorgio-v: Exactly. On the other hand, we cannot make any guarantee on the purity of  the underlying library connection, so I I guess it would be preferable using a pointer, correct?

discovery/zookeeper/zookeeper.go
giorgio-v: > Note that this implies that we probably should have this method take a pointer receiver too. 

Yes, see above.

Re. the high-level picture, I’ll have to reply once out of the office. 🙂

discovery/zookeeper/zookeeper.go
tgross: That's a good idea.

discovery/zookeeper/zookeeper.go
giorgio-v: Yes, this part of the ZK client library seems complicated. 

I’ll try my best to explain what I’ve understood:
- In ZK, a watcher gets notified _only once_ on relevant changes. If you want to receive another event you have to re-register the watch.
- These events are of different types: `EventNodeCreated`, `EventNodeDeleted`, `EventNodeDataChanged`, `EventNodeChildrenChanged`, `EvenSession` and `EventNotWatching`.
- In our context, we are interested in `EventNodeDataChanged` since we are setting the value every time the heartbeat is sent, ultimately via `registerService`. See also `TestZKWatchEventNodeDataChanged`
- Now, the library lets you handle the events in two ways: either with the channel returned by `GetW` or via the connection-level callback. The crucial (and confusing, I think) point is that even with this callback you have to set the watch anyway. Again, see `TestZKWatchEventNodeDataChanged`.

Does this clarify things a bit?

discovery/zookeeper/zookeeper_test.go
giorgio-v: I’m trying to wrap my head around the Makefile etc. I can create the Docker instance but I’m getting a bit lost. Could you please give me some pointers about other files to tweak? That would be really helpful. 

I could also document it in the README, if you think it could be useful. Let me know.

glide.yaml
giorgio-v: OK, got it at last. I’ll update the PR once I get `make test` pass (see https://github.com/joyent/containerpilot/pull/214/files#r79051645) so that Travis can be happy as well.

giorgio-v:
I have update the PR to address the last round of observations. I’d need some help / pointers to integrate ZK within the rest of the test infrastructure. I have already added ZK to the makefile (this part is not in the latest batch of commits) but I’m unsure about how to proceed.

Any info would be really appreciated. 🙂

vesa-alho-nc:
I would be interested in testing this. Do you have the makefile available somewhere?

giorgio-v:
@vesa-alho-nc Not yet, I’d like to have something minimally usable and currently it is not.

tgross:
For the unit tests you'll need a ZK target backend similar to how we did the Consul and etcd backends in [the makefile](https://github.com/joyent/containerpilot/blob/master/makefile#L97-L114).

For the integration tests you'll need the following:

- A test fixture for ZK. This will be at `./integration_tests/fixture/zk` and look a lot like the one for Consul: https://github.com/joyent/containerpilot/tree/master/integration_tests/fixtures/consul. It's a Dockerfile for a ZK container that we can build as a test target.
- The tests. This will probably be at `./integration_tests/tests/test_discovery_zk` and look a lot like the one for Consul: https://github.com/joyent/containerpilot/tree/master/integration_tests/tests/test_discovery_consul. It's a `docker-compose.yml` file and a run script to step thru the Compose, wait for components to settle, etc.

You'll see that the Consul and Etcd discovery tests are using the [`test_probe`](https://github.com/joyent/containerpilot/tree/master/integration_tests/fixtures/test_probe) application as part of the tests so that we don't have to write complicated testing in bash.

discovery/zookeeper/zookeeper.go
tgross: We haven't removed this bit or the `Tags` field yet.

discovery/zookeeper/zookeeper.go
tgross: This change looks good.

discovery/zookeeper/zookeeper.go
tgross: The fix here looks good.

tgross:
I want to break out the algorithm discussion b/c it's getting lots in the GH UI:

@giorgio-v wrote:

> I’ll try my best to explain what I’ve understood:
- In ZK, a watcher gets notified only once on relevant changes. If you want to receive another event you have to re-register the watch.
- These events are of different types: `EventNodeCreated`, `EventNodeDeleted`, `EventNodeDataChanged`, `EventNodeChildrenChanged`, `EvenSession` and `EventNotWatching`.
In our context, we are interested in `EventNodeDataChanged` since we are setting the value every time the heartbeat is sent, ultimately via `registerService`. See also `TestZKWatchEventNodeDataChanged`
- Now, the library lets you handle the events in two ways: either with the channel returned by `GetW` or via the connection-level callback. The crucial (and confusing, I think) point is that even with this callback you have to set the watch anyway. Again, see `TestZKWatchEventNodeDataChanged`.

Ok, having both a `GetW` vs connection-level callback seems like a weird API on the part of the ZK lib but there's probably a use case I'm not thinking of.

The `TestZKWatchEventNodeDataChanged` helps me understand the sequence of events but this all seems really racy. Don't we potentially drop events between the `Create` and `GetW` calls? If we have to re-register after receiving an event on the watch don't we potentially drop events during that race? Or do we just not care because the only event we expect to receive and handle is the timeout?

giorgio-v:
Thanks Tim for hints and the comments. I’m going to work on the testing / makefile side of things and update the PR, probably during this week. 

Re. your observations on the ”racyness”. Yes, my assumption was that we have to deal only with the timeout, but I’ll try to see if we can find another solution / proposal by watching the parent node.

giorgio-v:
I’m working on the testing stuff. Unit tests execute correctly, but I’m having some problems with integration tests and I see them in `master` _not_ in my branch. 

As you can see in https://gist.github.com/giorgio-v/320b6320d595bf3c5d8e0a16f8ac2964 I see two anomalies:
- The creation of the fixture `cpfix_zookeeper`. I don’t where this comes from, since I’m executing the tests in `master`
- Well, the second one is not an “anomaly”: the tests just fail and I don’t know why.

I’m running the tests on Mac OS and I have included in the gist the output from `docker info`.

tgross:
> I’m having some problems with integration tests and I see them in master not in my branch.

The integration test script may be descending into an empty directory left behind when you switch branches -- git will leave the directory in place. You may want to check this and do a `make clean` before trying to run the integration tests on a different branch.

giorgio-v:
> The integration test script may be descending into an empty directory left behind when you switch branches -- git will leave the directory in place. You may want to check this and do a make clean before trying to run the integration tests on a different branch.

Yep, I don’t mentions of zookeeper anymore, but it still fails the build. I have updated the gist with the new output.

tgross:
It looks like we've got nearly identical Docker setups: [here's a gist](https://gist.github.com/tgross/3ba9566fda2a89c83cd9ba681bf39052) with my equivalent outputs. I'm on 1.12.2-rc1 but that really shouldn't make a difference in these tests -- we've been running them successfully on various versions for a while now.

I see that you have 6 containers on your machine in `docker info`. Any chance one of these has the name `app`? That might be interfering with the `test_sigterm` result, but that's really just guessing.

What we need to do here is debug the tests. You can run individual tests by doing `TEST=test_tasks make integration`. And you can remove the cleanup of the tests by commenting out a couple sections (arguably these should be flags in the test rig):

- the cleanup in test.sh: https://github.com/joyent/containerpilot/blob/master/scripts/test.sh#L166
- the container removal and and log removal in https://github.com/joyent/containerpilot/blob/master/integration_tests/tests/test_tasks/run.sh#L59-L67

This will give you logs and containers we can start looking into what's wrong with them.

giorgio-v:
Thanks Tim,
- Ran `make clean`
- I have commented out the aforementioned lines
- Ran `TEST=test_tasks make integration` with this output (I remove the first part; I don’t think it’s relevant)

```
No active host found
Error getting IP address: Host is not running
2016/10/04 20:44:22 ==== Create Test Fixtures ====
2016/10/04 20:44:23 Skipping Fixture cpfix_app ... already exists
2016/10/04 20:44:23 Skipping Fixture cpfix_consul ... already exists
2016/10/04 20:44:23 Skipping Fixture cpfix_etcd ... already exists
2016/10/04 20:44:23 Skipping Fixture cpfix_nginx ... already exists
2016/10/04 20:44:23 Skipping Fixture cpfix_test_probe ... already exists
2016/10/04 20:44:23 Skipping Fixture cpfix_zombie_maker ... already exists
2016/10/04 20:44:23 ==== Running Integration Tests ====
2016/10/04 20:44:23 TEST: test_tasks
Creating testtasks_consul_1
Creating testtasks_app_1
Expected task1 to have 7 or 8 executions: got 48
Expected task2 to have 2 or 3 executions: got 16
Expected task2 to have 1 or 2 timeouts after 1500ms: got 14
Expected task3 to have 2 or 3 executions: got 16
Expected task3 to have 16 timeouts after 100ms: got 15
No stopped containers
2016/10/04 20:47:44 FAIL: test_tasks
2016/10/04 20:48:51 ==== Tests Failed! ====
make: *** [integration] Error 1
```

```
> cat integration_tests/tests/test_tasks/c1e5957b24d0141ee7d714fea44a6d060049f8a1d14a89136a587f1ac3ece979.task3
1475609044.616961486
1475609046.115614200
1475609047.619418546
1475609049.115091528
1475609050.616504837
1475609052.120485472
1475609053.617633188
1475609055.116824757
1475609056.617132390
1475609058.119762342
1475609059.620281521
1475609061.114398466
1475609062.624535730
1475609064.119334758
1475609065.618982580
1475609067.122617829
```

```
> cat ./integration_tests/tests/failed.log
test_tasks
```

```
> docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
```

Are there any other relevant file that could be helpful in debugging this?

tgross:
@giorgio-v sorry I got tied up in some other work over the last week. Could you push whatever commits you have so far so that I could try and run those tests on my own? I'd like to see if its some unexpected environmental problem.

giorgio-v:
@tgross Thanks, really appreciate the help. Push coming.

tgross:
@giorgio-v I've checked out your branch and run the tests. The tasks test is working consistently on my machine with your branch, which worries me because that suggests we've got some kind of hard-to-reproduce-and-possibly-environmental problem in that code path.

The ZK test is definitely failing though. I've added these lines to your test to dump the logs before we remove the containers:

```
docker logs testdiscoveryzookeeper_zookeeper_1 > zk1.log 2>&1
docker logs testdiscoveryzookeeper_app_1 > app1.log 2>&1
docker logs testdiscoveryzookeeper_nginx_1 > nginx1.log 2>&1
docker logs testdiscoveryzookeeper_test_1 > test1.log 2>&1
```

There's a configuration failure in app1.log:

```
2016/10/17 13:20:52 Could not read config file: open /app-with-zookeeper.json: no such file or directory
```

And a configuration failure in nginx1.log:

```
2016/10/17 13:20:52 Could not read config file: open /etc/nginx-with-zookeeper.json: no such file or directory
```

Could you try dumping the logs of all the containers in the tasks test as I've done here?

giorgio-v:
@tgross Do you want me to gather the logs from the master branch, i.e. something like this?
```
docker logs testdiscoveryetcd_zookeeper_1 > etcd.log 2>&1
docker logs testdiscoveryetcd_app_1 > app1.log 2>&1
docker logs testdiscoveryetcd_nginx_1 > nginx1.log 2>&1
docker logs testdiscoveryetcd_test_1 > test1.log 2>&1
```

tgross:
> Do you want me to gather the logs from the master branch, i.e. something like this?

Yes, for whatever containers/branches where you've still got test failures.

giorgio-v:
Here they are:
```
./app1.log
Error: No such container: testdiscoveryconsul_app_1
./consul1.log
Error: No such container: testdiscoveryconsul_consul_1
./etcd.log
Error: No such container: testdiscoveryetcd_etcd_1
./integration_tests/tests/failed.log
test_discovery_zookeeper
test_sigterm
test_tasks
test_telemetry
./integration_tests/tests/test_version_flag/app1.log
Error: No such container: testdiscoveryzookeeper_app_1
./integration_tests/tests/test_version_flag/app1_consul.log
Error: No such container: testdiscoveryconsul_app_1
./integration_tests/tests/test_version_flag/app1_etcd.log
Error: No such container: testdiscoveryetcd_app_1
./integration_tests/tests/test_version_flag/consul1.log
Error: No such container: testdiscoveryconsul_consul_1
./integration_tests/tests/test_version_flag/etcd1.log
Error: No such container: testdiscoveryetcd_etcd_1
./integration_tests/tests/test_version_flag/nginx1.log
Error: No such container: testdiscoveryzookeeper_nginx_1
./integration_tests/tests/test_version_flag/nginx1_consul.log
Error: No such container: testdiscoveryconsul_nginx_1
./integration_tests/tests/test_version_flag/nginx1_etcd.log
Error: No such container: testdiscoveryetcd_nginx_1
./integration_tests/tests/test_version_flag/test1.log
Error: No such container: testdiscoveryzookeeper_test_1
./integration_tests/tests/test_version_flag/test1_consul.log
Error: No such container: testdiscoveryconsul_test_1
./integration_tests/tests/test_version_flag/test1_etcd.log
Error: No such container: testdiscoveryetcd_test_1
./integration_tests/tests/test_version_flag/zk1.log
Error: No such container: testdiscoveryzookeeper_zookeeper_1
./nginx1.log
Error: No such container: testdiscoveryconsul_nginx_1
./test1.log
Error: No such container: testdiscoveryconsul_test_1
./zk1.log
Error: No such container: testdiscoveryzookeeper_zookeeper_1
```

tgross:
Ok, you didn't remove the cleanup of containers as I described in https://github.com/joyent/containerpilot/pull/214#issuecomment-251377323. Right now the test containers are all getting cleaned up by the test rig, so when you grab those logs the containers are already gone.

giorgio-v:
Mmh, no :) 

I have commented line 166 on `test.sh` and after that I’ve added the `docker logs` statements. I’ve commented line 66 on `run.sh`. 

I’m confused. :/

giorgio-v:
Folks, sorry to bother you, but I’d need your help / suggestions to fix the problem with the integration tests on my machine in `master` and to complete writing the integration tests in this PR, otherwise I’m stuck. :(

Thanks for any help.

tgross:
Sorry it's taken me a while to get back to this @giorgio-v, I've been doing some traveling and been tied up in some other work.

---

For the failing tests you're getting on `master`:

 It's a little unclear to me why you're not able to reproduce the logs. In https://github.com/joyent/containerpilot/pull/214#issuecomment-255017619 we can see that the ZK tests are in the list of tests, which suggests that when you're checking out the `master` branch that you're not removing the now-empty and untracked `test_discovery_zookeeper/` directory (git won't remove the empty dir, but our test script will try to run stuff there anyways). I've pushed up a commit onto a testing branch for you ([tgross/gh214_testing_problems](https://github.com/tgross/containerpilot/tree/gh214_testing_problems)). I've tested the following with it and got useful logs out of it:

```
rm -rf ./integration_tests/tests/test_discovery_zookeeper
rm -rf integration_tests/fixtures/zookeeper
make clean
TEST=test_tasks make integration
docker logs testtasks_app_1 > app1.log 2>&1
```

Copy _all_ the output that you get into a gist, including all the clean and build info (maybe there's something we're missing in there). And then copy `./app1.log` and `./integration_tests/tests/test_tasks/<some long id>.task{1..3}` into a gist.

---

For this ZK support PR:

The last build https://travis-ci.org/joyent/containerpilot/builds/168061433 is still showing no ZK test running but there's a ZK fixture creates and the test run aborts and fails. If you have a test completed, it would be helpful to get it committed so that Travis runs it and we can debug further from there.

Also, seeing as how this PR has been open for a while and we've landed some bugfixes, it might be a good idea to rebase onto master for this PR to make sure you've got all the latest working stuff.

giorgio-v:
Thanks @tgross for getting back to me. 

Regarding the ZK tests, I didn’t write them because it wasn’t clear to me how to write them. I have tried to simply replicate / copypaste the ones used with Consul, but I couldn’t wrap my head around them. It’s a bit opaque. I’ll try to take a look at it again.

giorgio-v:
Hey @tgross here’s [the output](https://gist.github.com/giorgio-v/feb7e7e0bba68cfc91d88f44b3001da1). I see no errors. Beats me why.

I’ll rebase my branch and work on the tests again.

tgross:
> And then copy ./app1.log and ./integration_tests/tests/test_tasks/<some long id>.task{1..3} into a gist.

Ok, but we need this bit too.

giorgio-v:
Gist updated.

tgross:
We're going to close this PR, as ContainerPilot v3 is consolidating on Consul per the [RFD](https://github.com/joyent/rfd/pull/13). I really appreciate the effort spent on this though.

giorgio-v:
👍 it's the right decision. Hope to have something else to contribute in the future. 

--
Giorgio Valoti

Issue #178, opened 2016-06-15
wait: no child processes


kenjones-cisco:
I keep see the `wait: no child processes` show up in my log files but I've not been able to determine the root cause of why that happens. 

Does this mean that a health check or onChange script execution couldn't be executed as there were no available child processes?

The only reference I saw previously dealt with the reapChildren work done previously, so does this mean that the attempt to reap found no children to reap?

tgross:
I've run into this intermittently. The code section in question is in utils/run.go [`ExecuteAndWait`](https://github.com/joyent/containerpilot/blob/37b946527b49d95997536cbdc4035cd7cb2735b8/utils/run.go#L42-L54). If you check out the golang source code for [`cmd.Run`](https://golang.org/src/os/exec/exec.go?s=7135:7160#L248) you'll see a race condition. The process is started and then we wait for it. But if the process completes and exits before the wait happens (because, say, the goruntime decides to do a GC pause right then or the goroutine yields for the syscall), then we'll get an error there.

You should be able to add some logging to your health/onChange handlers to verify this for yourself; you'll see your logs and then the `wait: no child processes` error. That being said, I'd like to come up with a better way to handle this. We want to make sure we're logging if there's a real error, but we don't want to log in this case which is really just a limitation of the underlying API.

tgross:
I'm going to tag this as a bug although it's really more of a UX thing.

tgross:
@kenjones-cisco are you running consul-template in the container where you're seeing this? And if so, which version? I've been having trouble reproducing lately and I wonder if you're seeing https://github.com/hashicorp/consul-template/issues/506 which was fixed in consul-template 0.13 which was released in Feb.

kenjones-cisco:
consul-template v0.15

tgross:
Ok, I have definitely been able to reproduce now... just need to figure out how to fix it next. 😀

tgross:
I believe that I've fixed this in #184 and #199. We're no longer calling `cmd.Run` and instead have called `cmd.Start` and `cmd.Wait` separately for all `health`, `onChange`, and `sensor` handlers.

We _may_ want to try to catch the error `wait: no child processes` [here](https://github.com/joyent/containerpilot/blob/5ba69b4e042a497a30912445161a553c4667c8a2/commands/commands.go#L189) and just `log.Debugf` it, but let me see if I can verify that first.

tgross:
[...]

PR #222, opened 2016-09-23
Prevent 'no child process' err from breaking heartbeats

tgross:
For https://github.com/joyent/containerpilot/issues/178

The golang stdlib (and our RunAndWait implementation) has a race where if we call `Wait` after the child process has already exited and been cleaned up then an error bubbles up through the stack. We don't need these to break heartbeat TTLs.

In https://github.com/joyent/containerpilot/issues/178#issuecomment-237945112 I mentioned I thought I'd fixed this but I got it backwards -- the `RunAndWait` implementation just repeats the same golang stdlib quirk!

cc @misterbisson @justenwalker @kenjones-cisco

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -12,6 +12,8 @@ import (
 	"github.com/joyent/containerpilot/utils"
 )
 
+const errNoChild = "wait: no child processes"
+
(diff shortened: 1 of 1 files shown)


justenwalker:
LGTM

Issue #223, opened 2016-09-29
Health check not honoring exit code


nickwales:
Creation of health check and service is working fine, but health check status doesn't update when my health endpoint returns non 200 status codes. The health endpoint is powered with a simple webrick server, the response looks similar to this.

```
  response.status = 500
  response.content_type = 'text/plain'
  response.body = "Something is wrong"
```

With CP DEBUG turned on I can see the health checks happening and then the associated webrick access log event, so I can tell it is configured OK.

A response of 500 from the health endpoint gives the curl command `curl --fail -s http://localhost:8008/alive.txt` (I'm running with `--net=host`) an exit code of 22 but CP continues to send passing health checks to consul.

I've been using CP version 2.4.1 and 2.4.2 since yesterday.

tgross:
Just for clarification, is this a new issue with 2.4.2 or were you seeing this with 2.4.1 as well?

nickwales:
I'm seeing this with both versions. 


This is the services portion of the config file I'm using, might be useful!
```
 "services": [
    {
      "name": "service-name",
      "port": 8080,
      "health": "/usr/bin/curl --fail -s --show-error --output /dev/null http://0.0.0.0:8080/alive.txt",
      "interfaces": [
        "eth0",
        "eth1[1]",
        "192.168.0.0/16",
        "2001:db8::/64",
        "eth2:inet",
        "eth2:inet6",
        "inet",
        "inet6"
      ],
      "poll": 10,
      "ttl": 30,
      "timeout": "10s",
      "tags": ["tag1"]
    }
  ]
```

tgross:
Ok I was able to confirm this behavior with another container image running 2.4.2 and 2.4.1. Sorry about that -- that's obviously a serious bug and I'm not sure how it made past the test suite but we'll make it right.

I'm in the process of trying to determine what version the problem appeared in so I can fix it ASAP.

nickwales:
Thanks for the quick response! Looking forward to 2.4.3

tgross:
Adding the following failing test to `commands_test.go` demonstrates out the behavior:

```go
func TestRunWithTimeoutFailed(t *testing.T) {
	log.SetLevel(log.DebugLevel) // DEBUG
	cmd, _ := NewCommand("./testdata/test.sh failStuff --debug", "0")
	fields := log.Fields{"process": "test"}
	if err := RunWithTimeout(cmd, fields); err == nil {
		t.Errorf("Expected error but got nil")
	}
}
[...]

PR #224, opened 2016-09-29
Fix exit codes not being respected by RunWithTimeout

tgross:
For https://github.com/joyent/containerpilot/issues/223

We're assuming that `Process.Wait()` returns an error if the process exits with an error code. `Cmd.Wait()` returns with an error value on a non-zero exit code but `Process.Wait()` returns an error on OS-specific details (like if the process isn't a child process) and instead we should be checking the value of the `ProcessState.Success()` method.

It looks like this got introduced to most of the pollables during the refactoring to support timeouts for health checks (#184); the logic was lifted from the original tasks implementation but was totally safe there because we didn't really care about the exit codes. I've added two tests that ensure we don't miss this again.

cc @misterbisson @nickwales @justenwalker

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


tgross:
[Curse you golint!](https://travis-ci.org/joyent/containerpilot/builds/163847195) Fixed that and squashed the fix. Should be good-to-go now.

Issue #225, opened 2016-09-30
Discovery backends should be healthy on 1st heartbeat


tgross:
Both the `discovery/consul` and `discovery/etcd` backends register the service upon first receiving the `SendHeartbeat`. However, they then exit and don't send that heartbeat until the next health check. This works just fine but it increases the amount of time it takes for new instances of a service to come online unnecessarily.

The relevant section of `discovery/consul`:

``` go
// SendHeartbeat writes a TTL check status=ok to the consul store.
// If consul has never seen this service, we register the service and
// its TTL check.
func (c *Consul) SendHeartbeat(service *discovery.ServiceDefinition) {
    if err := c.Agent().PassTTL(service.ID, "ok"); err != nil {
        log.Infof("%v\nService not registered, registering...", err)
        if err = c.registerService(*service); err != nil {
            log.Warnf("Service registration failed: %s", err)
        }
        if err = c.registerCheck(*service); err != nil {
            log.Warnf("Check registration failed: %s", err)
        }
    }
}
```

The relevant section of `discovery/etcd`:

``` go
// SendHeartbeat refreshes the TTL of this associated etcd node
func (c *Etcd) SendHeartbeat(service *discovery.ServiceDefinition) {
    if err := c.updateServiceTTL(service); err != nil {
        log.Infof("Service not registered, registering...")
        if err := c.registerService(service); err != nil {
            log.Warnf("Error registering service %s: %s", service.Name, err)
        }
    }
}
```

We should rework these slightly so that after the error is received we register the service and then immediately register the check (in the case of Consul) and update the service TTL (in both).

(We should probably also log errors on failed registration rather than warns, for that matter.)

tgross:
Marking this as a 2.x-compatible enhancement.

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/2.5.0

PR #247, opened 2016-11-18
Retry heartbeat after initial service registration

tgross:
For #225 

Sending a heartbeat to a not-yet registered service fails on both the `discovery/consul` and `discovery/etcd` backends. Previously we'd catch the error, register the service, and return so that we'd send a passing heartbeat on the next health check poll. This change retries the heartbeat on the newly registered service so that we become healthy more quickly.

cc @misterbisson @jasonpincin @justenwalker


tgross:

PR #248, opened 2016-11-18
upgrade to consul ~0.7.0

flyinprogrammer:
I'm running the tests locally right now, but let's see if this works in automation too?

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 21a6a3480ab824375461d97a619484adae555789815f71308e65b42416258f17
-updated: 2016-07-22T16:00:15.451451414Z
+hash: 874e4f3d7d988b377f35781a0bc6b063efe4c70422072bc4a0ab8861fc2d4e7b
+updated: 2016-11-18T13:01:24.244820685-06:00
 imports:
 - name: github.com/beorn7/perks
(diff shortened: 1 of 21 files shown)


tgross:
LGTM. Can we rebase this on master so that we can run thru the integration tests again with the updated Consul container?

flyinprogrammer:
yup - brb!

flyinprogrammer:
interesting running this locally results in all kinds of intermittent failures, I'll have to dig into what these integration tests do/are and see if a pattern and solution emerges 😀

flyinprogrammer:
I suspect this will pass; however, my `wait4consul` pattern is probably unacceptable. I can refactor this functionality into `test_probe`, and the already established pattern for accomplishing this goal.

tgross:
> I suspect this will pass; however, my wait4consul pattern is probably unacceptable. I can refactor this functionality into test_probe, and use the established pattern for accomplishing this goal.

Is the problem just that we're using the `-dev` server now instead of doing a `bootstrap=true` / `bootstrap-expect=1`?

flyinprogrammer:
no the problem is consul is just slower to boot up now, and we never had explicit waits in place.

tgross:
Ok, LGTM.

tgross:

Issue #229, opened 2016-10-12
CONTAINERPILOT_* environment variables for hook scripts


jasonpincin:
Proposing that ContainerPilot set certain environment variables (prefixed by CONTAINERPILOT_) that are accessible by all hook scripts, tasks, co-processes, and the "main" application process.. 
### List of environment variables
- `CONTAINERPILOT_PID` - The PID of ContainerPilot itself, useful for signaling it to reload it's own config
- `CONTAINERPILOT_APP_PID` - The PID of the primary application 

That's all I have for now. There may be other valuable pieces of information now or later that can follow the same naming scheme. 

See #232 for a related proposal.

tgross:
> The PID of ContainerPilot itself

Isn't that always 1?

misterbisson:
> > The PID of ContainerPilot itself

> Isn't that always 1?

What if ContainerPilot is being run via another supervisor/init system, as we've suggested for use in VMs or infrastructure containers?

tgross:
Ah, fair enough I'd forgotten that use case. Totally trivial to get with `os.Getpid()` too.

jasonpincin:
Updated description to mention #232, and asset that the env vars would be exported for tasks, co-processes, and the main application (instead of just hook scripts as I originally implied).

tgross:
Marking this as a 2.x-compatible enhancement.

tgross:
So looking into this a bit, getting the `CONTAINERPILOT_PID` into the environment will be trivial. We can do this right before we load the config via `os.Setenv("CONTAINERPILOT_PID", os.Getpid())`

Getting the `CONTAINERPILOT_APP_PID` is trickier though. Currently for the main app in `RunAndWait` we do a `exec.Cmd.Run()` which blocks until exit, so we don't have a way of getting at the process PID. We can change this to `exec.Cmd.Start()`, get the PID, and then do `exec.Cmd.Process.Wait()`. We do this in the pollables now, but its racy in terms of getting errors on the PID and Wait if the command exits quickly (i.e. it crashes). This should be less of a problem with the main app.

We also use this section for the other hooks that are running exclusively: `preStart`, `preStop`, and `postStop`. We could reduce special casing and also expose a `CONTAINERPILOT_PRESTART_PID`, `CONTAINERPILOT_PRESTOP_PID` and `CONTAINERPILOT_POSTSTOP_PID` variable if we construct the env var name from the name of whatever `Command` struct we're running.

[...]

PR #250, opened 2016-11-18
Expose PIDs of ContainerPilot and hooks as environment variables

tgross:
For https://github.com/joyent/containerpilot/issues/229

This adds the following environment variables by grabbing the PID right after we start a process but before we block waiting for it to finish. This PID is set as an env var available to all _subsequently started_ processes.

- `CONTAINERPILOT_PID`: the PID of ContainerPilot itself. This is available to all hooks and to the main application.
- `CONTAINERPILOT_APP_PID`: the PID of the main shimmed application. This is available to all hooks except for the `preStart`. It is not available to the shimmed application itself (we need to start the application first to get its PID).
- `CONTAINERPILOT_PRESTART_PID`: the PID of the `preStart` hook while it runs.
- `CONTAINERPILOT_PRESTOP_PID`: the PID of the `preStop` hook while it runs.
- `CONTAINERPILOT_POSTSTOP_PID`: the PID of the `postStop` hook while it runs.
- `CONTAINERPILOT_{COPROCESS_NAME}_PID`: the PID of a `coprocess` hook while it runs.

cc @jasonpincin @misterbisson

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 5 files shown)


jasonpincin:
Only one thought: would there be any advantage to setting the new ENV vars via the `Command` struct's `Env` rather than changing the environment of ContainerPilot itself? 

That question aside, 👍

tgross:
I can't think of any advantage and it means we'd have multiple places for that data to live in memory, which will make it slightly harder to update any information we want on a config reload.

tgross:

Issue #243, opened 2016-11-17
Support Consul EnableTagOverride


flyinprogrammer:
If I'm reading this correctly:
https://github.com/joyent/containerpilot/blob/master/glide.yaml#L24
It means we're using consul from this commit:
https://github.com/hashicorp/consul/tree/158eabdd6f2408067c1d7656fa10e49434f96480

Which means we're using consul bindings from 0.5.2
There have been a lot of updates since then (we're on 0.7.1 now), like `EnableTagOverride`
https://github.com/hashicorp/consul/blob/73b281a27c748210f64e4e83b23312d23afb4593/api/agent.go#L51

And I was hoping to add an option to use `EnableTagOverride` here:
https://github.com/joyent/containerpilot/blob/master/discovery/consul/consul.go#L127

Which would mean probably adding a field to the `ServiceDefinition`?
https://github.com/joyent/containerpilot/blob/master/discovery/discovery.go#L22
But I'm hesitant to do that because things like `etcd` have no concept of an `EnableTagOverride` flag. So I just don't know how we want to accomplish this.

I'm sorry I'm so new to this code base. I demand :stuck_out_tongue: that I write the code and submit the patch, if this is a feature we want? (I need it.), so I would appreciate any and all feedback so I can start contributing more silly ideas.

tgross:
There's no specific reason we've fallen behind on that, so updating the bindings sounds like a good idea. What would `EnableTagOverride` look like in the case where we have non-Consul backends?

Just for the record, because I had to look up what that does (from https://www.consul.io/docs/agent/services.html):

> The enableTagOverride can optionally be specified to disable the anti-entropy feature for this service. If enableTagOverride is set to TRUE then external agents can update this service in the catalog and modify the tags. Subsequent local sync operations by this agent will ignore the updated tags.

tgross:
@flyinprogrammer I'd like to suggest that you make two PRs if you decide to take this on: one to upgrade Consul and a second one to tackle the question of `EnableTagOverride`

tgross:
For `EnableTagOverride` I'm going to suggest you go with the proposal in https://github.com/joyent/containerpilot/issues/213 in terms of config syntax. i.e. namespace the configuration change.

tgross:
Being implemented in https://github.com/joyent/containerpilot/pull/252

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/2.6.0

Issue #218, opened 2016-09-09
Need node deregistration from Consul cluster when a container exited


joongyu:
I'm using containerpilot with consul as backend discovery. The consul agent is running as a coprocess for containerpilot. 

I see that it successfully register its node and service to the consul when it started. But, I found out that containerpilot does not deregisters the node itself from consul cluster when it's terminated while it deregisters its service.  (I've checked this on latest release version of containerpilot 2.4.1)

Is there a reason containerpilot does not deregister its node on consul? 
If not, it seems to leave itself when exited.

tgross:
ContainerPilot doesn't contain any Consul-agent specific behavior for co-processes, so when ContainerPilot receives SIGTERM it leaves the co-processes running until the entire container exits. 

You can use the `postStop` hooks to cause the agent to [leave the cluster gracefully](https://www.consul.io/docs/commands/leave.html). Your configuration might something look like this:

```json
{
  "consul": "localhost:8500",
  "coprocesses": [
    {
      "command": ["/usr/local/bin/consul", "agent",
                  "-data-dir=/var/lib/consul",
                  "-config-dir=/etc/consul",
                  "-rejoin",
                  "-retry-join", "{{ .CONSUL }}",
                  "-retry-max", "10",
                  "-retry-interval", "10s"],
      "restarts": "unlimited"
    }
  ],
  "postStop": "consul leave"
}
```

Note that the `postStop` will not fire if the underlying compute host crashes, so you can still end up with stale Consul agents in that scenario.

fitz123:
@joongyu You can use consul option "leave_on_terminate" for this.

https://www.consul.io/docs/agent/options.html#leave_on_terminate

joongyu:
@tgross I'm using postStop in order to leave from consul. I thought conatinerpilot does something regarding node registration. Now I see it doesn't :) Thanks.
@fitz123 Good to know that such option exists!. I should try this.

misterbisson:
Thanks for the tip @fitz123. I opened https://github.com/autopilotpattern/nginx/issues/27 to implement `leave_on_terminate` in that repo and others.

fitz123:
@misterbisson, @joongyu additional info that could be helpful is [DeregisterCriticalServiceAfter](https://www.consul.io/docs/agent/http/agent.html#agent_check_register) option, implemented in Consul v0.7. It'll help with dead services which I'm having a lot in Docker Swarm cluster environment with unstable networking.
[...]

Issue #213, opened 2016-08-24
Discussion: Discovery service-specific configuration


justenwalker:
I see some times a lot of consul-specific configuration options come up, and it is met with some trepidation, since ContainerPilot's configs are meant to be agnostic. But there appear to be cases where we cannot ignore the differences between those backends, and users do want to take advantage of those specific features.

I'd like to open this issue to get some ideas thrown around about how we can cope with the dissimilarities between the backends. The goal is to allow us to safely create specific features that may only work for a single backend and design configuration in a way that targets it directly and unambiguously.

Here is one such proposal: **namespaced extensions**

Given our current service configuration object such as the one from the documentation.

``` json
    {
      "name": "app",
      "port": 80,
      "health": [
        "/usr/bin/curl",
        "--fail",
        "-s",
        "http://localhost/app"
        ],
      "interfaces": [
        "eth0",
        "eth1[1]",
        "192.168.0.0/16",
        "2001:db8::/64",
        "eth2:inet",
        "eth2:inet6",
        "inet",
        "inet6"
      ],
      "poll": 10,
      "ttl": 30,
      "timeout": "10s",
      "tags": ["tag1"]
    }
```

If we wanted to take advantage of consul's unique service options; perhaps we could extend this to include a `CheckID` by including a `consul` namespace for those options (see #193)

``` json
{
  "name": "app",
  "port": 80,
  "health": [
    "/usr/bin/curl",
    "--fail",
    "-s",
    "http://localhost/app"
  ],
  "interfaces": [
    "eth0",
    "eth1[1]",
    "192.168.0.0/16",
    "2001:db8::/64",
    "eth2:inet",
    "eth2:inet6",
    "inet",
    "inet6"
  ],
  "poll": 10,
  "ttl": 30,
  "timeout": "10s",
  "tags": ["tag1"],
  "consul": {
    "CheckID": "myservice-check"
  }
}
```

The goal is still to put as much as we can into the root namespace, but also to not be so constrained by the fact that other backends are not entirely compatible.

tgross:
I'm open to the idea but I also wonder whether we have enough use cases to justify the added complexity. Like in the case of #193 having a field that the etcd backend ignores seems easy enough.

That being said, I definitely like the syntax here if we're going to do this; it certainly beats "vendor prefix" type config like having `consul-checkId` or something like that.

justenwalker:
> I also wonder whether we have enough use cases to justify the added complexity.

[...]

Issue #164, opened 2016-05-25
Consul availability timeout


mterron:
I'd like to be able to configure a period of time to wait for Consul to be available before killing the container on startup (configurable).
At the moment I'm running this script as CMD in all my CP containers: [https://github.com/mterron/logstash-autopilot/blob/master/bin/prestart.sh](https://github.com/mterron/logstash-autopilot/blob/18a55c1a7405efb89248bbfa26e913d54367cb98/bin/prestart.sh)

This is useful when Consul is part of the composition you will run.

tgross:
That could go in a `preStart` script in most circumstances, but I agree that having some way to wait for the discovery service to be available before starting up would be a good idea.

tgross:
@mterron I've updated your link to a specific commit so that it doesn't 404 on us in the future.

tgross:
Marking this as a 2.x-compatible enhancement.

tgross:
This will be available in v3 as part of the new configuration model. Each service (which now includes what in v2 was preStart, preStop, postStop, coprocess, and task) will be able to set a start timeout. 

Ground work for this has been released in #290 and I'll add the config hooks for it during the v3 config update. Refer to [RFD86](https://github.com/joyent/rfd/blob/master/rfd/0086/README.md) and the v3 roadmap in #283 for details.

tgross:
Implemented in https://github.com/joyent/containerpilot/pull/312

tgross:
Pre-released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-dev.2

PR #252, opened 2016-11-21
setup ConsulConfig and ConsulExtras

flyinprogrammer:
This might address:
https://github.com/joyent/containerpilot/issues/213
https://github.com/joyent/containerpilot/issues/218
https://github.com/joyent/containerpilot/issues/243

I should write more tests.

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -125,18 +125,27 @@ func (c *Consul) SendHeartbeat(service *discovery.ServiceDefinition) {
 }
 
 func (c *Consul) registerService(service discovery.ServiceDefinition) error {
+	var enableTagOverride bool
+	if service.ConsulExtras != nil {
+		enableTagOverride = service.ConsulExtras.EnableTagOverride
+	}
 	return c.Agent().ServiceRegister(
 		&consul.AgentServiceRegistration{
-			ID:      service.ID,
-			Name:    service.Name,
-			Tags:    service.Tags,
-			Port:    service.Port,
-			Address: service.IPAddress,
+			ID:                service.ID,
+			Name:              service.Name,
+			Tags:              service.Tags,
+			Port:              service.Port,
+			Address:           service.IPAddress,
+			EnableTagOverride: enableTagOverride,
 		},
 	)
 }
 
 func (c *Consul) registerCheck(service discovery.ServiceDefinition) error {
+	var deregisterCriticalServiceAfter string
+	if service.ConsulExtras != nil {
+		deregisterCriticalServiceAfter = service.ConsulExtras.DeregisterCriticalServiceAfter
+	}
 	return c.Agent().CheckRegister(
 		&consul.AgentCheckRegistration{
 			ID:        service.ID,
@@ -145,6 +154,7 @@ func (c *Consul) registerCheck(service discovery.ServiceDefinition) error {
 			ServiceID: service.ID,
 			AgentServiceCheck: consul.AgentServiceCheck{
 				TTL: fmt.Sprintf("%ds", service.TTL),
+				DeregisterCriticalServiceAfter: deregisterCriticalServiceAfter,
 			},
 		},
 	)
--- a/discovery/consul/consul_test.go
+++ b/discovery/consul/consul_test.go
@@ -19,6 +19,23 @@ func setupConsul(serviceName string) (*Consul, *discovery.ServiceDefinition) {
 	return consul, service
 }
 
+func setupWaitForLeader(consul *Consul) error {
+	maxRetry := 30
+	retry := 0
+	var err error
+
+	// we need to wait for Consul to start and self-elect
+	for ; retry < maxRetry; retry++ {
+		if retry > 0 {
+			time.Sleep(1 * time.Second)
+		}
+		if leader, err := consul.Status().Leader(); err == nil && leader != "" {
+			break
+		}
+	}
+	return err
+}
+
 func TestConsulObjectParse(t *testing.T) {
 	rawCfg := map[string]interface{}{
 		"address": "consul:8501",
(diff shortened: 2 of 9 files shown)


flyinprogrammer:
So far as I can tell, there's actually no way to get the value of `DeregisterCriticalServiceAfter` out of the consul API because the `/v1/agent/checks` API returns an [AgentCheck](https://github.com/hashicorp/consul/blob/master/api/agent.go#L8) which does not posses the [CheckType](https://github.com/hashicorp/consul/blob/master/command/agent/check.go#L44) data.

services/services.go
```diff
@@ -101,13 +109,22 @@ func parseService(s *Service, disc discovery.ServiceBackend) error {
 	}
 	s.IPAddress = ipAddress
 
+	var consulExtras *discovery.ConsulExtras
+	if s.ConsulConfig != nil {
+		consulExtras = &discovery.ConsulExtras{
+			DeregisterCriticalServiceAfter: s.ConsulConfig.DeregisterCriticalServiceAfter,
```
tgross: Do we ever validate this is a valid time?

services/services.go
```diff
@@ -23,11 +23,18 @@ type Service struct {
...
+	ConsulConfig     *ConsulConfig `mapstructure:"consul"`
 	healthCheckCmd   *commands.Command
 	discoveryService discovery.ServiceBackend
 	definition       *discovery.ServiceDefinition
 }
 
+// ConsulConfig handles additional Consul configuration.
+type ConsulConfig struct {
+	EnableTagOverride              bool   `mapstructure:"enabletagoverride"`
+	DeregisterCriticalServiceAfter string `mapstructure:"deregistercriticalserviceafter"`
```
tgross: I feel like these should probably be camelCase, given that we use it elsewhere (ex. `preStart`)

tgross:
> So far as I can tell, there's actually no way to get the value of DeregisterCriticalServiceAfter out of the consul API because the /v1/agent/checks API returns an AgentCheck which does not posses the CheckType data.

Don't we only ever write it though? Not read it? Or are you concerned (reasonably) about testing it?

services/services_test.go
```diff
@@ -101,6 +101,58 @@ func TestServicesConfigError(t *testing.T) {
...
+  "name": "serviceA",
+  "port": 8080,
+  "interfaces": "eth0",
+  "health": ["/bin/to/healthcheck/for/service/A.sh", "A1", "A2"],
+  "poll": 30,
+  "ttl": 19,
+  "timeout": "1ms",
+  "tags": ["tag1","tag2"],
+  "consul": {
+	  "enabletagoverride": true
```
tgross: We definitely should be testing for misconfigurations for both of these features too.

tgross:
Looking pretty good with just the few questions and comments. It's just about as ugly as I expected in terms of what configuration we end up with. 😀  We should update the [documentation](https://github.com/joyent/containerpilot/tree/master/documentation) as well as part of the same PR.

flyinprogrammer:
Alright see what you think of this @tgross and I'll start working on documentation tonight.

discovery/consul/consul.go
```diff
@@ -47,6 +48,25 @@ func NewConsulConfig(config interface{}) (*Consul, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	maxRetry := 30
+	retry := 0
+	// we need to wait for Consul to start and self-elect
```
tgross: This section appears to have been added to support https://github.com/joyent/containerpilot/issues/164 but I'd really like that behavior change to be a separate PR and design discussion.

I'm not comfortable adding a hard-coded retry here nor making it non-optional. It's entirely reasonable to have an application start up without Consul being ready, and to have it simply fail to make health checks until Consul is available (so long as ContainerPilot can handle this without crashing, which as far as I'm aware it can currently do just fine).

discovery/consul/consul_test.go
```diff
@@ -88,3 +88,35 @@ func TestConsulCheckForChanges(t *testing.T) {
...
+		ConsulExtras: &discovery.ConsulExtras{
+			EnableTagOverride: true,
+		},
+	}
+	id := service.ID
+	if consul.CheckForUpstreamChanges(backend, "") {
+		t.Fatalf("First read of %s should show `false` for change", id)
+	}
+	consul.SendHeartbeat(service) // force registration
+	consul.SendHeartbeat(service) // write TTL
```
tgross: You should be able to do without the second heartbeat now that https://github.com/joyent/containerpilot/pull/247 has been merged.

tgross:
Looks good. There are some integration tests failing intermittently -- I kicked the build again to see if it was a flaky test and a different one failed. I suspect that we're seeing the bug I hopefully fixed in https://github.com/joyent/containerpilot/pull/256 yesterday here, so once that's merged we should rebase on master to double-check.

discovery/consul/consul.go
flyinprogrammer: Hahaha you give me too much credit @tgross, I'm much lazier than this 😄. I went to go add a `WaitForLeader()` type method to our tests, and figured why not add it to the code that actually creates the client. I'm in full agreement with you, this will be gone in a moment, and I'd love to work on #164 in a separate PR as I have a similar desire/use case on wanting an app on startup to crash, but if it's already up, to stay running.

discovery/consul/consul_test.go
flyinprogrammer: Awesome, I'll try this out in a few minutes.

flyinprogrammer:
>  so once that's merged we should rebase on master to double-check.
I await the merge.

This is fun, thanks for your help, and I'll refactor this again based on your comments!

tgross:
Ok #256 has been merged, let's give that rebase a go.

tgross:
This all LGTM. Let's merge it and I'll release this as 2.6.0. Thanks again for your help @flyinprogrammer!

PR #259, opened 2016-12-14
add logging of config object when in debug mode

flyinprogrammer:
Sometimes it's nice to able to see what's been parsed and about to be used as config in the beginning?

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
@@ -95,6 +96,8 @@ func NewApp(configFlag string) (*App, error) {
 	if err = cfg.InitLogging(); err != nil {
 		return nil, err
 	}
+	configJSON, _ := json.Marshal(*cfg)
+	log.Debugf("Loaded config: %v", string(configJSON))
 	a.PreStartCmd = cfg.PreStart
 	a.PreStopCmd = cfg.PreStop
 	a.PostStopCmd = cfg.PostStop
--- a/core/app.go
+++ b/core/app.go
@@ -96,8 +96,13 @@ func NewApp(configFlag string) (*App, error) {
 	if err = cfg.InitLogging(); err != nil {
 		return nil, err
 	}
-	configJSON, _ := json.Marshal(*cfg)
-	log.Debugf("Loaded config: %v", string(configJSON))
+	if log.GetLevel() >= log.DebugLevel {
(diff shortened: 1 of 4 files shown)


flyinprogrammer:
Similarly, and I can open a second PR, what would you think about adding:

```go
log.Infof("Service now registered.")
```

In the line following this one, https://github.com/joyent/containerpilot/blob/master/discovery/consul/consul.go#L123, so that we get some confirmation that we've successfully made the REST calls to register the service?

core/app.go
```diff
@@ -95,6 +96,8 @@ func NewApp(configFlag string) (*App, error) {
 	if err = cfg.InitLogging(); err != nil {
 		return nil, err
 	}
+	configJSON, _ := json.Marshal(*cfg)
```
flyinprogrammer: If you'd prefer I could make this something like this instead:

```go
if strings.ToLower(cfg.LogConfig.Level) == "debug" {
    configJSON, _ := json.Marshal(*cfg)
    log.Debugf("Loaded config: %v", string(configJSON))
}
```

So that we only marshal if we're in debug mode. I'm not sure how much you care about the extra overhead of this.

tgross: Probably not a bad idea. Note that logrus exports the log levels so you can do something like what they do in [`Logger.Debug`](https://github.com/sirupsen/logrus/blob/881bee4e20a5d11a6a88a5667c6f292072ac1963/logger.go#L178) rather than downcasing and whatnot.

tgross: Also we're skipping over the error here. Are we sure there's no case in which the JSON serialization can fail?

tgross:
> Similarly, and I can open a second PR, what would you think about adding:

Yeah, and I'm not wild about the message we have [at the top of that function](https://github.com/joyent/containerpilot/blob/348b2544c8eb7a4edb234665a0cace0da6b0262e/discovery/consul/consul.go#L110). Breaking the line up screws up log wrappers (like the Docker log drivers).

Maybe we make that something like:

```go
	if err := c.Agent().PassTTL(service.ID, "ok"); err != nil {
		log.Infof("service not registered: %v", err)
		if err = c.registerService(*service); err != nil {
		...
		log.Infof("service registered: %v", service.Name)
	}
```

If you want to add that to this PR I'm game for that.

tgross:
LGTM. Sorry about the delayed review for the holiday!

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/2.6.1

Issue #261, opened 2016-12-15
Define split function for use in containerpilot.json template?


deitch:
Any plans to add `split` to the `FuncMap` passed to the templates in https://github.com/joyent/containerpilot/blob/bad44bd445806614d91f1746a9722d88a043a8db/config/template.go#L48 ?

Very useful for converting strings (e.g. from env vars) to slices, which can then be used for `range` to iterate over actions.

tgross:
There aren't any plans to do so, but I'm not sure I understand how that would look in practice. Particularly because it sounds like the sort of thing that would break backwards compatibility (any env var w/ a `,` in it would now become an array, right?). Can you provide a code/template example here?

deitch:
> Particularly because it sounds like the sort of thing that would break backwards compatibility (any env var w/ a , in it would now become an array, right?)

Nope, not at all. Here is how it works:

```
{{ .SOMEVAR }}
```

The above returns the value of `$SOMEVAR` as a string, while:

```
{{ .SOMEVAR | split " " }}
```

returns a slice of strings, by splitting `$SOMEVAR` on space (or you could do `{{ .SOMEVAR | split "," }}` if you prefer commas, etc.

This in turn lets you iterate over values, e.g.:

```
{{ range $i := .SOMEVAR | split " " }}
// do stuff with $i
{{ end }}
```

Of course, with pure consul templates, that would be:

```
{{ range $i := env "SOMEVAR" }}
```

but containerpilot template already interpolates `.SOMEVAR` with `{{ env "SOMEVAR" }}`

tgross:
Ok, that all seems reasonable then. I'll mark this as an enhancement. Can you think of any other string-munging functions we should add to the template processing step?

And we would certainly welcome a PR if you feel like taking a crack at it!

deitch:
> Can you think of any other string-munging functions we should add to the template processing step?

Hmm, in addition to `split`, the inverse `join` makes sense. It may also make sense to have some form of `replace` and basic regex.

Consul template provides the following key ones: `split`, `join`, `regexMatch`, `replaceAll`, `regexReplaceAll`. Most of those are taken from pkg `regexp`, and not `strings`.

> And we would certainly welcome a PR if you feel like taking a crack at it!
I will try to find some time, but client work piling up. I did just contribute https://github.com/deitch/gopiloted ; is that my good deed for the day? :-)

tgross:
[...]

PR #265, opened 2017-01-04
Add split and join to templates

deitch:
I will try to add the `replace*` ones after, but here is a good start. Obviously includes a test line as well. I like how I just tested both in one line. :-)

Per the request for PR from @tgross at https://github.com/joyent/containerpilot/issues/261#issuecomment-270125563

### DIFF
--- a/config/template.go
+++ b/config/template.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 	"text/template"
 )
@@ -26,6 +27,22 @@ func join(sep string, s []string) (string, error) {
 		return "", nil
 	}
 	return strings.Join(s, sep), nil
+}
+
+// replace replaces all occurrences of a value in a string with the given
+// replacement value.
+func replace(from, to, s string) (string, error) {
+	return strings.Replace(s, from, to, -1), nil
+}
+
+// regexRplace replaces all occurrences of a regex in a string with the given
+// replacement value.
+func regexReplace(re, to, s string) (string, error) {
+	compiled, err := regexp.Compile(re)
+	if err != nil {
+		return "", err
+	}
+	return compiled.ReplaceAllString(s, to), nil
 }
 
 func parseEnvironment(environ []string) Environment {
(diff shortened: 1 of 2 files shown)


deitch:
OK, added `replace` and `regexReplace` along with an alias `regexpReplace`.

In your hands!

config/template.go
```diff
@@ -48,7 +82,12 @@ func defaultValue(defaultValue, templateValue interface{}) string {
 func NewTemplate(config []byte) (*Template, error) {
 	env := parseEnvironment(os.Environ())
 	tmpl, err := template.New("").Funcs(template.FuncMap{
-		"default": defaultValue,
+		"default":       defaultValue,
+		"split":         split,
+		"join":          join,
+		"replace":       replace,
+		"regexReplace":  regexReplace,
+		"regexpReplace": regexReplace,
```
tgross: Why the alias? If we were going to have more than one regex option at all I'd rather we match what `consul-template` did with having both a [`regexMatch`](https://github.com/hashicorp/consul-template#regexmatch) and [`regexReplaceAll`](https://github.com/hashicorp/consul-template#regexreplaceall).

If we don't want to implement the `regexMatch` bool (which doesn't seem unreasonable), then having the name be `regexReplaceAll` has the added advantage of giving us room to have a "replace _n_" option down the road.

tgross:
Just had the one comment about the field name but other than that LGTM.

config/template.go
deitch: > Why the alias? 

Um... just because? In all honesty, my first run of tests failed because I made that mistake! Seemed easiest to fix it by aliasing.

I don't care much one way or the other. If we want to use template's `replaceAll`, `regexMatch`, `regexReplaceAll`, I am fine with that. Whatever you want, just let me know,  happy to modify, rerun tests and push out.

tgross: Let's match `consul-template`, it'll be easier for end users.

deitch: FYI, consul template uses `replaceAll` instead of `replace`, so if we are going to match, that one too?

tgross:
LGTM. Will merge once Travis gives us the 👍

PR #281, opened 2017-02-17
Remove etcd

tgross:
This PR removes support for etcd in ContainerPilot v3. Removing this early in the development cycle will reduce the overall testing and development load.


misterbisson:
We'll also need to deprecate https://github.com/autopilotpattern/etcd

tgross:
We're getting an odd seemingly-unrelated integration test error here that bears looking into:

```
2017/02/17 17:58:36 TEST: test_envvars
Creating testenvvars_consul_1
Timed out waiting for Consul.
2017/02/17 17:58:47 FAIL: test_envvars
```

tgross:
I dug into the failed integration test and it appears to just be that the Consul target used by the env vars test sometimes takes longer than 10 seconds to self-elect.

I've increased the timeout and have been reliably able to get it to pass and we're now passing integration tests. In https://github.com/joyent/containerpilot/pull/282 I'll tweak the integration tests a bit so that we can share more of the test setup code and make the test targets a bit more clean / faster.

tgross:
> We'll also need to deprecate https://github.com/autopilotpattern/etcd

Done in https://github.com/autopilotpattern/etcd/pull/3. I've also updated the docs to remove the relevant sections.

misterbisson:
:pickles:

tgross:
(Holding this up for an internal discussion. Don't merge please.)

Issue #266, opened 2017-01-06
Use Context to cleanup timeouts


tgross:
Our timeout code is a bit hacky. Go 1.7 [moves `Context` into the stdlib](https://golang.org/doc/go1.7#context), and this library provides mechanisms for controlling goroutines.

Refs:
- https://blog.golang.org/context
- https://rakyll.org/leakingctx/

tgross:
Merged in https://github.com/joyent/containerpilot/issues/287

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-dev

Issue #206, opened 2016-08-12
pollables should timeout by default


tgross:
In 2.x, the `health`, `onChange`, and `sensor` timeouts added for https://github.com/joyent/containerpilot/issues/139 default to not having a timeout for backwards compatibility. In a future 3.x version, these should instead default to a timeout equal to the `poll` value. 

So for example this config would timeout after 10 seconds:

``` json
{
  "services": [
    {
      "name": "myService",
      "health": "/path/to/my/check.sh",
      "poll": 10,
      "ttl": 20
    }
  ]
}
```

This config would timeout after 5 seconds:

``` json
{
  "services": [
    {
      "name": "myService",
      "health": "/path/to/my/check.sh",
      "poll": 10,
      "ttl": 20,
      "timeout": "5s"
    }
  ]
}
```

And this config would never timeout:

``` json
{
  "services": [
    {
      "name": "myService",
      "health": "/path/to/my/check.sh",
      "poll": 10,
      "ttl": 20,
      "timeout": "0s"
    }
  ]
}
```

ref discussion in https://github.com/joyent/containerpilot/pull/199

tgross:
Merged in https://github.com/joyent/containerpilot/issues/287

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-dev

PR #287, opened 2017-02-23
Replace timeout code with Context and force default timeouts

tgross:
For https://github.com/joyent/containerpilot/issues/266 and https://github.com/joyent/containerpilot/issues/206

This replaces the timeout code we wrote for commands with `Context` (available in the stdlib after 1.7), which reduces the total amount of custom code and test surface we have. Note that in `RunWithTimeout` we can't simply instantiate the `Command` via `exec.CommandContext` because we need to kill the whole process tree. If we were to use the `exec.CommandContext` version then any children would continue to run after we run `Kill`.

cc @jasonpincin

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -142,8 +142,12 @@ func RunWithTimeout(c *Command, fields log.Fields) error {
 	go func() {
 		select {
 		case <-ctx.Done():
-			log.Warnf("%s timeout after %s: '%s'", c.Name, c.Timeout, c.Args)
-			c.Kill()
+			// if the context was canceled we don't want to kill the
+			// process because it's already gone
+			if ctx.Err().Error() == "context deadline exceeded" {
+				log.Warnf("%s timeout after %s: '%s'", c.Name, c.Timeout, c.Args)
+				c.Kill()
+			}
 		}
 	}()
 
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -129,13 +129,12 @@ func RunWithTimeout(c *Command, fields log.Fields) error {
 		// anyway in case the caller cares
 		return errors.New("Command for RunWithTimeout was nil")
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), c.TimeoutDuration)
-	defer cancel()
-
 	log.Debugf("%s.RunWithTimeout start", c.Name)
 	c.setUpCmd(fields)
 	defer c.closeLogs()
 	log.Debugf("%s.Cmd start", c.Name)
+	ctx, cancel := context.WithTimeout(context.Background(), c.TimeoutDuration)
+	defer cancel()
 	if err := c.Cmd.Start(); err != nil {
 		log.Errorf("unable to start %s: %v", c.Name, err)
 		return err
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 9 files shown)


tgross:
I'm digging into why the tests are failing on Travis... they're working on both my MacOS machine and in Docker for me. ☹️

tgross:
Actually turns out this tasks test is failing on my machine too. Oops! 😊  The sleep is for 200ms but the timeout _should_ be 500ms by the new default. It looks like all of the executions are failing.

```
2017/02/23 22:09:38 TEST: test_tasks
Creating network "testtasks_default" with the default driver
Creating testtasks_consul_1
Creating testtasks_app_1
Expected task1 to never time out: got 7
Going to remove testtasks_test_run_1
Removing testtasks_test_run_1 ... 
2017/02/23 22:09:51 FAIL: test_tasks
```

Likewise the unit test that's failing is getting timed out when it should be completing before the timeout:

```
=== RUN   TestOnChangeCmd
time="2017-02-23T22:06:34Z" level=info msg="Running doStuff with args: --debug" backend= process=onChange 
time="2017-02-23T22:06:34Z" level=warning msg="./testdata/test.sh timeout after 1s: '[doStuff --debug]'" 
time="2017-02-23T22:06:34Z" level=warning msg="killing command at pid: 402" 
time="2017-02-23T22:06:34Z" level=error msg="./testdata/test.sh exited with error: ./testdata/test.sh: signal: killed" 
--- FAIL: TestOnChangeCmd (0.00s)
```

So I suspect the problem is how we're setting the timeouts for these, and that's what I'll dig into next.

tgross:
I've done some sleuthing on my initial hypothesis with the unit test and it looks like the timeout is being set correctly but that the logging and setup of the command were blocking long enough that we're exceeding the 1 second timeout! I'll move the start of the timer down and this should clear up that the onChange test. I'll rebase the debugging commits and then tackle the tasks integration test errors.

tgross:
By adding a long sleep and debug logging to the end of one of the unit tests I've discovered that we're calling the timeout even after the process is exiting. It only seems to be happening with short commands and it's racy, however. So I suspect it's an issue with the scope of the `defer cancel()`

tgross:
Gah, I'm just misusing `cancel` entirely. [From the docs](https://blog.golang.org/context):

> WithTimeout returns a copy of parent whose Done channel is closed as soon as parent.Done is closed, cancel is called, or timeout elapses.

So when we hit cancel at the end we're firing `Done` even though the process is already gone. I've got a fix that I'm testing now.

tgross:
Ok, that fixed the tests, so I've pushed some minor cleanups and I need to update the docs to match, and then this will be good-to-go.

commands/commands.go
```diff
@@ -141,15 +132,49 @@ func RunWithTimeout(c *Command, fields log.Fields) error {
...
-		log.Errorf("Unable to start %s: %v", c.Name, err)
+		log.Errorf("unable to start %s: %v", c.Name, err)
+		return err
+	}
+	go func() {
+		select {
+		case <-ctx.Done():
+			// if the context was canceled we don't want to kill the
+			// process because it's already gone
+			if ctx.Err().Error() == "context deadline exceeded" {
```
jasonpincin: Hopefully Go is good about not changing error text without considering it a breaking change. I have no idea whether it is or not, but I've been burned in other languages.

jasonpincin:
That's all I could come up with. Code looks cleaner 👍

commands/commands.go
tgross: Yeah this is a very unfortunate design choice in the golang stdlib. There are very few cases where there's an error type wrapping `error` that you can assert types on, as they think this is an anti-pattern (wtf?). Fortunately we have several unit tests that will break if they change the text, so I think we're ok here.

tgross: Derp, just saw that I missed the `const` in the `context` package for this. Well in _most_ of the stdlib they're lacking errors... I'll fix this by checking vs the `DeadlineExceeded` error.

Issue #166, opened 2016-05-27
Allow telemetry server port to change on reload


tgross:
As per https://github.com/joyent/containerpilot/pull/165 and https://github.com/joyent/containerpilot/pull/123#issuecomment-208455770, the native golang HTTP server cannot be reloaded gracefully if we're on the same port. We backed out the change to try this in order to fix the bug, but this is something we definitely should support. There are a couple options here as I see it:
- Only attempt to reload the server if the port changes, and then just kill the old server. This would require some changes to what we pass into NewTelemetry because right now we don't consider the old Telemetry struct at all.
- Find a working library for graceful server reload that handles the socket correctly and replace the native server with that.

cc @misterbisson and @justenwalker for discussion

tgross:
@justenwalker wrote:
> There are wrappers around the http package that support reload though we might consider using: (See [graceful.go](https://github.com/tylerb/graceful/blob/master/graceful.go))

This seems to be the sort of thing we want but the developer is using gopkg.in so we'll need to fork them if we wanted to use it.

justenwalker:
If it's only a single go file under MIT license, we could possibly just copy the source we need provided we meet the MIT license requirements (which are [minimal](http://choosealicense.com/licenses/mit/))

tgross:
Oh, it looks like the author hasn't done something that another widely used library that shall remain nameless [\**cough*\*](https://github.com/go-yaml/yaml/tree/master) did, which is break `master` so that *only* gopkg.in works. We should be able use graceful using Godeps normally. Assuming it does  the job for us, of course.

tgross:
Marking this as a 2.x-compatible enhancement.

tgross:
This was added as a feature in golang1.8, which we'll support in v3

tgross:
Pre-released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-dev.3

PR #289, opened 2017-03-01
WiP: Initial implementation of HTTP control plane

cheapRoc:
This PR tracks discussion and implementation of ContainerPilot's control socket and integration endpoints. The following is a quick list of running tasks in order to complete a first pass of the functionality.

- [x] Parse out control socket configuration from `containerpilot.json`
- [ ] Initialize and serve HTTP endpoints through the control socket file.
- [ ] Implement PutEnviron `POST /v3/environ`
- [ ] Implement PutMetric `POST /v3/metric`
- [ ] Implement Reload `POST /v3/reload`
- [ ] Implement MaintenanceMode `POST /v3/maintenance`


cheapRoc:
So far I've got the general plumbing worked through but still non-functional. I was able to easily reuse much of Telemetry's Prometheus handler setup. Left off with my first unit test stating the control socket comes up but 404s... but I'll debug next chance I have.

I realize I don't need as much in the JSON `config/config` and `core/app` files but I needed something to bootstrap the socket behavior and test the init path.

And, for good measure, I also included the Mariposa RFD within the documentation dir.

Pretty basic so far but feedback appreciated!

PS: This is in that "pre-squash" phase so bear with my commits.

cheapRoc:
Found the misspelling right before I signed off for the evening. See if Travis enjoys that one. 😄

cheapRoc:
At some point I'll have some time to enforce those failing lint rules, fully implement `/status` handler API, and generally tighten things up more.

core/app.go
```diff
@@ -343,6 +349,9 @@ func (a *App) handlePolling() {
 		}
 		a.Telemetry.Serve()
 	}
+	if a.ControlSocket != nil {
```
tgross: We should start the control socket before we start any other work rather than here at the end of `handlePolling` Otherwise we won't be listening to the `preStart` handler or (if particularly unlucky) any of the other pollables if they manage to get scheduled and run before we can start listening on the socket.

If you look a bit higher-up in `Run` you'll see `handleSignals`. We should start the listener there, as that's what this control plane is going to replace. (Let's start it immediately afterwards for now; it'll be gone when we're done here.)

socket/control.go
```diff
@@ -0,0 +1,84 @@
...
+	Path string
+	mux *http.ServeMux
+	addr net.UnixAddr
+	listening bool
+	lock sync.RWMutex
+}
+
+// Handle `/status` control endpoint
+// NOTE: Stubbed out temporarily....
+func GetStatusHandler(w http.ResponseWriter, r *http.Request) {
```
tgross: This status handler should get served from the telemetry server, so let's move this handler into that package.

socket/control.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+var listener net.Listener
+
+// Serve our present ControlSocket
+func (cs *ControlSocket) Serve() {
+	cs.lock.Lock()
+	defer cs.lock.Unlock()
+
+	// ref https://github.com/joyent/containerpilot/pull/165
```
tgross: Now that we're on go1.8 we should be able to avoid the problems in that PR, as we have graceful shutdown. We're planning on implementing that for the Telemetry server (ref https://github.com/joyent/containerpilot/issues/166) but let's do that from the start for the socket server.

documentation/23-mariposa/README.md
```diff
@@ -0,0 +1,152 @@
+Title: Mariposa integration
```
tgross: The RFD is using the working name "Mariposa" but it's almost certain this isn't going to be what we call the container supervisor/project service. Let's just say "Triton integration" here for now.

documentation/23-mariposa/README.md
```diff
@@ -0,0 +1,152 @@
...
+----
+Text:
+
+There are a number of cases where we've wanted for lifecycle hooks to be able to
```
tgross: These docs get published to https://www.joyent.com/containerpilot and the publishing platform we use there is unfriendly to manually-wrapped text. So no need for line breaks here.

tgross:
Had a couple comments but looking good so far!

socket/control.go
```diff
@@ -0,0 +1,84 @@
+package socket
```
cheapRoc: This makes perfect sense, removing impl from naming. This was one of those 2am smart ideas as I had it as `control` originally. 😁

socket/control.go
cheapRoc: Sounds like I should read through [graceful.go](https://github.com/tylerb/graceful/blob/master/graceful.go) for context.

cheapRoc:
Thanks, sounds like we're aligned so far.

socket/control.go
tgross: That 3rd-party library is no longer necessary. It should be handled in the stdlib now.

core/app.go
cheapRoc: This was definitely neglected during my first pass so I've got the appropriate placing now. I'll also remove the stuff I put in the config parsing (for no reason at all).

Would you also like a CLI flag to override the default FS location of the socket file?

tgross: > Would you also like a CLI flag to override the default FS location of the socket file?

We've put all the config in the config file rather than flags (except for the optional config file location). I think for consistency we should keep it that way.

cheapRoc: Sounds good, I'll flesh it out a bit more.

socket/control.go
cheapRoc: I moved `socket/` back to `control/` but now receive the following failing lint rule...

```
control.ControlSocket [is used] by other packages, and that stutters; consider calling this Socket
```

I feel like `control.Server` gives better context within the application since we're serving HTTP.  I know `Telemetry` does also.

Thoughts?

tgross: `control.Server` sounds good to me.

cheapRoc:
Sweet, passing again. 😉 

I cleaned up some of the things we spoke about. Also continued working through bootstrapping and config parsing. Even though it's not a big addition, my config changes might not merge cleanly with any of the JSON -> YAML config format changes happening elsewhere in CP3.

I'm now starting to work through what the best approach should be for passing along App state into the control plane endpoint handlers. We can probably init the server with a pointer to App and/or pass through some sort of interface into handlers themselves. I'm still formalizing my thoughts on this behavior in code.

tgross:
> Sweet, passing again. 

👍 

> Even though it's not a big addition, my config changes might not merge cleanly with any of the JSON -> YAML config format changes happening elsewhere in CP3

Don't worry about that... the config format changes will be the last bit we do and I'm assuming there's going to be a bit of pain in that. We'll figure it out.

cheapRoc:
Nothing to ship here, this commit just exercises the config a bit more and HTTP handlers attached to App state. I also had a couple questions for yah...

core/app.go
```diff
@@ -373,3 +376,37 @@ func (a *App) stopCoprocesses() {
...
+	w.WriteHeader(http.StatusOK)
+	w.Header().Set("Content-Type", "application/json")
+	w.Write([]byte(`{ "message": true }`))
+}
+
+func (a *App) handleServer() *control.Server {
+	s := a.Server
+	s.Mux.HandleFunc("/status", a.getStatusHandler)
+	s.Mux.HandleFunc("/v3/env", a.postEnvHandler)
+	s.Serve()
```
cheapRoc: @tgross This setup does output Services as JSON through the socket at runtime. Sort of a control case, but nice to see when I manually tested within a running container...

How do you feel about this function (`handleServer`) staying in `core/app.go`?

core/app_test.go
```diff
@@ -262,6 +262,23 @@ func TestPidEnvVar(t *testing.T) {
 	}
 }
 
+func TestControlServerConfigParse(t *testing.T) {
+	jsonFragment := `{
+    "consul": "consul:8500",
+    "control": {
+	"socket": "/var/run/cp.socket"
+    }
```
cheapRoc: As of right now, this is required for the socket to exist. I need to add the default case so the socket always comes up.

telemetry/telemetry.go
```diff
@@ -96,11 +96,3 @@ func (t *Telemetry) Serve() {
 func (t *Telemetry) Shutdown() {
 	log.Debug("telemetry: shutdown received but currently a no-op")
 }
-
-// GetStatusHandler generates HTTP response for Telemetry status control plane
-// endpoint.
-func GetStatusHandler(w http.ResponseWriter, r *http.Request) {
-	w.WriteHeader(http.StatusOK)
-	w.Header().Set("Content-Type", "application/json")
-	w.Write([]byte(`{ "message": true }`))
```
cheapRoc: This will return.... I just moved it temporarily but forgot to simply comment out.

cheapRoc:
Ahhh... yeah that's the old test I still had committed. I'm still writing one under `core/app.go` if that's staying there.

core/app.go
cheapRoc: A little more background on my question. We have two options here and I wasn't sure which you thought was clearer for the code base...

1. We can push the App object/pointer through the call to `s.Serve()` and provide it to the handlers by pushing it through more hoops and function calls...
2. Or... we can simply implement the App handlers on App itself and perform the work close to the existing implementations of `Reload()` and etc.

I'm siding on option 2 but I'm open to suggestions. Also unclear if there's a better Golang primitive or pattern.

tgross: I'd like to minimize the surface area of `App` as much as possible. As a bit of history here the original interface of both `core/App` and `config/Config` was a lot smaller but as we added features we didn't do an awesome job of refactoring the internals to match the greater complexity. 

In https://github.com/joyent/containerpilot/pull/290 I'm going work to make the various packages pass messages on an event bus. So the more strongly you can separate `core/App` from the server the better this will all come together when it's time to integrate them.

cheapRoc: Glad I asked, internal events within the main process will definitely help control surface. Maybe even direct the bus straight out of a websocket? ;)

I'll break up what I have and keep future integration in mind for when #290 drops.

cheapRoc:
Hey @tgross, I want to pick this back up. Apologize for the delay but I'm also following your work in #290. Couple questions...

- Should the socket file and control server be init'd via a subscriber watching for the GlobalStartup event?
- Does PutEnviron need to support multiple commands? i.e. are we updating env vars across multiple commands or using the same across all commands? (does #290 change the HTTP handlers at all?)
- Should HTTP events be published on the bus so that multiple handlers can subscribe and react... or... is that overuse of the event bus sub system?

The event bus definitely helps my implementation. I'll rebase off your branch once it's stabilized.

tgross:
> Should the socket file and control server be init'd via a subscriber watching for the GlobalStartup event?

We start up the telemetry server (ref [core/app.go#L259](https://github.com/tgross/containerpilot/blob/6910cdd4fdf30faca52a228102b89801d2ffec26/core/app.go#L259)) right after we start all the subscribers but before we fire off the `GlobalStartup` event. I imagine we'd want to start the control server at the same time.

> Does PutEnviron need to support multiple commands? i.e. are we updating env vars across multiple commands or using the same across all commands? (does #290 change the HTTP handlers at all?)

The `PutEnviron` handler should operate by updating the env vars of the ContainerPilot process via [`os.Setenv`](https://golang.org/pkg/os/#Setenv). The current environment is by default inherited by a child processes whenever we fork. We can't update the env vars of already-running child processes without cooperation from the child process. So there's no change needed for the HTTP handler as far as I know.

> Should HTTP events be published on the bus so that multiple handlers can subscribe and react... or... is that overuse of the event bus sub system?

I _think_ we'll want the handlers to mostly call functions we already have wired up to the event bus like [`ToggleMaintenanceMode`](https://github.com/tgross/containerpilot/blob/6910cdd4fdf30faca52a228102b89801d2ffec26/core/app.go#L177) and [`Reload`](https://github.com/tgross/containerpilot/blob/6910cdd4fdf30faca52a228102b89801d2ffec26/core/app.go#L223). So publishing events for the handlers themselves is probably redundant (and super easy to add on later if we need it).

cheapRoc:
So not much is changing in terms of where I was headed, that's good.

cheapRoc:
Going to close this as I have another branch with the same work ported overtop the latest master.

Issue #211, opened 2016-08-19
ContainerPilot without service discovery


thoughtarray:
Great project, really like the pattern.  This project really puts together a lot of other stuff that we had been trying to solve with [S6](https://github.com/just-containers/s6-overlay) such as startup and shutdown behaviors.

It would be nice to use ContainerPilot for the behaviors and _opt out_ of the service discovery bits.  Would this be possible?

fitz123:
You could skip service defining part and it will not try to connect to Service Discovery, but you still have to provide any dumb address for consul, though. I worked for me before

tgross:
This was raised in https://github.com/joyent/containerpilot/issues/117 and has the same configuration option -- simply elide the `services` block and ContainerPilot will never fire health checks and heartbeats to Consul. Or elide the `backends` block if you have no dependencies you care about.

The service catalog option (`consul` or `etcd`, currently) is a required parameter. In fact it's the only required parameter. But if you don't have `services`, `backends`, or `telemetry` blocks then there's never any requests to it.

thoughtarray:
Thanks for the information.  I tried to make sure I wasn't asking a question that was already asked.  I'll play around with it and see how it works.

Aside, I would like to see the consul or etcd options optional.  Or maybe have a noop catalog option.  That way people could use ContainerPilot for behaviors, tasks, and coprocesses without it feeling hacky.

tgross:
If those fields were optional it'd be very easy to have an invalid configuration so we'd have to work out the expected behavior. If you'd like to open an issue w/ a proposal where we could suss out that behavior we could at least discuss it.

Closing this issue as answered.

justenwalker:
@tgross if we added a service discovery option explicitly, we could enforce that Service and Backends are *not* configured and fail if they are.

However, it may be a bit awkward in the config since it's a key and needs a value:

```json
{
  "noop": null
}
```
the null value does not matter here, what matters is the key only.

thoughtarray:
@tgross: I will put something intelligible together soon and open a proposal.

@justenwalker: Yeah, it is kind of awkward.  I thought it was weird that the service catalog option was a top-level key as opposed to a nested dict, but it does make configuration very easy.

```json
{
  "consul": "...",
  ...
}
```
--- vs ---
```json
{
  "service_catalogs": [
    {"type": "consul", "host": "foo:8500"},
    {"type": "zk", "host": "foo:2181", "service_root": "sd/prod/bar-app/"}
  ],
  ...
}

PR #291, opened 2017-03-07
add discovery disable flag

riot-jayb:
Using containerpilot without discovery was raised in #211.

Here is my simple proposal to add a --no-discovery flag for containerpilot. If we don't find any configured backends while the --no-discovery flag is true, then ignore the error when parsing the config instead of dying with a fatal error.

The main concern raised in #211 was that under normal usage the error about not defining any backends is a valid one. In this case the user needs to opt-in with a flag, so they would never need a configuration warning.

Seems to be working well for me in a development build. Still evaluating this project for using in production :)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -186,7 +186,7 @@ func RenderConfig(configFlag, renderFlag string) error {
 }
 
 // ParseConfig parses a raw config flag
-func ParseConfig(configFlag string) (*Config, error) {
+func ParseConfig(configFlag string, noDiscoveryFlag bool) (*Config, error) {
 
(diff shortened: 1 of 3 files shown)


tgross:
Thank you very much for the PR, but this isn't the approach I'd like to take with this. In ContainerPilot v3 we're merging the concepts of services, tasks, and coprocesses ([ref RFD86](https://github.com/joyent/rfd/blob/master/rfd/0086/multiprocess.md). Services that don't need advertisement will be able to simply exclude that section of the configuration.

PR #292, opened 2017-03-08
Ability to configure telemetry service name

commarla:
This PR add the option to configure the telemetry service name.

### DIFF
--- a/documentation/19-telemetry/README.md
+++ b/documentation/19-telemetry/README.md
@@ -16,6 +16,7 @@ A minimal configuration for ContainerPilot including telemetry might look like t
 {
   "consul": "consul:8500",
   "telemetry": {
+    "servicename": "containerpilot",
     "port": 9090,
     "interfaces": ["eth0"],
     "tags": ["tag1"],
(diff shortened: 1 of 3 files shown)


tgross:
Thanks for the PR @commarla! I'm not sure we want to do this, however.

Leaving out configurability of that service name was intentional in the original design (ref https://github.com/joyent/containerpilot/issues/27#issuecomment-198631596). I'm not sure about the reasoning _why_ though because unfortunately it wasn't written down. I'm pretty sure there was a sidebar discussion with @misterbisson but I don't recall the details... maybe he has some insight.

Design intent aside, before we could accept this PR we'd need to have a test demonstrating it works, that we check for invalid names, etc.

commarla:
Hi @tgross,
Thanks for the quick reply.
I've just add the test.

misterbisson:
@commarla it's very intentional that the ContainerPilot telemetry service name is not configurable. As @tgross pointed out, there were a lot of internal discussions on that point, as well as [some in the PR](https://github.com/joyent/containerpilot/pull/119#issuecomment-206955477).

The reasoning for this is our work to build a library of interoperable applications, as well as tools to manage them. Such a plan depends on well-known interfaces, and making the interface name configurable adds complexity well beyond the code changes here.

An example of the integration planned for this interface is in [RFD86](https://github.com/joyent/rfd/blob/master/rfd/0086/mariposa.md), which builds on [additional features proposed in the original ticket](https://github.com/joyent/containerpilot/issues/27#issuecomment-207462572) to report what's in the container, what it depends on, and what its status is.

With that roadmap and reasoning explained, I'd also like to better understand your motivation for wanting such a change?

tgross:
Going to close this PR but I'm open to discussing it in a new issue if there are strong feelings about the design rationales laid out by @misterbisson in https://github.com/joyent/containerpilot/pull/292#issuecomment-291155741

Issue #295, opened 2017-03-13
timing issues with multiple tasks


riot-jayb:
Hi, 

I'm seeing a strange issue when using multiple tasks with small frequencies. Either the tasks are sometimes not executed, or else just not logged. I have a "heartbeat" with 15s frequency and "backup" task with 20s frequency (these are just stubs for testing the config atm), and the logs give very inconsistent results.

Here is my containerpilot config:

```
{
  "postStop" : "percona-mngr leave",
  "tasks": [
    {
      "name": "heartbeat",
      "command": "percona-mngr heartbeat",
      "frequency": "15s"
    },
    {
      "name": "backup",
      "command": "percona-mngr backup",
      "frequency": "20s"
    }
  ]
}
```

And here is the output log during container run time:

```2017/03/13 05:07:01 percona-mngr heartbeat (stub)
2017/03/13 05:07:11 percona-mngr backup (stub)
2017/03/13 05:07:16 percona-mngr heartbeat (stub)
2017/03/13 05:07:31 percona-mngr heartbeat (stub)
2017/03/13 05:07:46 percona-mngr heartbeat (stub)
2017/03/13 05:07:51 percona-mngr backup (stub)
2017/03/13 05:08:16 percona-mngr heartbeat (stub)
2017/03/13 05:08:46 percona-mngr heartbeat (stub)
2017/03/13 05:08:51 percona-mngr backup (stub)
2017/03/13 05:09:16 percona-mngr heartbeat (stub)
2017/03/13 05:09:46 percona-mngr heartbeat (stub)
2017/03/13 05:09:51 percona-mngr backup (stub)
2017/03/13 05:10:11 percona-mngr backup (stub)
2017/03/13 05:10:16 percona-mngr heartbeat (stub)
2017/03/13 05:10:31 percona-mngr heartbeat (stub)
2017/03/13 05:10:46 percona-mngr heartbeat (stub)
2017/03/13 05:10:51 percona-mngr backup (stub)
2017/03/13 05:11:01 percona-mngr heartbeat (stub)
2017/03/13 05:11:11 percona-mngr backup (stub)
2017/03/13 05:12:16 percona-mngr heartbeat (stub)
2017/03/13 05:13:16 percona-mngr heartbeat (stub)
2017/03/13 05:13:31 percona-mngr heartbeat (stub)
2017/03/13 05:13:46 percona-mngr heartbeat (stub)
2017/03/13 05:13:51 percona-mngr backup (stub)
[...]

Issue #294, opened 2017-03-09
coprocess logs missing from container output


sourcec0de:
Hey guys, awesome work on container pilot. This is way easier than stuffing supervisor into a container.

I've got an issue where the `coprocesses` output is not being captured.

Here's my container pilot installation script.
```
ENV CONTAINERPILOT_VERSION 2.7.0
ENV CONTAINERPILOT file:///etc/containerpilot.json
RUN export CP_SHA1=687f7d83e031be7f497ffa94b234251270aee75b \
    && curl -Lso /tmp/containerpilot.tar.gz \
         "https://github.com/joyent/containerpilot/releases/download/${CONTAINERPILOT_VERSION}/containerpilot-${CONTAINERPILOT_VERSION}.tar.gz" \
    && echo "${CP_SHA1}  /tmp/containerpilot.tar.gz" | sha1sum -c \
    && tar zxf /tmp/containerpilot.tar.gz -C /bin \
    && rm /tmp/containerpilot.tar.gz
```

Here's my `containerpilot.json`
```
{
  "consul": "consul:8500",
  "preStart": "/etc/lifecycle/preStart",
  "coprocesses": [
    {
      "name": "sync-nginx",
      "command": ["python3", "sync-nginx.py"],
      "restarts": "unlimited"
    }
  ]
}
```

I can see the process is running but the output is missing from the logs.
The process isn't crashing and it's working as expected. Any ideas?

Thanks again.

tgross:
Are you getting _any_ log output from the container? For example, are you seeing the output of your `preStart` and/or the main process?

In any event, can you run with debug logging and drop the output here? Add the following section to your config:

```json
  "logging": {
    "level": "DEBUG",
    "format": "default",
    "output": "stdout"
  }
```

sourcec0de:
@tgross it looks like the output was just really delayed.

I just opened my laptop again and the output was there. Odd.

tgross:
That's strange. Can you tell me a bit about where you're deploying this? Is this all in Docker on your laptop or something like Triton?

sourcec0de:
@tgross 

Locally: Docker for mac via `docker-compose`
Production: I'm on Google Container Engine (GKE) kubernetes.

tgross:
And you saw this behavior in both environments?

tgross:
Possibly related: https://github.com/joyent/containerpilot/issues/295

sourcec0de:
@tgross I did see this issue on both environments.

[...]

Issue #267, opened 2017-01-11
Container pilot should log failed health check


rickardrosen:
We spent some time trying to figure out why containerpilot did not register our container with service discovery (consul).

Turned out that, for whatever reason, we had the wrong path for our health check...
Our bad, but we could not see anything logged from containerpilot regarding the initial failed health check, even after enabling debug logging.

Even though this was a configuration error, it would be great to be able to pick this kind of issue up immediately through CP logging, or did we overlook something here?

tgross:
Yeah, you definitely should be seeing logs if the command has the wrong path. [`RunWithTimeout`](https://github.com/joyent/containerpilot/blob/9397588538e42b6fca0876d9b57758681917a548/commands/commands.go#L133-L153) is what runs the commands in the health check. Assuming your logging in debug you should be seeing something like this in that situation:

```
2017/01/11 09:00:00 mycommand.RunWithTimeout start
2017/01/11 09:00:00 mycommand.Cmd.Start
2017/01/11 09:00:00 Unable to start mycommand: command not found: /usr/bin/nonexistingcommand
```

So either there's a bug or something not quite right in the config. Can you provide the version of ContainerPilot you're using and some details about your logging configuration?

tgross:
Without more information I believe this is another case of #295, which I'll be opening a backported fix for shortly.

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/2.7.1. I'm going to mark this as fixed and close this issue, but feel free to reopen if it looks like we missed something.

PR #297, opened 2017-03-23
backport fix for dropped logs to v2

tgross:
For #294 and #295, possibly #267.

Instead of attaching the logger's `io.WriteCloser` at each execution, where it's possibly dropped when it goes out of scope before it can be flushed, we move the `io.WriteCloser` to the `Command` that wraps the stdlib `os/exec.Cmd` and only close it explicitly when we throw away the `Command` for reloading config.

cc @riot-jayb @sourcec0de @rickardrosen


tgross:
Ok just pushed up a fix for the null-pointer I missed in build [214410759](https://travis-ci.org/joyent/containerpilot/builds/214410759). I'll check in on the new build in the morning and cut a patch release (2.7.1) if all is well.

tgross:
Sometimes I hate linting 😬  From [214420535](https://travis-ci.org/joyent/containerpilot/builds/214420535)
```
/cp/src/github.com/joyent/containerpilot/commands/commands.go:226:1: exported method Command.CloseLogs should have comment or be unexported
Found 1 lint suggestions; failing.
make: *** [lint] Error 1
The command "make lint build" exited with 2.
```

Pushed that fix and rebased. Once this is rebuilt I'll cut the patch release.

PR #300, opened 2017-03-25
Initial implementation of HTTP control plane

cheapRoc:
This includes two commits, one for the configuration work and a second for the HTTP server over control socket. I've included a simple endpoint for testing and would definitely like feedback.

I also need to test a manual build of this version of CP3 inside a live container, just to make sure the unit tests are telling the truth.

### DIFF
--- a/control/config.go
+++ b/control/config.go
@@ -13,13 +13,13 @@ const (
 // Config represents the location on the file system which serves the Unix
 // control socket file.
 type Config struct {
-	Socket    string      `mapstructure:"socket"`
+	SocketPath string `mapstructure:"socket"`
 }
 
 // NewConfig parses a json config into a validated Config used by control
 // Server.
 func NewConfig(raw interface{}) (*Config, error) {
-	cfg := &Config{Socket: DEFAULT_SOCKET} // defaults
+	cfg := &Config{SocketPath: DEFAULT_SOCKET} // defaults
 	if raw == nil {
 		return cfg, nil
 	}
--- a/control/config.go
+++ b/control/config.go
@@ -1,0 +1,42 @@
+package control
+
+import (
+	"fmt"
+
+	"github.com/joyent/containerpilot/utils"
+)
+
+const (
+	DEFAULT_SOCKET = "/var/run/containerpilot.socket"
+)
+
+// Config represents the location on the file system which serves the Unix
+// control socket file.
+type Config struct {
+	Socket    string      `mapstructure:"socket"`
+}
+
+// NewConfig parses a json config into a validated Config used by control
+// Server.
+func NewConfig(raw interface{}) (*Config, error) {
+	cfg := &Config{Socket: DEFAULT_SOCKET} // defaults
+	if raw == nil {
+		return cfg, nil
+	}
+
+	if err := utils.DecodeRaw(raw, cfg); err != nil {
+		return nil, fmt.Errorf("control config parsing error: %v", err)
+	}
+
+	if err := cfg.Validate(); err != nil {
+		return nil, fmt.Errorf("control config validation error: %v", err)
+	}
+
+	return cfg, nil
+}
+
+// Validate parsed control configuration and the values contained within.
+func (cfg *Config) Validate() error {
+	// TODO: Validate NestedConfig and socket's file system location ...
+	return nil
+}
--- a/control/config.go
+++ b/control/config.go
@@ -29,15 +29,5 @@ func NewConfig(raw interface{}) (*Config, error) {
 		return nil, fmt.Errorf("control config parsing error: %v", err)
 	}
 
-	if err := cfg.Validate(); err != nil {
-		return nil, fmt.Errorf("control config validation error: %v", err)
-	}
-
 	return cfg, nil
-}
-
-// Validate parsed control configuration and the values contained within.
-func (cfg *Config) Validate() error {
-	// TODO: Validate NestedConfig and socket's file system location ...
-	return nil
 }
--- a/control/config.go
+++ b/control/config.go
@@ -7,7 +7,8 @@ import (
 )
 
 const (
-	DEFAULT_SOCKET = "/var/run/containerpilot.socket"
+	// DefaultSocket is the default location of the unix domain socket file
+	DefaultSocket = "/var/run/containerpilot.socket"
 )
 
 // Config represents the location on the file system which serves the Unix
@@ -19,7 +20,7 @@ type Config struct {
 // NewConfig parses a json config into a validated Config used by control
 // Server.
 func NewConfig(raw interface{}) (*Config, error) {
-	cfg := &Config{SocketPath: DEFAULT_SOCKET} // defaults
+	cfg := &Config{SocketPath: DefaultSocket} // defaults
 	if raw == nil {
 		return cfg, nil
 	}
--- a/control/control.go
+++ b/control/control.go
@@ -12,11 +12,13 @@ import (
 )
 
 const (
-	SOCKET_TYPE = "unix"
+	// SocketType is the default listener type
+	SocketType = "unix"
 )
 
-// Server contains the state of the HTTP Server used by ContainerPilot's HTTP
-// transport control plane. Currently this is listening via a UNIX socket file.
+// HTTPServer contains the state of the HTTP Server used by ContainerPilot's
+// HTTP transport control plane. Currently this is listening via a UNIX socket
+// file.
 type HTTPServer struct {
 	mux        *http.ServeMux
 	addr       net.UnixAddr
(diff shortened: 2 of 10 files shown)


cheapRoc:
I'll checkout the integration tests. Didn't get the chance to run those yet or manually test.

cheapRoc:
Manual test looks like the control socket works as expected (for the time being). Running `curl -v --unix-socket /var/run/containerpilot.socket http://unix/env` retrieves JSON version of `os.Environ`.

```
$ docker run test-cp3
time="2017-03-26T03:03:46Z" level=debug msg="loaded config: {\"Discovery\":{},\"LogConfig\":{\"level\":\"DEBUG\",\"format\":\"text\",\"output\":\"stdout\"},\"StopTimeout\":5,\"Services\":null,\"Checks\":null,\"Watches\":null,\"Telemetry\":null,\"Control\":{\"SocketPath\":\"/var/run/containerpilot.socket\"}}"
time="2017-03-26T03:03:46Z" level=info msg="control: Serving at /var/run/containerpilot.socket"
time="2017-03-26T03:03:46Z" level=debug msg="event: {Startup global}"
time="2017-03-26T03:04:45Z" level=debug msg="Marshaled environ: [\"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin\",\"HOSTNAME=3fb0e2e8c8bb\",\"CONTAINERPILOT=file:///etc/containerpilot.json\",\"HOME=/root\",\"CONTAINERPILOT_PID=1\"]"
```

Unsure about the integration tests, quite a few are failing.

cheapRoc:
Looks like an issue with the socket. Still debugging but rather obvious once I tested with an autopilot pattern using nginx.

cheapRoc:
Looks like only a few lints are failing now. I'd like to allow all caps for constants otherwise I can easily solve the rest.

control/control.go
```diff
@@ -0,0 +1,104 @@
...
+func (s *HTTPServer) Serve() {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	// ref https://github.com/joyent/containerpilot/pull/165
+	if listener != nil {
+		return
+	}
+
+	s.mux.HandleFunc("/env", s.getEnvHandler)
```
tgross: We should URL-version this endpoint: `/v3/env`

control/control.go
```diff
@@ -0,0 +1,104 @@
...
+		log.Infof("control: Serving at %s", s.addr.String())
+		log.Fatal(http.Serve(ln, s.mux))
+		log.Debugf("control: Stopped serving at %s", s.addr.String())
+	}()
+}
+
+// Shutdown shuts down the control server
+func (s *HTTPServer) Shutdown() {
+	s.listening = false
+	log.Debug("control: Shutdown received but currently a no-op")
```
tgross: Now that we're on golang 1.8 we can use [Server.Shutdown](https://golang.org/pkg/net/http/#Server.Shutdown) on the server to do a graceful shutdown. That should let us get rid of the `listening` flag too I think.

tests/assert/assert.go
```diff
@@ -21,3 +21,13 @@ func Error(t *testing.T, err error, expected string) {
 		t.Fatalf("expected '%s' but got '%s'", expected, err.Error())
 	}
 }
+
+// True asserts that an interface is equal to boolean "true"
+func True(t *testing.T, expected interface{}) {
```
tgross: Can we match the signature of [testify's `assert.True`](https://godoc.org/github.com/stretchr/testify/assert#True) so that we can drop-in-replace it later?

control/control.go
```diff
@@ -0,0 +1,104 @@
...
+	"net"
+	"net/http"
+	"os"
+	"sync"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+const (
+	SOCKET_TYPE = "unix"
```
tgross: Although every sane language wants the constants in all-caps, in golang capitalization effects whether we export the variable, so we'll want this to be all lower-case as the linter wants.

control/control.go
```diff
@@ -0,0 +1,104 @@
...
+	// ref https://github.com/joyent/containerpilot/pull/165
+	if listener != nil {
+		return
+	}
+
+	s.mux.HandleFunc("/env", s.getEnvHandler)
+
+	ln, err := net.Listen(s.addr.Network(), s.addr.String())
+	if err != nil {
+		log.Fatalf("Error serving socket at %s: %v", s.addr.String(), err)
```
tgross: _Super_ nitpicky but I'm trying to get all the error messages so they start with lower case unless starting with a proper name, so that when errors wrap each other they look nice. (Apparently this is a golang style thing.)

control/control.go
cheapRoc: I hear yah. It's not a biggie and helps contributions from the community. But still not a fan. Constants should easily read in an API and stand out in the middle of an implementation. Helps narrow down on what defaults are versus what internal data structures are representing. Maybe it's the long lost C programmer in me.

tests/assert/assert.go
cheapRoc: I can match it but when can we include testify?

cheapRoc:
This is also screaming for an integration test. I didn't have the spare time yet but it's on my list.

control/control.go
tgross: Believe me, me too. 😀

tests/assert/assert.go
tgross: I didn't include it yet just because we only had the two assert methods we needed. I'm open to including it whenever we feel like we've crossed the line where it's silly to implement ourselves. So if you feel like it makes sense now then feel free to go for it.

control/control.go
cheapRoc: Thanks for the reminder, I'll look into this now that there's an initial draft of the control server.

cheapRoc:
Looks like my most recent push just failed due to that timing test again...

```
=== RUN   TestServiceRunPeriodic
time="2017-04-01T01:28:02Z" level=warning msg="myservice timeout after 10ms: '[doStuff runPeriodicTest]'" 
--- FAIL: TestServiceRunPeriodic (0.12s)
	services_test.go:151: expected 9 or 10 task fires but got 7
		[{Startup global} {ExitSuccess myservice} {ExitSuccess myservice} {ExitSuccess myservice} {ExitSuccess myservice} {ExitSuccess myservice} {ExitSuccess myservice} {ExitSuccess myservice} {Stopping myservice} {Stopped myservice}]
FAIL
exit status 1
FAIL	github.com/joyent/containerpilot/services	0.351s
```

Must be the same ghost I was chasing recently.

tgross:
> Looks like my most recent push just failed due to that timing test again...

Yeah I'm not wild about that test. I suspect it's inherently flaky because the timers are running async to the test. I'll open a PR to try to change how that test works.

control/control.go
tgross: You know what, let's do that under a separate PR so we can move towards merging this sooner rather than later.

tests/assert/assert.go
tgross: Likewise here, let's fix-up this signature and then we can drop-in testify under a separate PR.

control/config.go
```diff
@@ -0,0 +1,43 @@
...
+	if err := cfg.Validate(); err != nil {
+		return nil, fmt.Errorf("control config validation error: %v", err)
+	}
+
+	return cfg, nil
+}
+
+// Validate parsed control configuration and the values contained within.
+func (cfg *Config) Validate() error {
+	// TODO: Validate NestedConfig and socket's file system location ...
```
tgross: There doesn't seem to be much to validate here. We can get a time-of-check-time-of-use problem if we check the socket file system location, and the error we get back when we bind to the listener is both fatal and pretty good: `listen unix /path/to/my.sock: bind: no such file or directory`. I think we can just skip this, at least for now.

tgross:
I think I've fixed the test flakiness in https://github.com/joyent/containerpilot/pull/303, so let's revert e68bb40.

Otherwise I think this is looking pretty good. Let's hit the last couple open comments and this should be good to merge.

control/control.go
cheapRoc: Sounds good, I have a bunch of work around HTTP handlers. Testing can go into this once that becomes more defined and stable.

tests/assert/assert.go
cheapRoc: I push this a little later then.

control/config.go
cheapRoc: Makes perfect sense, we're only handling the config at this point. I had left it to see if there was any validation you'd want.

tgross:
@cheapRoc I think this is ready to get merged, we just need to get the merge conflict resolved.

cheapRoc:
That should be it once those tests pass.

tests/assert/assert.go
tgross: For the record, the answer to when we might adopt testify at the moment is "when they fix their dependency chain." They've vendored their own dependencies even though they're a library, so I'm really not interested into wading into that mess right now (and `dep` is still in alpha)

PR #307, opened 2017-04-05
Jobs not services

tgross:
Without changing the configuration syntax in any way, this renames the package `services` to `jobs`, along with all related types. This will reflect the configuration changes in https://github.com/joyent/rfd/pull/31


jobs/config.go
```diff
@@ -111,7 +111,7 @@ func NewConfigs(raw []interface{}, disc discovery.Backend) ([]*Config, error) {
 // Validate ensures that a Config meets all constraints
 func (cfg *Config) Validate(disc discovery.Backend) error {
 	if disc != nil {
-		// non-advertised services don't need to have their names validated
+		// non-advertised jobs don't need to have their names validated
 		if err := utils.ValidateServiceName(cfg.Name); err != nil {
```
geek: `ValidateServiceName` -> `ValidateJobName` ?

jobs/jobs.go
```diff
@@ -0,0 +1,235 @@
...
+	restartLimit   int
+	restartsRemain int
+	frequency      time.Duration
+
+	events.EventHandler // Event handling
+}
+
+// NewJob creates a new Job from a Config
+func NewJob(cfg *Config) *Job {
+	service := &Job{
```
geek: `service` -> `job`

jobs/jobs.go
```diff
@@ -0,0 +1,235 @@
...
+func FromConfigs(cfgs []*Config) []*Job {
+	services := []*Job{}
+	for _, cfg := range cfgs {
+		service := NewJob(cfg)
+		services = append(services, service)
+	}
+	return services
+}
+
+// SendHeartbeat sends a heartbeat for this service
```
geek: `// SendHeartbeat sends a heartbeat for this service` -> `// SendHeartbeat sends a heartbeat for this job` ?

jobs/config.go
tgross: Not in that case, actually. We want to validate a service name in the Consul sense of the term.

jobs/jobs.go
```diff
@@ -0,0 +1,235 @@
...
+func (job *Job) SendHeartbeat() {
+	if job.discoveryService != nil || job.Definition != nil {
```
geek: should we rename to `discoveryCatalog` ?

jobs/jobs.go
tgross: Arguably not here either, as this is sends for service definition and not the job.

jobs/jobs.go
tgross: Couldn't hurt

tgross:
I think I've hit all your comments @geek and a couple others I'd missed. The big thing here to do was the package and struct rename so that my later changesets are smaller. 😀

geek:
LGTM

PR #308, opened 2017-04-05
transparently swap out JSON config parsing for JSON5

tgross:
This PR swaps out the JSON config parser for a JSON5 config parser (per https://github.com/joyent/rfd/pull/31). This doesn't change out any of the expected configuration, so this should work transparently.


misterbisson:
:pickles:

geek:
LGTM

PR #310, opened 2017-04-05
catch the 'waitid: no child processes' message after exec wait

tgross:
For https://github.com/joyent/containerpilot/issues/309

The `os` package says we should be getting [`wait: no child processes`](https://golang.org/src/os/exec_unix.go?s=962:989#L22) from our `wait` call. But we're also getting [`waitid: no child processes`](https://golang.org/src/os/wait_waitid.go?s=1174:1202#L37) from the earlier [`blockUntilWaitable`](https://golang.org/src/os/exec_unix.go?s=482:504#L22). Both "errors" are harmless to ignore as it just means we've lost the race to reap any children with the signal handler.

ref https://github.com/joyent/containerpilot/pull/222 for the previous work done on this subject.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -14,7 +14,10 @@ import (
 	"github.com/joyent/containerpilot/events"
 )
 
-const errNoChild = "wait: no child processes"
+const (
+	errWaitNoChild   = "wait: no child processes"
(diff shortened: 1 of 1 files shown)


geek:
LGTM, happy to see this fixed.

Issue #299, opened 2017-03-25
Move JSON fragments into tests/fixture package


cheapRoc:
I think it would be beneficial to move many of the JSON fragments into their own fixtures directory/package, possibly under `tests/`. It'll be easier to ensure code formatting rules of certain editors and remove the possibility of reformats causing failures for some JSON parser tests. Some of these tests (`core/app_test.go`) look for specific ranges like lines/cols and editors that reformat will break the tests. `git status` is the only way to see the difference and can lead to some headaches.

This is also super simple and could help contributor dev experience, especially other people that still use Emacs (*cough*me*cough*).

I can submit a PR for this but I wanted to double check first.

cheapRoc:
Also, I don't believe `gofmt` honors reformatting string literals, which is the exact problem I have.

cheapRoc:
Actually, if CP is moving to YAML I might not even need to do this.

tgross:
It does seem like a good idea even with a config language change -- the tiny single-line snippets in some of the tests don't really need it but certainly for some of the big ones this makes sense.

tgross:
Done in #311

PR #311, opened 2017-04-06
move some configuration test fixtures into their own testdata files

tgross:
For https://github.com/joyent/containerpilot/issues/299

This moves a bunch of the larger test configuration JSON5 snippets into their own test fixture files under the `testdata/` directory for their respective packages. This will make it a bit easier to edit test configs without making silly syntax mistakes, as well as making it easier to diff them in code reviews. It's entirely possible I missed one or two that we might want to move, but we can do that incrementally as part of other work -- it's a judgement call as to "at what point does this chunk of text want to be its own file"

cc @cheapRoc

### DIFF
--- a/checks/config_test.go
+++ b/checks/config_test.go
@@ -1,6 +1,8 @@
 package checks
 
 import (
+	"fmt"
+	"io/ioutil"
 	"testing"
 
 	"github.com/joyent/containerpilot/tests"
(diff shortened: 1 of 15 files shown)


tgross:
Oops... I'd changed how I named the files but got caught by case-sensitive/case-preserving issues. Fixed (I think) and rebased.

geek:
LGTM, you may want to convert one of the test configs to json5 style, removing quotes, adding comments.

tgross:
> LGTM, you may want to convert one of the test configs to json5 style, removing quotes, adding comments.

Agreed, but I'm going to pick that up in subsequent PRs. (Doing a lot of incremental work here.)

PR #316, opened 2017-04-07
Split-out health check config

tgross:
This PR splits out the `health` config from the `jobs` config, per the design in [RFD86](https://github.com/joyent/rfd/blob/master/rfd/0086/multiprocess.md). One thing that diverges from the RFD is that we were missing the `heartbeat` field for jobs. We _could_ call this `poll` to be consistent with the checks, but I'd rather call it `heartbeat` because a) that's what we're calling it internally, and b) it has a different behavior than the polling.

This PR depends on and includes the work in #312, #314, and #315 so I'll rebase this when those are merged. The only commit in this PR will be 2791b36

cc @misterbisson @cheapRoc @jasonpincin @geek


tgross:
This feature has undergone some internal review and we've decided to cut it, as the semantics of multiple health checks for a service don't really make sense given that we can only report to Consul once.

PR #319, opened 2017-04-10
move health checks back into job definition

tgross:
This PR eliminates the separate `checks` package and moves health checks back under the job definition.

Included:
- changing of terminology `poll` and `frequency` to `interval` in all user-facing configs (I've left these separate internally for clarity when they're side-by-side in the code
- made log entries more consistent between `watches`, `jobs`, and `sensors`.
- `test_tasks` in the integration tests was flaky because the timeout of one of tests was very close to its frequency (now "interval"), and the test has been constrained to eliminate this flakiness. An end-user that makes the timeout and interval very close together will not be able to guarantee a specific number of `restarts` if that field is set. This is an inherent limitation in handling the timeouts and restarts asynchronously, but it's also not really a problem under realistic use. We'll document it though.
- a bunch of test and test fixture cleanup

cc @misterbisson @jasonpincin @geek @cheapRoc


tgross:
I'll push up a PR matching these requirements to RFD86 later today or tomorrow morning.

tgross:
Pushed up https://github.com/joyent/rfd/pull/32

PR #325, opened 2017-04-20
control: Use subtests and temporary file for domain socket

### LINKED ISSUE #306: control socket tests fail on MacOS
A fairly minor item for the v3.0.0 TODO pile: running the `control` package's unit tests fails on MacOS because the control socket directory `/var/run` won't typically be owned by the user. We should have the unit tests open the socket within `/tmp` or somewhere similar.

cc @cheapRoc if you feel like hitting this as you do some of the other control plane work

jen20:
Fixes #306

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,7 @@ vendor/
 
 # fake paths from overlaying volumes in Docker
 /src
+
+# IDE files
+.idea
+*.iml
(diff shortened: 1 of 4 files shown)


tgross:
LGTM. Once TravisCI gives us the 👍  let's merge this.

PR #326, opened 2017-04-20
control: Use github.com/sean-/seed instead of time

jen20:


jen20:
Nothing specifically - in a test we could probably get away with using time as a seed. However, treating test code in the same manner as production code is probably worthwhile, so we should use the correct approach vs the sloppy version.

tgross:
Why not just use https://golang.org/pkg/io/ioutil/#TempFile then?

control/control_test.go
```diff
@@ -10,14 +10,14 @@ import (
...
-	"time"
 
 	"github.com/joyent/containerpilot/tests"
 	"github.com/joyent/containerpilot/tests/assert"
+	"github.com/sean-/seed"
 )
 
 func init() {
-	rand.Seed(time.Now().UTC().UnixNano())
+	seed.MustInit()
```
sean-: Because this is a test, it can be downgraded to `seed.Init()`

cheapRoc:
Normally I'm not crazy about adding dependencies solely for the sake of testing, even if they're supporting best practices... then again, seeding a RNG should never use system time and the library is reviewable in 2 seconds (tiny).

👍 for adding this.

tgross:
I really hate to be a stickler over such a small change. But if we're talking code quality here... then even if we cargo-cult this notion that RNGs should _never_ be seeded with system time, and even if we ignore that this is a non-security-related use of randomness, and even if we ignore that the extra lib we're importing is "left-pad"-scale ceremony around the following: 

```go
func init() {
	n, err := crand.Int(crand.Reader, big.NewInt(math.MaxInt64))
	if err != nil {
		rand.Seed(time.Now().UTC().UnixNano())
		return err
	}
	rand.Seed(n.Int64())
}
```

then we still end up with something that is achieving cryptographic _randomness_ when the goal of a tempfile in this scenario is _collision resistance_. The stdlib's `ioutil.TempFile` give us this already and (bonus!) in a way that is safe to use across multiple processes.

cheapRoc:
Then the RNG line never really belonged in the first accepted PR. I missed that fact until now. I thought that was something being enhanced here because it was just accepted.

cheapRoc:
`ioutil.TempFile` also uses time to seed random generation. If that allows file names to be created via pseudo-randomization than there could be collisions in tests, no? In which case this library could help seed better randomization into those stdlib calls... or I'm completely wrong. I do know I haven't yet seen this happen in Go tests before.

cheapRoc:
Looking at the standard library more closely I'm unsure what this library is adding if `ioutil.TempFile` ensures random names sitting on the file system before returning. I had assumed this was solving for pseudo-randomization but I don't see that happening.

tgross:
One last item: that this library is MIT licensed but claims to have been lifted out of the MPL2.0-licensed Nomad project smells super dubious to me. I know we're probably fine because it's basically trivial, but if it's trivial to that degree then we don't need it as an external dependency either (if it did anything we needed, which it doesn't).

jen20:
Now these have been collapsed into subtests, the race I had attributed to being because of the domain socket file already existing (and us therefore needing a _path_ instead of a _file_) (but actually due to something inside the HTTP server) does not present, so something like this: https://play.golang.org/p/XXdAcxJHyr will work instead. I'll open a new PR in due course to improve this.

Issue #331, opened 2017-04-26
 If the preStart handler returns a non-zero exit code, ContainerPilot will not exit.


ev4ern:
```
preStart
The executable (and its arguments) that will be called immediately prior 
to starting the shimmed application. 
This field is optional. 
If the preStart handler returns a non-zero exit code, ContainerPilot will exit.
```

It doesn`t working in 2.7.2

It worked in 2.4.4

tgross:
Do you have a minimal repro?

tgross:
Nevermind, I've got a repro here. Config file:

```json
{
  "consul": "localhost:8500",
  "preStart": "cat /non-existent-file",
  "logging": {
    "level": "DEBUG",
    "format": "text"
  }
}
```

Version check:
```
$ docker run -it -v "$(pwd)/build/containerpilot:/bin/containerpilot" alpine:latest /bin/containerpilot -version
Version: '2.7.2'
GitHash: 'c0b4faa'
```

Results:
```
$ docker run -it -e "CONTAINERPILOT=file:///test.json" -v "$(pwd)/test.json:/test.json" -v "$(pwd)/build/containerpilot:/bin/containerpilot" alpine:latest /bin/containerpilot sleep 300
DEBU[0000] Loaded config: {"ServiceBackend":{},"LogConfig":{"level":"DEBUG","format":"text","output":"stdout"},"PreStart":{"Name":"preStart","Cmd":null,"Exec":"cat","Args":["/non-existent-file"],"Timeout":"0","TimeoutDuration":0},"PreStop":null,"PostStop":null,"StopTimeout":5,"Coprocesses":null,"Services":[],"Backends":[],"Tasks":null,"Telemetry":null}
DEBU[0000] preStart.RunAndWait start
DEBU[0000] preStart.Cmd.Run
cat: can't open '/non-existent-file': No such file or directory
DEBU[0000] <nil>
DEBU[0000] APP.RunAndWait start
DEBU[0000] APP.Cmd.Run
```

That's definitely a bug and not sure how that got thru the last release. I'll take a look and circle back here ASAP.

tgross:
Ok, that turned out to be pretty easy. We really need to be checking the exit code that comes up [here](https://github.com/joyent/containerpilot/blob/b99f7f866a62094f10d998ddebed84f44e3c5138/commands/commands.go#L98) as well as the error that bubbles up from the stdlib. I just tried this fix:

```
diff --git a/core/app.go b/core/app.go
[...]

Issue #330, opened 2017-04-26
Service does not remain unhealthy


lucj:
I have an API service that depends on a mongodb database.
The service is started once the underlying db is fine.

Below is the logs from compose.

Note: the log message ```debug: -> db.checkConnection: db object not instantiated``` results of the database connection check done every 3 seconds and occurs as the database cannot be reached.

```
$ docker-compose -f docker-compose-cp.yml logs -f api
Attaching to compose_api_1
api_1            | Received event:prestart
api_1            | db is not yet healthly...
api_1            | db is healthly, moving on...
api_1            | 2017/04/26 13:07:59 ==> Starting Consul agent...
api_1            | 2017/04/26 13:07:59 ==> Starting Consul agent RPC...
api_1            | 2017/04/26 13:07:59 ==> Consul agent running!
api_1            | 2017/04/26 13:07:59            Version: 'v0.7.5'
api_1            | 2017/04/26 13:07:59            Node ID: '7ba08ece-d74e-4b39-b5fb-c9b18a396864'
api_1            | 2017/04/26 13:07:59          Node name: 'b53fc98d8a7b'
api_1            | 2017/04/26 13:07:59         Datacenter: 'dc1'
api_1            | 2017/04/26 13:07:59             Server: false (bootstrap: false)
api_1            | 2017/04/26 13:07:59        Client Addr: 127.0.0.1 (HTTP: 8500, HTTPS: -1, DNS: 8600, RPC: 8400)
api_1            | 2017/04/26 13:07:59       Cluster Addr: 172.18.0.13 (LAN: 8301, WAN: 8302)
api_1            | 2017/04/26 13:07:59     Gossip encrypt: false, RPC-TLS: false, TLS-Incoming: false
api_1            | 2017/04/26 13:07:59              Atlas: <disabled>
api_1            | 2017/04/26 13:07:59
api_1            | 2017/04/26 13:07:59 ==> Log data will now stream in as it occurs:
api_1            | 2017/04/26 13:07:59
api_1            | 2017/04/26 13:07:59     2017/04/26 13:07:59 [INFO] serf: EventMemberJoin: b53fc98d8a7b 172.18.0.13
api_1            | 2017/04/26 13:07:59     2017/04/26 13:07:59 [WARN] manager: No servers available
api_1            | 2017/04/26 13:07:59     2017/04/26 13:07:59 [ERR] agent: failed to sync remote state: No known Consul servers
[...]

PR #333, opened 2017-04-26
fix child process exit handling (v3)

tgross:
For #330 and #331 in v3 only. (v2 will be under separate PR.)

There's a bunch of logic we don't need to mess with here because we no longer care about the exact exit code to return back to the container runtime now that we're not "shimming" an application.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -62,6 +62,7 @@ func (c *Command) Run(pctx context.Context, bus *events.EventBus) {
 	c.lock.Lock()
 	log.Debugf("%s.Run start", c.Name)
 	c.setUpCmd()
+	defer reapChildren(c.Cmd.SysProcAttr.Pgid)
 	c.Cmd.Stdout = c.logger
 	c.Cmd.Stderr = c.logger
 
@@ -102,7 +103,7 @@ func (c *Command) Run(pctx context.Context, bus *events.EventBus) {
 		}
 		// blocks this goroutine here; if the context gets cancelled
 		// we'll return from wait() and do all the cleanup
-		if _, err := c.wait(); err != nil {
+		if err := c.wait(); err != nil {
 			log.Errorf("%s exited with error: %v", c.Name, err)
 			bus.Publish(events.Event{events.ExitFailed, c.Name})
 			bus.Publish(events.Event{events.Error, err.Error()})
@@ -127,6 +128,7 @@ func (c *Command) RunAndWaitForOutput(pctx context.Context, bus *events.EventBus
 	c.lock.Lock()
 	log.Debugf("%s.Run start", c.Name)
 	c.setUpCmd()
+	defer reapChildren(c.Cmd.SysProcAttr.Pgid)
 	var (
 		ctx    context.Context
 		cancel context.CancelFunc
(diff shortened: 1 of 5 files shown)


tgross:
I'm now remembering the reason we have all this logic in there: https://github.com/joyent/containerpilot/pull/222. Let me check to make sure we won't have regressed on that bug.

tgross:
Ok, I've corrected the previous work I think... the thing that we want to do here is to make sure that we return the error if there's an error, but _not_ in the case where there's an error from `Wait` but the process itself is successful.

tgross:
Just noting here that I'm not merging this yet because I've finally got a reliable reproduction of the `wait: no child process` problem, which turns out to be an interaction between this code section and our child-reaping code in the `core` package. I'm working up a spike solution out-of-tree and once I've got that in good shape I'll push commits to this branch. Should have it sometime tomorrow.

tgross:
Ok this appears to be working; I'm going to backport it to https://github.com/joyent/containerpilot/pull/334 and then this should be ready to go.

tgross:
Here’s how the problem with the SIGCHLD reaper breaks down:

- We start jobs and wait for them to complete, and then use their exit code in some way (health, preStart, whatever)
- We also have a reaper when we run as PID1
- If the process exits and gets reaped before we get the exit code, we get `wait: no child process`, which blows away any error we get from `wait`, so we can’t tell if it succeeded or not
- If we use the lock around reaping from hashicorp/go-reap, then we get the proper exit codes, but we can’t run the reaper when other jobs are running. This doesn’t work with long-running jobs (and screws up timing of periodic jobs anyways).

The solution is to remove the out-of-band child reaping entirely. We make sure every process we start has a unique process group already, so we can pass that process group ID (negated) to the syscall to `wait4` so that it only cleans up processes with that process group ID. So rather than having a background task running to reap children, we explicitly call it with each group of processes we're done with.

The potential failure mode here is if the child process does setpgid on its own children and then daemonizes them. If this proves to be a problem we can set up a periodic reaper and accept that it will from time to time run into the race condition, but for now I'd prefer we leave it out.

commands/reaper.go
```diff
@@ -0,0 +1,32 @@
...
+			// so we keep trying until there's nothing left
+			var wstatus syscall.WaitStatus
+			pid, err := syscall.Wait4(-pgid, &wstatus, syscall.WNOHANG, nil)
+			switch err {
+			case nil:
+				if pid > 0 {
+					continue
+				}
+				return
+			case syscall.ECHILD:
```
cheapRoc: Does this switch condition need to be here if `default` also covers the same return case?

commands/commands.go
```diff
@@ -62,6 +62,7 @@ func (c *Command) Run(pctx context.Context, bus *events.EventBus) {
 	c.lock.Lock()
 	log.Debugf("%s.Run start", c.Name)
 	c.setUpCmd()
+	defer reapChildren(c.Cmd.SysProcAttr.Pgid)
```
cheapRoc: I think the defer helps to make `reapChildren` place within `Command.Run` easier to grok.

cheapRoc:
I think this turned out great. Added benefit, reaping comes across a bit clearer now that there are explicit conditions, error assertions, and a simpler for loop within the main goroutine.

commands/reaper.go
tgross: Definitely not required, but having the separate case calls out that we're explicitly looking for `ECHILD` and makes it a bit easier to debug.

PR #334, opened 2017-04-26
fix child process exit handling (v2)

tgross:
For #330 and #331 in v2 only. (v3 is in #333)

This adjusts the logic for handling the case we supposedly fixed in #222 where a process that exits before it can be `wait`'d upon will return an error from the stdlib. But we failed to assert that the exit code was actually 0, which means that a process that exits _quickly_ with an error will cause health checks to pass when they should not.

This also includes the check for the exit code described in https://github.com/joyent/containerpilot/issues/331#issuecomment-297501072

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -97,7 +97,6 @@ func RunAndWait(c *Command) (int, error) {
 		if exitErr, ok := err.(*exec.ExitError); ok {
 			if status, ok := exitErr.Sys().(syscall.WaitStatus); ok {
 				if status.ExitStatus() == 0 {
-					log.Debug(err)
 					return 0, nil
 				}
 				return status.ExitStatus(), err
@@ -221,7 +220,6 @@ func (c *Command) waitForTimeout() error {
 		if exitErr, ok := err.(*exec.ExitError); ok {
 			if status, ok := exitErr.Sys().(syscall.WaitStatus); ok {
 				if status.ExitStatus() == 0 {
-					log.Debug(err)
 					return nil // process exited cleanly before we hit wait4
 				}
 				return fmt.Errorf("%s exited with error", c.Name)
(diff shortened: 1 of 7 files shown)


tgross:
Just noting here that I'm not merging this yet because I've finally got a reliable reproduction of the `wait: no child process` problem, which turns out to be an interaction between this code section and our child-reaping code in the `core` package. I'm working up a spike solution out-of-tree and once I've got that in good shape I'll push commits to this branch. Should have it sometime tomorrow.

tgross:
I've backported the fix in https://github.com/joyent/containerpilot/pull/333 to this PR as well. The major difference in the two implementations is that we can't use the read lock for the reaper around the "main" app because otherwise the reaper will never run. Let's see if TravisCI likes this fix and we should be good-to-go.

tgross:
Ok, I've got an explanation for what's been done here in https://github.com/joyent/containerpilot/pull/333#issuecomment-299022241. I've tested this out locally and TravisCI is happy, so I think we're ready to merge this in.

commands/commands.go
```diff
@@ -211,13 +214,18 @@ func (c *Command) waitForTimeout() error {
...
-			log.Debugf(err.Error())
-			return nil // process exited cleanly before we hit wait4
+
+	defer reapChildren(c.Cmd.SysProcAttr.Pgid)
+	err := c.Cmd.Wait()
+	if err != nil {
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			if status, ok := exitErr.Sys().(syscall.WaitStatus); ok {
+				if status.ExitStatus() == 0 {
+					log.Debug(err)
```
cheapRoc: Was it your intention to keep the debug within v2's patch?

tgross: Oops, no. Looks like there a couple places I left a bit of debugging. Will clean that up.

PR #329, opened 2017-04-25
control: Refactor and clean-up for functional reload

cheapRoc:
Opening PR for #328 and continuation of #300.

This PR also implements #336 `PutEnviron` control plane endpoint.

PR #339, opened 2017-05-04
graceful control server reload

tgross:
This PR builds on @cheapRoc's work in https://github.com/joyent/containerpilot/pull/329 to have the control server use the same event bus that the rest of the application is using.

Some other highlights:
- I've removed the SIGHUP handler that this control endpoint replaces
- I've updated the sighup deadlock integration test to use this endpoint
- removed unused bits of the test_probe integration test fixture
- I've added a SIGINT handler which will make standalone integration testing a bit easier

To be done in upcoming PRs:
- the rest of the control plane endpoints
- update the telemetry server to use the event bus too so we can do graceful reloads there


control/control_test.go
```diff
@@ -2,20 +2,30 @@ package control
...
-	"strings"
 	"testing"
 	"time"
 
+	"github.com/joyent/containerpilot/events"
 	"github.com/joyent/containerpilot/tests"
 	"github.com/joyent/containerpilot/tests/assert"
 )
 
+type TestApp struct {
```
cheapRoc: Test plumbing can be removed if its not used. I believe I was going to mock out the App interface and test that the endpoint handlers at least fired.

tgross: Aye, good point. I'll do this and expand test coverage as needed as I start to hit the rest of the endpoints.

PR #346, opened 2017-05-05
refactor control server to allow for JSON response

tgross:
This PR is some prep work to allow for the control server to return a JSON response, which we'll need for the [MaintenanceMode endpoint](https://github.com/joyent/rfd/blob/master/rfd/0086/mariposa.md#maintenancemode-post-v3maintenanceenabledisable). I've also refactored some of the tests so that we have true unit tests of the end points -- I still have a test that smoke-tests the server itself though so we have the same coverage.

cc @cheapRoc


tgross:
The failing test here should be fixed here https://github.com/joyent/containerpilot/pull/345

control/endpoints.go
```diff
@@ -11,70 +11,66 @@ import (
...
+	resp, status := pw(r)
+	switch status {
+	case http.StatusOK:
+		if resp != nil {
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusOK)
+			json.NewEncoder(w).Encode(resp)
+		} else {
+			w.WriteHeader(http.StatusOK)
+			io.WriteString(w, "\n")
```
cheapRoc: I was never quite sure if this `\n` needed to be here or if it satisfied some manual test I was making at one point. It's possible you won't need this.

cheapRoc: I also really like the ability to have general behavior wrapping the individual endpoint funcs but this method will surely go ape-shit in the future. Probably want to watch its growth (this gets into using a third-party library for this kind of stuff....).

control/endpoints.go
```diff
@@ -11,70 +11,66 @@ import (
...
 		os.Setenv(envKey, envValue)
 	}
+	return nil, http.StatusOK
 }
 
 // PostReload handles incoming HTTP POST requests and reloads our current
-// ContainerPilot process configuration. Returns null HTTP response.
-func (e Endpoints) PostReload(w http.ResponseWriter, r *http.Request) {
+// ContainerPilot process configuration.  Returns empty response or HTTP422.
+func (e Endpoints) PostReload(r *http.Request) (interface{}, int) {
```
cheapRoc: `interface{}` are we able to do anything about these to formalize the contract?

control/cover.html
```diff
@@ -0,0 +1,326 @@
+
```
cheapRoc: Was this meant to be checked in?

tgross: Oops! Definitely not!

control/endpoints.go
tgross: Not yet, because I haven't written any of the endpoints that return anything other than nil. We can probably tighten up the interface at that point.

control/endpoints.go
tgross: It's something to watch out for, for sure. In previous web APIs I've done this as composable functions so that I end up with formatter functions (ex. content type renderers) that wrap the switch-on-error that wrap the underlying endpoint. Most of the golang 3rd party routers take this approach.

Issue #332, opened 2017-04-26
TestRunJobPeriodic is flaky


tgross:
Like many of the tests that rely on timing have been, `TestRunJobPeriodic` in the jobs package looks to be flaky. I'll hit this shortly but I want to wrap up the fix for #330 and #331 first. The symptom is that it looks like this, but very infrequently:

```
--- FAIL: TestJobRunPeriodic (0.10s)
        jobs_test.go:147: no events should have timed-out but got [{Startup global} {ExitSuccess myjob} {ExitFailed myjob} {Error myjob: signal: killed} {ExitSuccess myjob} {ExitSuccess myjob} {ExitSuccess myjob} {ExitSuccess myjob} {Stopping myjob} {Stopped myjob}]
FAIL
```

tgross:
This is _still_ flaky: https://travis-ci.org/joyent/containerpilot/builds/232573005

tgross:
Pre-released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-dev.3

Issue #344, opened 2017-05-04
provide a way to flush the event bus


tgross:
Right now we don't have a way to flush the event bus. This results in a lot of ceremony in testing where we have to set up a separate consumer of the event bus and ensure we've pulled _n_ events (whatever we expect). It also means we need to have a buffered channel for closing the event bus rather than having this be a blocking operation that we could use to force a bit more sanity into config reloads.

tgross:
https://github.com/joyent/containerpilot/issues/332 is almost certainly related to this item.

cheapRoc:
I'm sure your PR/diff will demonstrate the idea you had in mind but I'm curious where the flush will change existing API surface area. Also, will this replace or augment all of the event bus/debug subscriber setup work in each test?

tgross:
> I'm sure your PR/diff will demonstrate the idea you had in mind but I'm curious where the flush will change existing API surface area.

I'm still getting it together but essentially `events.Subscriber` will wrap a `sync.WaitGroup` and expose `Add()`, `Done()`, and `Wait()`. This way `Job`, `Watch`, etc. can increment the internal wait group for each event they process and decrement it when done. Tests can call `Wait` after publishing all the test events to force the processing of all of them. (I suspect we'll never call `Wait` in the production code though, only `Close`, but I'm still working that detail out.)

> Also, will this replace or augment all of the event bus/debug subscriber setup work in each test?

Yeah, this would get rid of `DebugSubscriber` entirely which I find to be both clunky and error-prone to use. I'll probably add is a ring buffer of events to the event bus, so that we can say "hey EventBus, what were the last _n_ events you published?"

tgross:
Pre-released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-dev.3

PR #349, opened 2017-05-10
control server MaintenanceMode implementation

tgross:
This PR is currently work-in-progress. Done so far:

- add `/v3/maintenance/{enable|disable}` endpoint
- updated `Job.Status` to include multiple states (Unknown, Healthy, Unhealthy, Maintenance) instead of just a bool up/down.
- removed the now-defunct SIGUSR1 handler

Needs to be done:
- [x] tests for Job maintenance mode behaviors
- [x] tests for the `v3/maintenance/{disable|enable}` endpoints

cc @cheapRoc


tgross:
Also (summarized from my remarks in our internal chat), these endpoints don't include the `updated` field that's in the [RFD](https://github.com/joyent/rfd/blob/master/rfd/0086/mariposa.md#maintenancemode-post-v3maintenanceenabledisable).

The `updated` field would imply that ContainerPilot has a global notion of maintenance mode, which isn’t really the case as each service is marked independently and doing so can fail independently when we talk to Consul. Additionally, setting maintenance mode for each service is async to the POST, so returning a response would be racy. The right thing to do here is for the user to check the (not yet done) `/status` endpoint rather than looking for this response.

So I think we should forget about the response body. There are no current consumers of it anyways -- the `SIGUSR1` was entirely without a response obviously. Getting rid of it now before it goes into the code base will be easier than getting rid of a misleading response that someone is trying to depend upon.

tgross:
I'm happy with where this is. I think the tests can be improved by the event bus improvements I've got planned in #344 (will also fix #332) but these will do for now.

Issue #340, opened 2017-05-04
add subcommands for control plane clients


tgross:
The control plane described in [RFD86](https://github.com/joyent/rfd/blob/master/rfd/0086/mariposa.md) uses a unix domain socket rather than a TCP endpoint. One usability problem with this is that the `curl` that ships with most distros (anything pre 7.40) does not support sockets, and `netcat` is a portability pain because seemingly every distribution has a different version of `netcat` with different command line flags. This isn't a problem for applications like autopilotpattern/mysql or other applications that have a lot of custom code, but it's not a great experience for blueprints that are mostly standalone 3rd-party applications like autopilotpattern/nginx.

We can make this partially less annoying by providing subcommands for the end user, which in turn send HTTP requests to the socket. Here's a proposed syntax. Note these are using the stdlib flags library for golang command line args (i.e. weird plan9-like syntax), just as our `containerpilot -version`, `containerpilot -config`, and `containerpilot -render` options do now.

- `containerpilot -reload`
- `containerpilot -maintenance enable`
- `containerpilot -maintenance disable`
- `containerpilot -putenv 'MY_ENV=my_val'` -putenv 'MY_OTHER_ENV=my_other_val'
- `containerpilot -putmetric 'MY_METRIC=my_val'`

cheapRoc:
I'd actually like to take this if I have time next week/soon.

tgross:
Sure thing! I've got a few implementations using `netcat` (messily) in ce0d0eb and [`reload-containerpilot.sh`](https://github.com/joyent/containerpilot/blob/master/integration_tests/fixtures/app/reload-containerpilot.sh#L5-L6) if you want to draw on that for inspiration outside the unit tests we've got.

tgross:
Pre-released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-dev.3

PR #353, opened 2017-05-12
Add subcommands and client for controlling ContainerPilot

cheapRoc:
Ref: #340. 

This PR introduces both the `subcommands` and `client` packages. 

`client.HTTPClient` provides a simple HTTP client that formalizes requests sent to CP3's new control socket. I've tied this into `subcommands` which map into the CLI flags of `app/core`. 

I believe breaking it out this way could more easily allow any external Go projects to import the client package. This client package can probably be utilized in unit tests as well.

Still **WiP** but I had reload working earlier so I decided to continue on with the remaining subcommands. Tests and lint will fail for the moment.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,12 +1,11 @@
 package client
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
 	"net"
 	"net/http"
-
-	log "github.com/Sirupsen/logrus"
 )
 
 type HTTPClient struct {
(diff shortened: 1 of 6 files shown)


cheapRoc:
**WiP** but I wanted to push up my `core/flags.go` file. This includes `MultiFlag` which allows passing/storing multiple CLI values using Plan9 syntax, such as `-putenv "MY_ENV=first" -putenv "MY_VAL=second"`.

cheapRoc:
So many tiny commits. I'll have this development finalized in the AM.

client/client.go
```diff
@@ -0,0 +1,114 @@
...
+// Reload makes a request to the reload endpoint of a ContainerPilot process.
+func (c HTTPClient) Reload() (*http.Response, error) {
+	resp, err := c.Post("http://control/v3/reload", "application/json", nil)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		mesg := fmt.Sprintf("expected 200 but got %v\n%+v", resp.StatusCode, resp)
```
tgross: This is probably a good use case for the `return nil, fmt.Errorf("...", obj, obj)` idiom. But I'm also wondering if it's the right level of abstraction. We're telling the caller about the HTTP status code when we really care about what that status code means. If we just wanted to return the status code we could do that.

Also, I don't think it's safe to return `*http.Response` here if we've read the `Body` out and closed it.

core/app.go
```diff
@@ -50,28 +51,54 @@ func EmptyApp() *App {
...
+
+		flag.StringVar(&configFlag, "config", "",
+			"file path to JSON5 configuration file.")
+
 		flag.StringVar(&renderFlag, "out", "-",
 			"-(default) for stdout or file path where to save rendered JSON config file.")
-		flag.BoolVar(&versionFlag, "version", false, "Show version identifier and quit.")
+
+		flag.StringVar(&maintFlag, "maintenance", "",
+			"enable/disable maintanence mode through a ContainerPilot process control socket.")
```
tgross: Super nitpick: I realize it's inconsistent in the existing, but can we be consistent with the config option descriptions in terms of first-letter capitalization? I don't have a strong opinion either way but we should pick one and stick to it.

subcommands/subcommands.go
```diff
@@ -0,0 +1,97 @@
...
+type Subcommand struct {
+	client *client.HTTPClient
+}
+
+// Init initializes the configuration of a Subcommand function and the
+// HTTPClient which they utilize for control plane interaction.
+func Init(configFlag string) (*Subcommand, error) {
+	var socketPath = "/var/run/containerpilot.socket"
+
+	if configFlag != "" {
```
tgross: In an effort to be robust we're repeating logic -- the `config.LoadConfig` already handles all the cases we're handling here, so if we wrap that we'll be hiding the errors it returns and that'll make for an inconsistent UI. We can just return the error from `LoadConfig`.

client/client.go
cheapRoc: There was a point in which it made sense to broaden the usability of the client package itself. Looking at it a bit further now, it makes more sense to only return the status code and handle the error message closer to where the client is being implemented. That'll allow the implementer (`subcommands`) to make a decision of how to handle the response.

I feel like the overall return of the response isn't warranted right now either. We're not making use of it and can revisit if we need it in the future.

Also, I had questioned the closing of `resp.Body` but wasn't familiar enough to pick up on the repercussions. Probably the old Rubyist in me.... 🛩

cheapRoc: Also, the user probably only needs us to handle the 500. I can't imagine anything else mattering.

core/app.go
cheapRoc: I'll capitalize them so we stick with what's already there.

subcommands/subcommands.go
cheapRoc: Nice catch... for whatever reason early on I was receiving the `-config` flag as required even when I knew the `cfg.Control.SocketPath` had a default value. As you suggested, I was able to remove the conditional, `socketPath` crap, and only perform....

```go
cfg, err := config.LoadConfig(configFlag)
if err != nil {
        return nil, err
}
```

client/client.go
tgross: There's probably an argument that could be made to return only an `error` object, but until there are consumers it's all handwavy. So returning the status code sounds like the right move now.

cheapRoc: Right, for the moment it's not being used by the current consumer so it doesn't really matter.

cheapRoc: In the past I've questioned whether or not a 500 is a true software error or just an unsuccessful request (but a request that was still successfully made).

cheapRoc: I gotcha now....

```go
if resp.StatusCode != http.StatusInternalServerError {
        return fmt.Errorf("internal error received by control server")
}

return nil
```

cheapRoc: In my recent pass, I'm only returning the single error but making an attempt to properly catch the specific status codes we're returning by the control server/endpoints. So far I've got 422 for `putenv` and `putmetric` and looking for others.

cheapRoc:
Closing in favor of rebased under #356

PR #354, opened 2017-05-15
add DebugEvents and use wait group sub/unsub from EventBus

tgross:
For https://github.com/joyent/containerpilot/issues/344 and https://github.com/joyent/containerpilot/issues/332.  This PR is intended to improve our control over the flow of events and provides us with synchronization points that we need for reliable testing.

- Added a `sync.WaitGroup` around subscribing/unsubscribing from the `EventBus`. This will allow us to force test code to wait until a `Subscriber` has processed all the events we expect it to receive before continuing.
- Changed `Subscriber.Close()` to `Subscriber.Quit()`, which uses the wait group described above to block until we're done.
- Added a ring bugger in the `EventBus` that collects the last 10 events, so that we can read these events back out by calling `DebugEvents()`.
- Eliminated the `TestSubscriber` mock because of all the above.
- Refactored `Job.Run()` to pull the processing of a single Event out from the loop, so that it's easier to understand (and test).
- Refactored the `Job`, `Sensor`, and `Watch` to handle their context being closed in a way that forces all logic thru a single `defer` (this should prevent easy-to-make mistakes in calling `Done()` an extra time in the internal wait group).

cc @cheapRoc


tgross:
I'll rebase on https://github.com/joyent/containerpilot/pull/355 once we're happy with it and that should clean up the failing test

Issue #341, opened 2017-05-04
update telemetry server to use event bus for reloads


tgross:
The telemetry server should use the same structure as the control server for reloading after a configuration update. See https://github.com/joyent/containerpilot/pull/339 for details.

tgross:
Pre-released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-dev.3

PR #358, opened 2017-05-15
update telemetry server to use EventBus for graceful reload

tgross:
For #341 and #166
Updates the Telemetry server to use the EventBus for control, and to use the graceful reload for the HTTP server in golang 1.8. This makes the Telemetry server largely mimic the control socket.

cc @cheapRoc
Note that this PR currently contains all the commits in #354 and I'll have to rebase it once that's merged. The only commit specific to this PR is 5e6b3a9

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -1,12 +1,13 @@
 package telemetry
 
 import (
+	"context"
 	"net"
 	"net/http"
-	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/joyent/containerpilot/events"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
@@ -16,10 +17,11 @@ type Telemetry struct {
 	Sensors   []*Sensor
 	Path      string
 	heartbeat time.Duration
-	mux       *http.ServeMux
-	lock      sync.RWMutex
+	router    *http.ServeMux
 	addr      net.TCPAddr
-	listening bool
+
+	http.Server
+	events.EventHandler // Event handling
 }
 
 // NewTelemetry configures a new prometheus Telemetry server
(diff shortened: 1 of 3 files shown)


telemetry/telemetry.go
```diff
@@ -29,48 +31,80 @@ func NewTelemetry(cfg *Config) *Telemetry {
...
-	}
-	ln, err := net.Listen(t.addr.Network(), t.addr.String())
-	if err != nil {
-		log.Fatalf("Error serving telemetry on %s: %v", t.addr.String(), err)
-	}
-	listener = ln
-	t.listening = true
+// Start starts serving the telemetry service
+func (t *Telemetry) Start() {
+	ln := t.listenWithRetry()
```
cheapRoc: Seeing this twice is starting to make me believe that its not such a terrible hack.

cheapRoc: I remember thinking of the same exact fix but not having the experience to know when or where to implement.

telemetry/telemetry.go
```diff
@@ -29,48 +31,80 @@ func NewTelemetry(cfg *Config) *Telemetry {
...
+		time.Sleep(time.Second)
+	}
+	log.Fatalf("error listening to socket at %s: %v", t.Addr, err)
+	return nil
+}
+
+// Stop shuts down the telemetry service
+func (t *Telemetry) Stop() {
+	log.Debug("telemetry: stopping server")
+	ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
```
cheapRoc: Have you tried removing this timeout? I wonder if we need it anymore....

tgross: We _probably_ don't need it on the Telemetry server (although it can't really hurt either), but I think we want to avoid getting an error message back on the Control server. Otherwise the `/bin/containerpilot -reload` will show errors frequently when not warranted.

telemetry/telemetry.go
tgross: Yah. If we had 3 servers I'd say "let's extract all this out into a server+eventhandler abstraction" but it doesn't seem like it's worth the effort or indirection with just the 2.

PR #359, opened 2017-05-15
increase frequency timer for periodic tasks

tgross:
For #332

The `TestRunJobPeriodicTask` unit test has been flaky, and if I'd thought about it a bit more I would have remembered [my remarks in our own documentation](https://github.com/joyent/containerpilot/tree/abd3070533108a98d559e809d6ce9884a5e7364c/documentation/18-tasks):

> Note on task frequency: Pick a frequency of 1s or longer. Although the task configuration permits frequencies as fast as 1ms, the overhead of spawning a process and its lifecycle is likely to be anywhere from 2ms to 25ms. Your task may not be able to run at all, or it might always be killed before it gets any useful work done.

:blush:

I don't really want a test that takes 4 seconds though, so I'm going to set the timer to 250ms which should realistically be more than enough. This lets us get 2 restarts which will fully exercise our restart logic in 1 second.

cc @cheapRoc
Note that this PR currently contains all the commits in #354 and I'll have to rebase it once that's merged. The only commit specific to this PR is 603fa45


tgross:
Closing in lieu of https://github.com/joyent/containerpilot/pull/361

PR #371, opened 2017-05-18
(wip) V3 docs

tgross:


docs/30-configuration/35-watches.md
```diff
@@ -1 +1,46 @@
...
+  }
+]
+```
+
+The `interval` is the time (in seconds) between polling attempts to Consul. The `name` is the service to query and the `tag` is the optional tag to add to the query.
+
+A watch keeps in memory a list of the healthy IP addresses associated with the service. The list is not persisted to disk and if ContainerPilot is restarted it will need to check back in with the canonical data store which is Consul. If this list changes between polls, the watch emits one or two events:
+
+- A `changed` event is emitted whenever there is a change.
+- A `healthy` event is emitted whenever the watched service becomes healthy. This might mean that the state was previously unknown (as when ContainerPilot first starts up) or that it was previously unhealthy and is now unhealthy. This event will only be fired once for each change in status or count of instances. Subsequent polls that return the same value will not emit the event again.
```
tviaud: There is a typo with "that it was previously unhealthy and is now unhealthy."  It is "healthy"

cheapRoc:
Reviewing this is on my agenda but I was waiting until the first complete draft to read through.

tgross:
@fitzage I've got a build script now for the documentation that'll convert the GitHub markdown to Kirby-compatible format with the appropriate front-matter. Can you check that this looks correct to you? You can check it out by doing the following:

```
git clone git@github.com:joyent/containerpilot.git
git checkout v3_docs
make kirby
```

The output will be in `build/docs` at that point.

docs/10-lifecycle.md
```diff
@@ -0,0 +1,81 @@
+# Lifecycle
+
+ContainerPilot's core functionality is centered around the concepts of jobs, the events that trigger those jobs or that jobs emit, and watches for the external service discovery catalog (Consul).
```
cheapRoc: What about starting this section by introducing the importance of Lifecycles themselves within ContainerPilot? This paragraph seems to introduce Jobs which are brought in as a secondary concept under the next paragraph.

cheapRoc: I guess also you're doing that within the next section as well, intermingled with jobs themselves. Which also makes sense.

fitzage:
@tgross I got these errors, and the docs do not appear to be complete. Maybe there's a typo somewhere, as there appears to be some element of the build looking for an "uild" folder instead of "build."

```
sed: -e: No such file or directory
sed: 1: "build/docs/10-lifecycle ...": undefined label 'uild/docs/10-lifecycle/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/20-design/in ...": undefined label 'uild/docs/20-design/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/31-installation/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/32-configuration-file/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/33-consul/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/34-jobs/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/35-watches/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/36-telemetry/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/37-control-plane/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/38-logging/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/39-examples/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/40-support/i ...": undefined label 'uild/docs/40-support/index.md'
sed: -e: No such file or directory
sed: 1: "build/docs/00-index/ind ...": undefined label 'uild/docs/00-index/index.md'
```

Secondarily, can we name these markdown files whatever we want as part of the build? Currently, they're all named README.md for joyent.com, which was kind of annoying when building the templates, and index.md would work, but I'd rather name them something that makes more semantic sense for the template, like docs.md.

docs/20-design.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+## Why Consul and not etcd or Zookeeper?
+
+Consul provides a number of higher-level capabilities than a simple KV store like etcd or ZK. Providing the ability to use these capabilities would mean either going with a least-common-denominator approach or having complex provider-specific configuration options for tagging interfaces, providing secure connection to the backend, and faster deregistration on shutdown, among others. Additionally, Consul has first-class support for multi-datacenter deployments.
+
+The primary argument for supporting etcd rather than Consul is that Kubernetes and related projects are using it as their service discovery layer. In our discussions with some end users, we haven't found that there's any resistance to the idea that the scheduler's own consensus & membership store doesn't need to be the same store used by applications. And even perhaps _should not_ be the same store, given that in most organizations the team responsible for application development will not be the same team responsible for running the deployment platform.
```
cheapRoc: > we haven't found that there's any resistance to the idea that the scheduler's 
> own consensus & membership store doesn't need to be the same store used
> by applications.

This snippet was a little tough to parse first pass. It's the wording of "haven't", "that there's any", and "doesn't". You're saying that the application's store can differ to the scheduler's, correct? Could this flow better?

docs/20-design.md
```diff
@@ -0,0 +1,59 @@
...
+
+This pattern creates responsibilities on the application that many legacy applications will be unable to fulfill, and ContainerPilot is designed first and foremost with supporting those kinds of applications. Greenfield applications in organizations that already have rich libraries and tooling for handling questions of service discovery, health checking, process supervision, metrics collection, etc. and that have their own service catalog already deployed will be unlikely to get much out of using ContainerPilot.
+
+Much of what ContainerPilot does can be assembled from existing components. One can include in a container an init system and supervisor like [`s6`](http://skarnet.org/software/s6/), which runs Consul agent and [`consul-template`], a [Prometheus](https://prometheus.io/) metrics collection agent, a tool like [registrar](https://gliderlabs.com/registrator/latest/), and a collection of bash scripts to tie all the pieces together. ContainerPilot is an opinionated choice of tooling, but if your organization has strong feelings about those tools then it's possible to implement its behaviors by a sufficiently motivated development team.
```
cheapRoc: > by a sufficiently motivated development team

Keyword, "sufficiently". 😁 

This paragraph sounds like a good blog post, if it wasn't already.

docs/30-configuration/31-installation.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+The above snippet adds the ContainerPilot binary to the container at `/usr/local/bin/containerpilot`. It also specifies the version to install and validates the application fingerprint to make sure that it's installing exactly the version you want.
+
+
+### Building yourself
+
+ContainerPilot is written in [go](https://golang.org/). The makefile at the root of the repository can build either via your local golang toolchain (currently golang 1.8) or in a Docker container.
+
+If you have Docker running, `make build` will build a container image that includes the golang toolchain, downloads and installs all the required libraries into the `vendor/` directory, and builds ContainerPilot. The compiled binary will be found at `build/containerpilot`
```
cheapRoc: Good place to mention `make help`?

README.md
```diff
@@ -22,34 +15,93 @@ To make this work, every application needs to do the following (at a minimum):
...
-- [Periodic tasks](https://www.joyent.com/containerpilot/docs/tasks)
-- [Telemetry](https://www.joyent.com/containerpilot/docs/telemetry)
-- [Coprocesses](https://www.joyent.com/containerpilot/docs/coprocesses)
-- [Frequently asked questions](https://www.joyent.com/containerpilot/docs/faq)
-- [Support](https://www.joyent.com/containerpilot/docs/support)
+ContainerPilot is an init system designed to live inside the container. It acts as a process supervisor, reaps zombies, run health checks, registers the app in the service catalog, watches the service catalog for changes, and runs your user-specified code at events in the lifecycle of the container to make it all work right. ContainerPilot uses Consul to coordinate global state among the application containers.
+
+## Quick Start Guide
+
+Check out our ["Hello, World" application](https://github.com/autopilotpattern/hello-world) on GitHub. Assuming you have Docker and Docker Compose available, it's as easy as:
```
heyawhite: This could be an opportunity to link out to either our docs about Docker / Docker Compose on Triton (especially considering the recent changes with `triton-docker`.

README.md
```diff
@@ -22,34 +15,93 @@ To make this work, every application needs to do the following (at a minimum):
...
+- [Where to file issues](./docs/40-support.md#where-to-file-issues)
+- [Contributing](./docs/40-support.md#contributing)
+- [Backwards compatibility](./docs/40-support.md#backwards-compatibility)
 
 You might also read [our guide building self-operating applications with ContainerPilot](https://www.joyent.com/blog/applications-on-autopilot) and look at the examples below.
 
-### Examples
+## Examples
 
 We've published a number of example applications demonstrating how ContainerPilot works.
```
heyawhite: If any of the below examples have a video, we should highlight that.

docs/10-lifecycle.md
```diff
@@ -0,0 +1,81 @@
+# Lifecycle
```
heyawhite: This is a great place to include a diagram of a lifecycle. 
containerpilot-flow-withprestop

heyawhite: Or possibly at the end. OR, broken up into pieces throughout.

docs/10-lifecycle.md
heyawhite: I agree with that sentiment. Users need to understand lifecycle as a concept first before they can understand jobs/events/etc.

Perhaps:

> ContainerPilot was created to manage the lifecycle of an application. It orchestrates the basic concept of application lifecycle – an object is created, things happen, the object is destroyed – the details differ based on all of the [jobs](#what-is-a-job) and  [events that trigger or are trigged by jobs](#what-is-an-event). All of this is then [watched](#what-is-a-watch) by the external service discovery catalog (Consul).

Each item linked to the relevant section beneath it.

docs/10-lifecycle.md
```diff
@@ -0,0 +1,81 @@
...
+
+ContainerPilot's core functionality is centered around the concepts of jobs, the events that trigger those jobs or that jobs emit, and watches for the external service discovery catalog (Consul).
+
+
+## What is a job?
+
+A ContainerPilot job is a user-defined process and rules for when to execute it, how to health check it, and how to advertise it to Consul. The rules are intended to allow for flexibility to cover nearly any type of process one might want to run. Some possible job configurations include:
```
heyawhite: Though "advertise" is technically correct, I recommend instead, "alert": 

> A ContainerPilot job is a user-defined process and rules for when to execute it, how to health check it, and how to alert Consul about the results of the process.

docs/10-lifecycle.md
```diff
@@ -0,0 +1,81 @@
...
+
+An event is a message about a change in the state of the container and its jobs. Every ContainerPilot job receives all the events as an asynchronous but ordered stream of messages. The events include the start of the container, a change in the health of a job, the exit (successful or not) of each job, a change to the state of a `watch`, the expiration of timeouts, etc. (A full listing with detailed explanation of each is available in the [job configuration](./30-configuration.md/34-jobs.md) section.)
+
+Events are internal to a single instance of ContainerPilot; they are not shared to other containers. The only way the information about a job's health can be seen by other containers is if a job has been configured for service discovery in Consul and the other containers have a `watch` for that information.
+
+
+## What is a watch?
+
+A watch is a configuration of a service to watch in Consul. The watch monitors the state of the service and emits events when the service becomes healthy, becomes unhealthy, or has a change in the number of instances. Note that a watch does not include a behavior; watches only emit the event so that jobs can consume that event.
```
heyawhite: > A watch is a configuration of a service to watch in Consul. 

Instead of using "watch" twice, use "monitor" the second time. That way there is clarity that continues on with the next sentence as to what is the `watch` and what watch does.

docs/10-lifecycle.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+## How do events trigger jobs?
+
+By default, a job's `exec` process starts as soon as ContainerPilot has finished startup. Many jobs will want to have a configuration that determines some specific event to wait for, using the `when` field. This field expects a `source` (the name of the event source) and either `once` (trigger one-time only) or `each` (trigger each time this event happens).
+
+For example, if we want a job to run only once `myDb` is healthy (but not each time its health changes), we might use the following configuration:
+
+```json5
+when: {
```
heyawhite: is `when` the event? Is it the whole bracketed piece of information?

docs/20-design.md
```diff
@@ -0,0 +1,59 @@
+# Design: the Why of ContainerPilot
+
+## Why active service discovery?
+
+The goal of service discovery in cloud-native applications is to configure dependencies for our applications and update them as changes occur without manual intervention. There are two major approaches to solving this problem:
+
+- Insert a proxy between the application and the dependency. The proxy knows which dependency instances are available and implements load balancing among them.
+- Add dependency load balancing and dynamic configuration to the application itself.
+
+The first option is a common approach in the container ecosystem. It represents all the right reasons to outsource a problem: we get quick scalability with no need to modify the existing application. The proxy is a centralized place to manage access to dependencies, so all we have to do is point the client application at the proxy and go. The proxy itself can be as simple as HAProxy, or it can be a more complex proxy that can direct writes to a master and reads to the read replicas. This doesn't entirely solve the discovery problem but at least it centralizes the solution.
```
heyawhite: > it can be a more complex proxy that can direct writes to a master and reads to the read replicas

Master application?

docs/20-design.md
```diff
@@ -0,0 +1,59 @@
...
+
+The difference between the two options is a matter of where decisions are made. Passive discovery patterns are those that separate the application from the decisions, leaving the application passive in both the choice of what back ends to connect to, and passive in resolving failures that may result. Active discovery patterns move those decisions into the application so it can have an active role in choosing the backend and working around failures it may encounter. By making the application an active participant in the discovery process, we can eliminate a layer of complexity, misdirection, and latency between the application and its backends and give us faster, more reliable, and more resilient applications.
```
heyawhite: You make an excellent point, and therefore more attention should be drawn to it. Break it up:

> The difference between the two options is a matter of where decisions are made. Passive discovery patterns are those that separate the application from the decisions, leaving the application passive in both the choice of what back ends to connect to, and passive in resolving failures that may result. Active discovery patterns move those decisions into the application so it can have an active role in choosing the backend and working around failures it may encounter.
>
> By making the application an active participant in the discovery process, we can eliminate a layer of complexity, misdirection, and latency between the application and its backends and give us faster, more reliable, and more resilient applications.

docs/20-design.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+## Why Consul and not etcd or Zookeeper?
+
+Consul provides a number of higher-level capabilities than a simple KV store like etcd or ZK. Providing the ability to use these capabilities would mean either going with a least-common-denominator approach or having complex provider-specific configuration options for tagging interfaces, providing secure connection to the backend, and faster deregistration on shutdown, among others. Additionally, Consul has first-class support for multi-datacenter deployments.
+
+The primary argument for supporting etcd rather than Consul is that Kubernetes and related projects are using it as their service discovery layer. In our discussions with some end users, we haven't found that there's any resistance to the idea that the scheduler's own consensus & membership store doesn't need to be the same store used by applications. And even perhaps _should not_ be the same store, given that in most organizations the team responsible for application development will not be the same team responsible for running the deployment platform.
+
+
+## Why are jobs not the same as services?
```
heyawhite: This is a missed opportunity. Instead, consider writing a section under the header "What's the difference between jobs and services?" so you can define what services are before explaining why jobs are not services. For the definition of jobs, you can direct back to the lifecycle page.

docs/20-design.md
```diff
@@ -0,0 +1,59 @@
...
+
+A job is the core abstraction for managing all processes in ContainerPilot. But the end user will not necessarily want to advertise all processes in a container to Consul. Health checks, sensors, setup tasks, etc. are all processes that a container needs to run that are internal to the container.
+
+
+## Why don't watches or metrics have behaviors?
+
+By having ContainerPilot's behaviors all in jobs rather than in watches or metrics, you can configure more than one behavior when an event is emitted.
```
heyawhite: First you need to define what a behavior is, or give examples.

docs/20-design.md
```diff
@@ -0,0 +1,59 @@
...
+
+For watches, ContainerPilot provides only the `changed`, `healthy`, and/or `unhealthy` events. The application developer can decide whether to coalesce multiple concurrent events into a single event, whether to start a job and then wait for that job to complete before starting a different one, or whatever they require for their application.
+
+Likewise, for metrics, a single job might take a measurement from the application environment but write several metrics after parsing that measurement. A good example of this is Nginx's `stub_status` module, which provides several numerical measurements to a single HTTP GET. A job designed as a sensor might take this result, do some math on some of the numbers, and then execute `containerpilot -putmetric` multiple times.
+
+
+## Why should you not use ContainerPilot?
```
heyawhite: What about repositioning this to "When"?

docs/30-configuration/32-configuration-file.md
```diff
@@ -0,0 +1,275 @@
+# Configuration file
+
+ContainerPilot expects a configuration file with details about what services it should register, how to check their health, and what to do at startup and shutdown, among others. There are two ways to specify the ContainerPilot configuration location:
+
+1. An environment variable `CONTAINERPILOT`, pointing to the file location in the container.
+2. As an argument to ContainerPilot via the `-config` flag.
+
+##### Examples: specifying the configuration file path
```
heyawhite: Add on "in the Dockerfile" for those who missed the installation.

> ##### Examples: specifying the configuration file path in the Dockerfile

README.md
```diff
@@ -22,34 +15,93 @@ To make this work, every application needs to do the following (at a minimum):
...
+
+This application blueprint demonstrates using ContainerPilot to update Nginx upstream configuration at runtime. Try scaling up via `docker-compose scale hello=2 world=3` to see the Nginx configuration updated.
+
+You can also [download](https://github.com/joyent/containerpilot/releases) the latest release of ContainerPilot from GitHub.
+
+## Documentation
+
+Documentation for ContainerPilot and where it fits with the rest of the Triton ecosystem can be found at [www.joyent.com/containerpilot](https://www.joyent.com/containerpilot). The index below links to the documentation in this repo for convenience.
+
+[Lifecycle](./docs/10-lifecycle.md)
```
heyawhite: Why does this document exist outside of the configuration docs? Some of the information appears to be repeated.

docs/30-configuration/35-watches.md
```diff
@@ -0,0 +1,46 @@
...
+
+- A `changed` event is emitted whenever there is a change.
+- A `healthy` event is emitted whenever the watched service becomes healthy. This might mean that the state was previously unknown (as when ContainerPilot first starts up) or that it was previously unhealthy and is now healthy. This event will only be fired once for each change in status or count of instances. Subsequent polls that return the same value will not emit the event again.
+- A `unhealthy` event is emitted whenever the watched service becomes unhealthy. This might mean that the service is not yet running when we first poll, or that it was previously healthy and is now unhealthy. This event will only be fired once for each change of status. Subsequent polls that return the same value will not emit the event again.
+
+The name of the events emitted by watches are namespaced so as not to collide with internal job names. These events are prefixed by `watch.`. Here is an example configuration for a job listening for a watch event:
```
heyawhite: You've got a period in `watch.`

docs/30-configuration/35-watches.md
```diff
@@ -0,0 +1,46 @@
+# Watches
+
+A watch is a configuration of a service to watch in Consul. The watch monitors the state of the service and emits events when the service becomes healthy, becomes unhealthy, or has a change in the number of instances. Note that a watch does not include a behavior; watches only emit the event so that jobs can consume that event.
```
heyawhite: > A watch is a configuration of a service to monitor in Consul. 

Or 

> A `watch` is a configuration of a service to watch in Consul. 

Or possibly a combination of the two.

docs/10-lifecycle.md
cheapRoc: Realizing that [advertise](https://www.consul.io/docs/agent/options.html#_advertise) is Consul parlance.... I like @heyawhite's suggestion.

README.md
tgross: The intent is that the Lifecycle section is the high-level view of only the core concepts work, whereas the Configuration section is the nitty-gritty. I'm also making the assumption that people don't read the whole documentation (alas!) so redundancy is ok.

docs/10-lifecycle.md
tgross: Yeah for sure. The lifecycle is all new though so we need a new diagram.

docs/10-lifecycle.md
tgross: Agreed on all this. 👍

docs/10-lifecycle.md
tgross: I have two problems with using the term "alert" here:
- It's backwards from how we're using it. The `job`'s service definition is what registers that the service exists. If anything could be described as an "alert" it's the `watch`.
- "Alert" is a term of art in operations that I don't want to overload.

How about "register" in Consul? This matches up with other projects as well as Consul's own APIs.

docs/10-lifecycle.md
tgross: Good call.

docs/10-lifecycle.md
tgross: The whole bracketed thing is an event, but we don't call it `event` because it might run on something other than an event (ex. "every 5 seconds").

docs/20-design.md
tgross: > that can direct writes to a master and reads to the read replicas

Oops, should be "...that can direct writes to a primary and reads to the read replicas." Using "master" in that scenario is considered outdated language.

docs/20-design.md
tgross: Ok!

docs/20-design.md
tgross: It definitely could. I cribbed it from the RFD where it probably flowed better in context and the whole concept of end user popularity mattered more. How about:

> The primary argument for supporting etcd rather than Consul is that Kubernetes and related projects are using it as their service discovery layer. But there's no particular reason that the scheduler's own consensus and membership store needs to be the same store used by applications. And even perhaps _should not_ be the same store, given that in most organizations the team responsible for application development will not be the same team responsible for running the deployment platform.

docs/20-design.md
tgross: Yeah, or eliminate the term "behavior" entirely as its not used by any of the configuration. It's an `exec` process.

docs/20-design.md
tgross: But then it doesn't start with "why?" 😀 

But agreed. Something needs to be done here to explain what a service is first.

docs/30-configuration/31-installation.md
tgross: Sure.

docs/30-configuration/32-configuration-file.md
tgross: Good call.

docs/30-configuration/35-watches.md
tgross: The first would make the second sentence partially redundant. I'll go with the second.

docs/30-configuration/35-watches.md
tgross: Oops. Thanks.

tgross:
@fitzage wrote:
> I got these errors, and the docs do not appear to be complete. Maybe there's a typo somewhere, as there appears to be some element of the build looking for an "uild" folder instead of "build."

Ugh, I was _really_ hoping I was able to write that `sed` portably but it looks like no. I'll have to take another crack at that.

@fitzage wrote:
> Secondarily, can we name these markdown files whatever we want as part of the build? Currently, they're all named README.md for joyent.com, which was kind of annoying when building the templates, and index.md would work, but I'd rather name them something that makes more semantic sense for the template, like docs.md.

Absolutely! Consider it done!

docs/20-design.md
cheapRoc: Glad you pasted that. It's much easier to read...

> But there's no particular reason that the scheduler's own consensus and membership store needs to be the same store used by applications.

Also, you can include my anecdotal usage of DC/OS as further proof that a secondary store is REQUIRED especially when working within an opinionated framework/PaaS. I figured that's what you meant first go around, so I knew the wording was off when I first questioned it.

README.md
tgross: I'd rather not directly link the two in the docs, especially in the quick start guide where I want the minimum possible "this works".

docs/10-lifecycle.md
tgross: I'm going to do any diagrams under a separate PR though. Not really 100% necessary for launch given that the existing docs don't have diagrams.

docs/30-configuration/35-watches.md
tgross: Fixed.

docs/30-configuration/35-watches.md
tgross: Went with a slight variant:

> A `watch` is a configuration of a service to monitor in Consul. The watch polls the state of the service and emits events when the service becomes healthy, becomes unhealthy, or has a change in the number of instances.

docs/30-configuration/32-configuration-file.md
tgross: Done.

docs/20-design.md
tgross: Rephrased as:

> Why use something other than ContainerPilot?

so that I can keep this all as "why" questions.

docs/20-design.md
tgross: Updated to:

> ## Why don't watches or metrics have an exec field?

> By having ContainerPilot's `exec` fields all in jobs rather than in watches or metrics, you can configure more than one `exec` behavior when an event is emitted.

docs/20-design.md
tgross: Updated to:

> A job is the core abstraction for managing all processes in ContainerPilot. A service is a job that's been registered with Consul. But the end user will not necessarily want to register all processes in a container.

docs/20-design.md
tgross: Done

docs/20-design.md
tgross: Fixed

docs/10-lifecycle.md
tgross: Done.

docs/10-lifecycle.md
tgross: Updated to:

> A ContainerPilot job is a user-defined process and rules for when to execute it, how to health check it, and how to register it as a service in Consul.

docs/30-configuration/31-installation.md
tgross: Done.

tgross:
@fitzage I think I have the build script fixed now. I've also changed the files output to be `docs.md` instead of `index.md`

tgross:
I've addressed the commentary. At this point I think it's better to merge this to master. This way we can make further incremental improvements on it while the CPv3 release candidate is out and being looked at, rather than trying to do it all as a big-bang commit.

Issue #373, opened 2017-05-19
drop exec from sensors


tgross:
After seeing @misterbisson working on improving the sensor config for our v2 Nginx blueprint (ref https://github.com/autopilotpattern/nginx/pull/49), and then working on the docs PR #371, I'm thinking we should drop the `exec` field from the `sensors` configuration.

Instead the `sensors` would only configure the Prometheus metrics that we _expect_ to receive. And then it would be up to a `job` running periodically to record those metrics. This would have a couple of advantages:

- A single job can record multiple metrics in a single execution by passing multiple `-putmetric` flags to ContainerPilot (or multiple HTTP POSTs to the control socket). This would reduce overhead in cases where there are multiple metrics on the same application, and could help eliminate mistakes like https://github.com/autopilotpattern/nginx/issues/48#issuecomment-301589006.
- We wouldn't have to make an awkward documentation case for why `watches` don't have their own exec but `sensors` do.
- It would emphasize the core abstraction of jobs + events.
- Sensors would no longer be a special case of an exec, reducing our surface area for ContainerPilot bugs.

Thoughts @misterbisson or @cheapRoc? Internally the change would be totally trivial. But this would be a breaking config change, so we'd want to get it into v3.0.0 rather than waiting until a later v3.x.

misterbisson:
This makes sense.

We also should consider that Prom exporters for some apps are likely a better choice than our implementation. Consider:

https://github.com/knyar/nginx-lua-prometheus/blob/master/README.md

https://github.com/prometheus/mysqld_exporter

I don't think there's anything more we need to do to make those easier to run with ContainerPilot, though we might consider changing the default port to leave room. 

There's also an argument that ContainerPilot should expose more of its internal metrics, as the Prometheus app does.

tgross:
> We also should consider that Prom exporters for some apps are likely a better choice than our implementation.

Sure, although that's pretty much the case with every ContainerPilot feature. Most of what we're doing can be done more cleanly by a dedicated implementation, and ContainerPilot serves more as a "how do I workaround my lack of that dedicated implementation?" (ex. legacy lift-and-shift applications).

> There's also an argument that ContainerPilot should expose more of its internal metrics, as the Prometheus app does.

Agreed. I have some TODOs sprinkled around the code for figuring out just what those might be other than the golang runtime info (which we already expose).

cheapRoc:
[...]

PR #375, opened 2017-05-19
remove exec from sensors

tgross:
For https://github.com/joyent/containerpilot/issues/373. This PR renames the `sensors` field to `metrics` and drop the `exec` field and associated code as described in the issue:

> Instead the `sensors` would only configure the Prometheus metrics that we _expect_ to receive. And then it would be up to a `job` running periodically to record those metrics. This would have a couple of advantages:

> - A single job can record multiple metrics in a single execution by passing multiple `-putmetric` flags to ContainerPilot (or multiple HTTP POSTs to the control socket). This would reduce overhead in cases where there are multiple metrics on the same application, and could help eliminate mistakes like https://github.com/autopilotpattern/nginx/issues/48#issuecomment-301589006.
> - We wouldn't have to make an awkward documentation case for why `watches` don't have their own exec but `sensors` do.
> - It would emphasize the core abstraction of jobs + events.
> - Sensors would no longer be a special case of an exec, reducing our surface area for ContainerPilot bugs.


tgross:
cc @cheapRoc @misterbisson

Issue #372, opened 2017-05-18
test failure in test_discovery_consul


tgross:
Looks like there's a flaky code path being exercised by test_discovery_consul (ref [build 233791212](https://travis-ci.org/joyent/containerpilot/builds/233791212)):

```
2017/05/18 21:00:55 TEST: test_discovery_consul
Creating network "testdiscoveryconsul_default" with the default driver
Creating testdiscoveryconsul_consul_1
testdiscoveryconsul_consul_1 is up-to-date
Creating testdiscoveryconsul_app_1
Creating testdiscoveryconsul_nginx_1
2017/05/18 21:01:05 /app returned 404 Not Found
```

This build was in the docs PR, so there weren't any code changes. We haven't seen this fail in a couple dozen builds, so it must be something with a narrow failure window. ☹️ 

logs:
<details>
----- APP LOGS ------
==> Starting Consul agent...
==> Consul agent running!
           Version: 'v0.8.3'
           Node ID: '5db8d743-f6f6-f5ec-20a1-00708258649b'
         Node name: 'consul'
        Datacenter: 'dc1'
            Server: true (bootstrap: false)
       Client Addr: 0.0.0.0 (HTTP: 8500, HTTPS: -1, DNS: 8600)
      Cluster Addr: 172.18.0.2 (LAN: 8301, WAN: 8302)
    Gossip encrypt: false, RPC-TLS: false, TLS-Incoming: false
             Atlas: <disabled>
==> Log data will now stream in as it occurs:
    2017/05/18 21:00:57 [DEBUG] Using unique ID "5db8d743-f6f6-f5ec-20a1-00708258649b" from host as node ID
    2017/05/18 21:00:57 [INFO] raft: Initial configuration (index=1): [{Suffrage:Voter ID:172.18.0.2:8300 Address:172.18.0.2:8300}]
    2017/05/18 21:00:57 [INFO] raft: Node at 172.18.0.2:8300 [Follower] entering Follower state (Leader: "")
    2017/05/18 21:00:57 [INFO] serf: EventMemberJoin: consul 172.18.0.2
    2017/05/18 21:00:57 [INFO] consul: Adding LAN server consul (Addr: tcp/172.18.0.2:8300) (DC: dc1)
    2017/05/18 21:00:57 [INFO] serf: EventMemberJoin: consul.dc1 172.18.0.2
    2017/05/18 21:00:57 [INFO] consul: Handled member-join event for server "consul.dc1" in area "wan"
[...]

Issue #317, opened 2017-04-07
clean up consul configuration


tgross:
In the [`discovery`](https://github.com/joyent/containerpilot/tree/master/discovery) package we have a configuration hook in the `init` that was designed to allow for multiple possible discovery backends. As of #281 we no longer need this little bit of complexity so it'd be nice to tidy up the internals.

tgross:
Merged in https://github.com/joyent/containerpilot/issues/383

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-RC1

PR #378, opened 2017-05-22
protect map of last Consul check with lock, improve testability

tgross:
While trying to figure out #372, I discovered that out check of Consul for changes was subject to a data race that could cause a panic in go1.8+ builds (like v3). I've never seen this in the wild and it would be unlikely to hit, but we should definitely fix it anyways. I've also improved the testability of that code path so that I'm more confident in it and don't go down the rabbit hole on it as I did in #372.

cc @cheapRoc

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -188,7 +188,7 @@ func (c *Consul) CheckForUpstreamChanges(backendName, backendTag string) (didCha
 	}
 	isHealthy = len(instances) > 0
 	didChange = c.compareAndSwap(backendName, instances)
-	return
+	return didChange, isHealthy
 }
 
 // returns true if any addresses for the service changed and updates
@@ -203,15 +203,15 @@ func (c *Consul) compareAndSwap(service string, new []*consul.ServiceEntry) bool
 
 // Compare the two arrays to see if the address or port has changed
 // or if we've added or removed entries.
-func compareForChange(existing, new []*consul.ServiceEntry) (changed bool) {
-	if len(existing) != len(new) {
+func compareForChange(existing, newEntries []*consul.ServiceEntry) (changed bool) {
+	if len(existing) != len(newEntries) {
 		return true
 	}
 	sort.Sort(ByServiceID(existing))
-	sort.Sort(ByServiceID(new))
+	sort.Sort(ByServiceID(newEntries))
 	for i, ex := range existing {
-		if ex.Service.Address != new[i].Service.Address ||
-			ex.Service.Port != new[i].Service.Port {
+		if ex.Service.Address != newEntries[i].Service.Address ||
+			ex.Service.Port != newEntries[i].Service.Port {
 			return true
 		}
 	}
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -177,35 +177,36 @@ func (c *Consul) registerCheck(service discovery.ServiceDefinition) error {
 	)
 }
 
-// CheckForUpstreamChanges runs the health check
+// CheckForUpstreamChanges requests the set of healthy instances of a
+// service from Consul and checks whether there has been a change since
+// the last check.
 func (c *Consul) CheckForUpstreamChanges(backendName, backendTag string) (didChange, isHealthy bool) {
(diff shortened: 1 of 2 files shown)


discovery/consul/consul.go
```diff
@@ -50,7 +53,9 @@ func NewConsulConfig(config interface{}) (*Consul, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &Consul{*client, false}, nil
+	upstreams := make(map[string][]*consul.ServiceEntry)
+	consul := &Consul{*client, false, sync.RWMutex{}, upstreams}
```
cheapRoc: Another nit... Shouldn't we unravel this a bit for the sake of clarity like we do within `configFromMap`. 

Something like...
```go
return &Consul{
	*client,
	wasRegistered: false,
	lock:          syncRWMutex{},
	upstreams:     upstreams,
}, nil
```
The reason I'm bringing it up is because I think the linter caught me doing this one time.

discovery/consul/consul.go
```diff
@@ -20,6 +21,8 @@ func init() {
 type Consul struct {
```
cheapRoc:
```go
// Consul is a service discovery backend for Hashicorp Consul
type Consul struct
```

Nit but shouldn't we provide a more specific description of what this struct represents within the context of ContainerPilot itself? This is a client wrapper of a tool specific to a set of needs and not necessarily the tool itself.

Also, is the line number wrong on this? ::shrug:: Weird formatting on GH.

discovery/consul/consul.go
```diff
@@ -172,35 +177,36 @@ func (c *Consul) registerCheck(service discovery.ServiceDefinition) error {
...
-	didChange = compareForChange(upstreams[backendName], services)
-	if didChange || len(services) == 0 {
-		// We don't want to cause an onChange event the first time we read-in
-		// but we do want to make sure we've written the key for this map
-		upstreams[backendName] = services
-	}
-	return didChange, isHealthy
+	isHealthy = len(instances) > 0
+	didChange = c.compareAndSwap(backendName, instances)
+	return
```
cheapRoc: Is this return fulfilling `(didChange, isHealthy bool)`? I see it correct in the old version.

discovery/consul/consul.go
```diff
@@ -172,35 +177,36 @@ func (c *Consul) registerCheck(service discovery.ServiceDefinition) error {
...
+}
+
+// returns true if any addresses for the service changed and updates
+// the internal state
+func (c *Consul) compareAndSwap(service string, new []*consul.ServiceEntry) bool {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	existing := c.upstreams[service]
+	c.upstreams[service] = new
```
cheapRoc: Can we name this something besides `new`, maybe `newEntry`? Surprised this made it past compilation checks.

discovery/consul/consul.go
cheapRoc: Actually looks like the line number worked out.

tgross: Agreed on this but I'm going to cover that under https://github.com/joyent/containerpilot/issues/317.

discovery/consul/consul.go
tgross: I thought so to but apparently you can't mix the literal `*client` and the field names. The linter gives this a pass. I think we could do:

```go
consul := &Consul{*client}
consul.wasRegistered = false
consul.lock = sync.RWMutex{}
consul.upstreams = upstreams

return consul
```

discovery/consul/consul.go
tgross: Good catch.

discovery/consul/consul.go
tgross: Yes, it's an implicit return that works because the return parameters are named. But maybe it'd be clearer to be explicit.

cheapRoc: Ahhhh... implicit.... I'd rather the explicit return but this is also something I just need to get used to.

discovery/consul/consul.go
tgross: Oh, turns out you can't do that either:

> too few values in struct initializer

Oh well.

tgross:
Updated with the changes.

Issue #365, opened 2017-05-16
test_envvars is flaky


tgross:
The integration test for `test_envvars` is racy -- we're testing that the environment variables are set via a job exec before they might be set. So occasionally we get test results like this:

```
2017/05/16 15:02:38 TEST: test_envvars
2017/05/16 15:02:39 control: serving at /var/run/containerpilot.socket
2017/05/16 15:02:39 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
2017/05/16 15:02:39 FAIL: test_envvars
```

I believe this is a bug in the test rig rather than in the setting of the env vars themselves, but I need to look into it.

tgross:
Hm... the environment variable is being set in [`NewApp`](https://github.com/joyent/containerpilot/blob/bf56b79d6d4b287072905d15e4549606e934c601/core/app.go#L193), but this is entirely before we call `Run` and I was pretty sure this is all synchronous code up till this point.

tgross:
In [build 232519697](https://travis-ci.org/joyent/containerpilot/builds/232519697) we have an example of the failure:

```
2017/05/15 17:32:31 TEST: test_envvars
2017/05/15 17:32:33 control: serving at /var/run/containerpilot.socket
2017/05/15 17:32:33 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
2017/05/15 17:32:33 HOSTNAME=8e988f452852
2017/05/15 17:32:33 TERM=xterm
2017/05/15 17:32:33 NPM_CONFIG_LOGLEVEL=info
2017/05/15 17:32:33 NODE_VERSION=7.10.0
2017/05/15 17:32:33 FAIL: test_envvars
```

What's interesting here is that we don't see the `CONTAINERPILOT_PID` variable either, which is set before we even read our configuration (ref [app.go#L147](https://github.com/joyent/containerpilot/blob/bf56b79d6d4b287072905d15e4549606e934c601/core/app.go#L147))

tgross:
Got a clue on this one. Note that the full output is:

```
CONTAINERPILOT_BIN=./containerpilot docker-compose run test
2017/05/22 20:10:46 control: serving at /var/run/containerpilot.socket
2017/05/22 20:10:46 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
[...]

PR #379, opened 2017-05-22
avoid dropping job logs by moving log Close() to after unsubscribe

tgross:
For https://github.com/joyent/containerpilot/issues/365

For short-lived containers, logs are getting cut off before we can log them to stdout. The log close happens after the event bus is sent the "unsubscribe" message for the job, and that means the event bus can potentially halt ContainerPilot before we've closed the logs. By moving the `CloseLogs` until after this, we'll make sure we've logged everything we can.

I've also moved the step where we deregister from Consul for the same reason.

cc @misterbisson @cheapRoc

### DIFF
--- a/jobs/jobs.go
+++ b/jobs/jobs.go
@@ -314,10 +314,10 @@ func (job *Job) cleanup(ctx context.Context, cancel context.CancelFunc) {
 		}
 	}
 	cancel()
-	job.Unsubscribe(job.Bus) // deregister from events
-	job.Deregister()         // deregister from Consul
-	job.Bus.Publish(events.Event{Code: events.Stopped, Source: job.Name})
 	job.exec.CloseLogs()
+	job.Deregister()         // deregister from Consul
+	job.Unsubscribe(job.Bus) // deregister from events
+	job.Bus.Publish(events.Event{Code: events.Stopped, Source: job.Name})
 }
 
 // String implements the stdlib fmt.Stringer interface for pretty-printing
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -143,6 +143,8 @@ func (c *Command) Kill() {
 
 // CloseLogs safely closes the io.WriteCloser we're using to pipe logs
 func (c *Command) CloseLogs() {
+	// need to nil check these because they might have been closed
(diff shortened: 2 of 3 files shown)


jobs/jobs.go
```diff
@@ -314,10 +314,10 @@ func (job *Job) cleanup(ctx context.Context, cancel context.CancelFunc) {
 		}
 	}
 	cancel()
-	job.Unsubscribe(job.Bus) // deregister from events
+	job.exec.CloseLogs()
```
cheapRoc: Double checking this method and I noticed the following...

```go
func (c *Command) CloseLogs() {
	if c != nil && c.logger != nil {
```

Does the `c != nil` need to be there?

tgross: Hm, that really doesn't seem to do anything does it? If `c == nil` then we'd get a NPE when we called `CloseLogs` anyways. I wonder if it found its way into the function during a refactor. Will fix.

tgross: Done.

tgross:
Oh, well looks like that bit of code _was_ there for a reason:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x6faad2]
goroutine 47 [running]:
github.com/joyent/containerpilot/commands.(*Command).CloseLogs(0x0)
	/go/src/github.com/joyent/containerpilot/commands/commands.go:146 +0x22
github.com/joyent/containerpilot/jobs.(*Job).cleanup(0xc420096620, 0xa5e0a0, 0xc42006ce80, 0xc42015caa0)
	/go/src/github.com/joyent/containerpilot/jobs/jobs.go:317 +0x221
github.com/joyent/containerpilot/jobs.(*Job).Run.func1(0xc420096620, 0xa5e0a0, 0xc42006ce80, 0xc42015caa0)
	/go/src/github.com/joyent/containerpilot/jobs/jobs.go:194 +0x1aa
created by github.com/joyent/containerpilot/jobs.(*Job).Run
	/go/src/github.com/joyent/containerpilot/jobs/jobs.go:200 +0x136
```

Issue #370, opened 2017-05-17
more template functions


ahelal:
Hey, 

Thank you for container pilot. I would like to make a PR to give a little bit more power to the go templating, but would first want to get some feedback.

I want the backend to be dynamic. We use client side load balancing. Each container is packaged with HAProxy in our base image. And I want the images based on our base image, can configure the background vie environments. It is really hard to achieve that with limited template functionality.

For example exposing [env](https://github.com/hashicorp/consul-template/blob/8494b8b81df7c03a0503e08814e7ce24c576ee25/template/template.go#L217)  or [loop](https://github.com/hashicorp/consul-template/blob/8494b8b81df7c03a0503e08814e7ce24c576ee25/template/template.go#L221) like consul-template or others

Regards

tgross:
Can you provide an example of what this would look like?

ahelal:
@tgross thank you for your reply. 

docker file example
```
from XXXXXXX
ENV  SERVICE_NAME_1="pilot_redis"
ENV  SERVICE_PORT_1=6379
ENV  SERVICE_NAME_2="pilot_mysql"
ENV  SERVICE_PORT_2=3306
COPY ./app.py /app.py
CMD [ "/bin/containerpilot", "-config", "file://etc/containerpilot.json","/app.py"]
```

sample container.json
```
  "backends": [
    {{ range $i := loop 0 30 -}}{{ if (env (printf "SERVICE_NAME_%d" $i)) }}
    {
      "name": "{{ env (printf "SERVICE_NAME_%d" $i) }}",
      "onChange": "/bin/on_change.sh"
    },
    {{- end }}{{- end }}
```
Probably a might be a better way or a custom function that can solve this better.

tgross:
My main filters for adding new features and configuration complexity are that the added complexity is both necessary to the problem and isolated (opt-in). This feature seems to pass those filters; you can't dynamically add backends via environment otherwise and the functions don't exist to someone who doesn't use them. Matching what `consul-template` does is a nice touch too.

Note that you're showing v2 configuration here, and this feature can only go into v3 (the `master` branch). We're not adding any new features to v2 (the `v2_backports` branch).

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/3.0.0-RC1

PR #381, opened 2017-05-26
Support loop and env functions

ahelal:
Reference #370 

This is the initial attempt to support _loop_ and _env_ I added a few tests.
For the env it is just a wrapper for `os.Getenv` not sure if that is clean ?

Please let me know your feedback.
Cheers

### DIFF
--- a/config/template.go
+++ b/config/template.go
@@ -57,6 +57,39 @@ func parseEnvironment(environ []string) Environment {
 	return env
 }
 
+func envFunc(env string) string {
+	return os.Getenv(env)
+}
+
+// loop accepts 1 or two parameters
+// loop 5  returns 0 1 2 3 4 or loop 5 8 returns 5 6 7 or loop 5 1 returns
+func loop(params ...int) ([]int, error) {
+	var start, stop int
+	result := []int{}
+
+	switch len(params) {
+	case 1:
+		start, stop = 0, params[0]
+	case 2:
+		start, stop = params[0], params[1]
+	default:
+		return nil, fmt.Errorf("loop: wrong number of arguments, expected 1 or 2"+
+			", but got %d", len(params))
+	}
+
+	if stop < start {
+		for i := start; i > stop; i-- {
+			result = append(result, i)
+		}
+	} else {
+		for i := start; i < stop; i++ {
+			result = append(result, i)
+
+		}
+	}
+	return result, nil
+}
+
 // Template encapsulates a golang template
 // and its associated environment variables.
 type Template struct {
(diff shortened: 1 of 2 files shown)


config/template_test.go
```diff
@@ -27,7 +27,8 @@ func TestParseEnvironment(t *testing.T) {
 
 func TestTemplate(t *testing.T) {
 	env := parseEnvironment([]string{"NAME=Template", "USER=pilot", "PARTS=a:b:c"})
-
+	// To test env function
+	os.Setenv("NAME_1", "Template")
```
tgross: Can you `defer` unsetting this variable after you set it here, so that we're not interfering with the global environment for the rest of the tests?

config/template.go
```diff
@@ -57,6 +57,39 @@ func parseEnvironment(environ []string) Environment {
 	return env
 }
 
+func envFunc(env string) string {
+	return os.Getenv(env)
+}
+
+// loop accepts 1 or two parameters
+// loop 5 returns 0 1 2 3 4 or loop 5 8 returns 5 6 7 or loop 5 1 returns 5 4 3 2
+func loop(params ...int) ([]int, error) {
```
tgross: This `loop` function is different than [the one in `consul-template`](https://github.com/hashicorp/consul-template/blob/92746fc5cf86dbb113558bacec43459a65c8df14/template/funcs.go#L663-L685). But if I'm reading everything right this might mean that we have the added flexibility of having something like `{{ range $i, $e := loop 5 8 }}` work.

Can you demonstrate in tests how this function would be consumed? i.e. can you show in the test a couple examples of using `range` with this?

ahelal:
Hey Tim updated the tests.

here is an example of how the watchers might worker
```
  watches: [
    {{ range $i := loop 0 30 -}}{{ if (env (printf "SERVICE_NAME_%d" $i)) -}}
    {{ $ServiceName := (env (printf "SERVICE_NAME_%d" $i)) -}}
    {{ $ServiceInterval := (env (printf "SERVICE_INTERVAL_%d" $i)) | default 10 -}}
    {
      name: "{{ $ServiceName }}",
      interval: {{ $ServiceInterval }}
    },
    {{- end }}{{- end }}
  ],
```
But I am still struggling with trailing _,_ I am not familiar with JSON5, but what I read is that it trailing _,_ should work ?

ahelal:
@tgross sorry my bad the trailing _,_ is not a problem 👍

tgross:
> But I am still struggling with trailing , I am not familiar with JSON5, but what I read is that it trailing , should work ?

Yup! That's one of the reasons why we switched to JSON5 for CPv3.

This is looking great @ahelal, thanks for the contribution! It'll get released in the CPv3 release candidate coming out this week.

ahelal:
@tgross thank you for a great product 👍

PR #383, opened 2017-05-26
refactor discovery package to lift Consul methods into wrapper

tgross:
For #317 

- make it so that `Job` only has a single field in the `discovery` package (`ServiceDefinition`), which is responsible for keeping the pointer to the Consul client rather than having the job do both.
- lifted contents of `discovery/consul` subpackage up to `discovery`
- eliminated runtime registration of `discovery` package
- `discovery.Consul` now wraps all Hashi Consul api calls, rather than callers having to call `Agent()` and `Health()` and then calling _their_ methods transitively. This makes for easier mocking in consumer packages.

Somewhat unrelated to the original issue but fixed:
- fixed lots of incorrect nil interface{} comparisons in the config parsing tests. these were all passing but would not have detected a failure if there was one.
- fix a minor bug in #275 which is that the `wasRegistered` field is being kept alongside the client struct, and not the service definition. If a container has more than one service advertised then the second (or more) services to register will bypass this check and will log the warning that we intended to eliminate in #275.

@cheapRoc @misterbisson 

Right now I'm not quite comfortable merging this because reloading a configuration is now taking _much_ longer (6 or 7 seconds on my machine instead of subsecond). It looks like the problem is happening because the event bus is blocked in `Wait`, but other than that I haven't figured out the mechanism here.


tgross:
Re the slow reload, I captured logs from a passing case and it looks like the telemetry server's `stopping` and `stopped` events are falling very far apart (that's the `{Stopping containerpilot}` and `{Stopped containerpilot}` events).

```
time="2017-05-26T19:48:44Z" level=debug msg="control: stopping control server" 
time="2017-05-26T19:48:44Z" level=debug msg="control: completed graceful shutdown of control server" 
time="2017-05-26T19:48:44Z" level=debug msg="control: stopped serving at /var/run/containerpilot.socket" 
time="2017-05-26T19:48:44Z" level=debug msg="event: {Stopping app}" 
time="2017-05-26T19:48:44Z" level=debug msg="deregistering: app-57202de7b4d4" 
time="2017-05-26T19:48:44Z" level=debug msg="event: {Stopping preStart}" 
time="2017-05-26T19:48:44Z" level=debug msg="event: {Stopped preStart}" 
time="2017-05-26T19:48:44Z" level=debug msg="event: {Stopping containerpilot}" 
time="2017-05-26T19:48:44Z" level=debug msg="deregistering: containerpilot-57202de7b4d4" 
time="2017-05-26T19:48:44Z" level=debug msg=preStart.kill 
time="2017-05-26T19:48:44Z" level=debug msg="killing command 'preStart' at pid: 11" 
time="2017-05-26T19:48:44Z" level=error msg="preStart exited with error: preStart: signal: killed" 
time="2017-05-26T19:48:44Z" level=debug msg="event: {ExitFailed preStart}" 
time="2017-05-26T19:48:44Z" level=debug msg="event: {Error preStart: signal: killed}" 
time="2017-05-26T19:48:44Z" level=debug msg="preStart.Run end" 
time="2017-05-26T19:48:52Z" level=debug msg="event: {Stopped containerpilot}" 
time="2017-05-26T19:48:52Z" level=debug msg="event: {Stopped app}" 
time="2017-05-26T19:48:52Z" level=debug msg="loaded config: 
```

tgross:
Ok, well even weirder the build on TravisCI isn't seeing this problem at all. The reload there happens in less than a second. 😦

tgross:
Ok, I traced through all the execution of the reload and the time is entirely around this line of code: https://github.com/joyent/containerpilot/pull/383/files#diff-d7cfa9da32209118d868a9111ea50cafR70, which means that it's Consul that's taking several seconds to come back with a response.

It turns out we're trying to deregister a service that we haven't yet registered --  we reload before jobs get a chance to pass their heartbeat. Consul returns 200 OK in this case but logs a warning. Consul seems to take a long time to come back.

app log (with extra debugging around the line of code in question):
```
time="2017-05-30T14:01:12Z" level=debug msg="calling c.Agent().ServiceDeregister() for: containerpilot-4494fb06129b"
time="2017-05-30T14:01:12Z" level=debug msg="calling c.Agent().ServiceDeregister() for: app-4494fb06129b"
time="2017-05-30T14:01:18Z" level=debug msg="finished c.Agent().ServiceDeregister() for: app-4494fb06129b"
time="2017-05-30T14:01:18Z" level=debug msg="finished c.Agent().ServiceDeregister() for: containerpilot-4494fb06129b"

```

Relevant bits of consul log:
```
    2017/05/30 14:01:12 [WARN] agent: Failed to deregister service "app-4494fb06129b": Service does not exist
    2017/05/30 14:01:18 [DEBUG] http: Request PUT /v1/agent/service/deregister/app-4494fb06129b (5.912686686s) from=192.168.48.3:60568
    2017/05/30 14:01:18 [WARN] agent: Failed to deregister service "containerpilot-4494fb06129b": Service does not exist
    2017/05/30 14:01:18 [DEBUG] http: Request PUT /v1/agent/service/deregister/containerpilot-4494fb06129b (5.911411517s) from=192.168.48.3:60570
    2017/05/30 14:01:19 [DEBUG] http: Request PUT /v1/agent/service/register (386.811µs) from=192.168.48.3:60578
    2017/05/30 14:01:19 [DEBUG] http: Request PUT /v1/agent/service/register (373.764µs) from=192.168.48.3:60578
    2017/05/30 14:01:19 [DEBUG] http: Request PUT /v1/agent/check/register (440.275µs) from=192.168.48.3:60578
```

So this PR should be ok to merge. The nagging doubt is that I can't figure out is _why_ Consul is taking so long to come back and that makes me anxious. 😀  I haven't been able to replicate the behavior with hand-rolled `curl` yet, so it might be an issue of rapidly-sent requests to a very small Consul node.

tgross:
Ok, I can replicate the problem if I launch Consul and immediately hit it with the `v1/agent/service/deregister` requests, but not if I launch Consul, wait for it to settle, and then hit it with requests. So the problem here is just that the test is not waiting for Consul to be ready before we’re hitting it with requests. Which means we've actually verified that ContainerPilot degrades gracefully when the Consul server is slow.

This PR is ready to merge.

Issue #402, opened 2017-06-15
Docs build results in errors


fitzage:
After running `make kirby`, I get the following errors. It's unclear if all the docs get built successfully. Nothing appears to be missing at a glance.

```
bash ./scripts/docs.sh
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
sed: 1: "build/docs/10-lifecycle ...": undefined label 'uild/docs/10-lifecycle/docs.md'
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
sed: 1: "build/docs/20-design/do ...": undefined label 'uild/docs/20-design/docs.md'
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/31-installation/docs.md'
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/32-configuration-file/docs.md'
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/33-consul/docs.md'
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/34-jobs/docs.md'
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
sed: 1: "build/docs/30-configura ...": undefined label 'uild/docs/30-configuration/35-watches/docs.md'
sed: illegal option -- r
[...]

PR #404, opened 2017-06-15
remove accidental gnu-sed assumption in kirby doc gen

tgross:
For #402 

The doc generation script accidentally had a GNU-sed-ism in there. I've removed that and generally cleaned up the script so that it's more robust bash.

cc @fitzage


tgross:
@fitzage sez (in Slack):
> Tested the docs build from that branch, and it appears to be fixed.

PR #407, opened 2017-06-16
improve link rewrites and file paths for kirby docs

tgross:


tgross:
Published to https://www.joyent.com/containerpilot/docs

Issue #154, opened 2016-05-12
Expose ContainerPilot state and configuration through telemetry interface


misterbisson:
We introduced [application telemetry in ContainerPilot 2.0](https://www.joyent.com/blog/containerpilot-telemetry) as discussed in https://github.com/joyent/containerpilot/issues/27, but some of the features there remain. Specifically, those described in https://github.com/joyent/containerpilot/issues/27#issuecomment-207462572:

> In addition to the application performance metrics that can be used to support scaling decisions, this interface can provide a number of details that are known to Containerbuddy:
> - The service(s) provided by the container, what ports they're on
> - Their current health
> - The names of upstream services the container depends on
> 
> Using that information, a consumer application can traverse the all the containers in an application, graph their relationship, and display a visual representation of the entire application, its health, and performance.

...and from https://github.com/joyent/containerpilot/issues/135#issuecomment-211513745:

> The telemetry service [...] represents ContainerPilot itself. The roadmap for the telemetry service includes presenting additional metadata about the container. Predictability in the service name is intended as a way to simplify the development of consumers and new thinking about how this data can be used.

justenwalker:
Would this be another HTTP endpoint (eg: `/status` ) responding with this data as a JSON object?
To what schema would that JSON object conform? I don't really know of an existing schema, so we'd possibly be inventing one here.

Unfortunately, these characteristics do not fit to any metric type. Perhaps we could abuse gauge by alternating between 1 and 0 for each service name to indicate health - but it would be icky.

For reference, [Metric Types](https://prometheus.io/docs/concepts/metric_types/) are:

- **counter** - A counter is a cumulative metric that represents a single numerical value that only ever goes up.
- **guage** - A gauge is a metric that represents a single numerical value that can arbitrarily go up and down.
- **histogram** - A histogram samples observations and counts them in configurable buckets. It also provides a sum of all observed values.
- **summary** - A summary samples observations. While it also provides a total count of observations and a sum of all observed values, it calculates configurable quantiles over a sliding time window.

misterbisson:
> Unfortunately, these characteristics do not fit to any metric type

Agreed, this should not be wedged into the metrics data, and Prometheus (as it exists today) is not a likely consumer.

> I don't really know of an existing schema, so we'd possibly be inventing one here

Right. Action here is awaiting some use-cases to help define what needs to be in there, which might help identify useable prior art.

tgross:
[...]

PR #408, opened 2017-06-16
telemetry /status endpoint

tgross:
This is a work-in-progress PR for #154.

Internally this is inherently a little messy because we have to cross concerns between Jobs/Watches and Telemetry.  During each configuration load, we add the information about the current configuration's Jobs and Watches so that we can construct most of the response on a one-time basis and then only have to poll over the slice of pointers we have to all the `Job`s, calling `GetStatus()` on each. Then we just dump the whole struct to JSON.

Note that the `GetStatus()` call is just checking the status set by the job's health checks, so there's no requests to Consul or anything similarly failure-prone here. This minimizes the amount of cross-talk between components except at config loading time.

---

TODO:
- [x] comprehensive tests
- [x] add a call to `/status` in the telemetry integration tests

But I've smoke tested this with a local build. If I have a ContainerPilot config like 

```
{
  consul: "localhost:8500",
  //logging: { level: "DEBUG" },
  jobs: [
    {
      name: "app",
      exec: "sleep 8600",
      port: 80,
      health: {
        exec: "true",
        interval: 10,
        ttl: 20
      }
    },
    {
      name: "sidecar",
      exec: "sleep 8600"
    }
  ],
  watches: [
    {
      name: "app",
      interval: 10
    },
    {
      name: "otherapp",
      interval: 5
    }
  ],
  control: {
    socket: "./containerpilot.sock"
  },
  telemetry: {
    port: 9090,
    interfaces: ["lo0"]
  }
}
```

Then when I run this via `./build/containerpilot -config ./local.json` and hit it with `curl -s localhost:9090/status | jq .`, I get the following which meets the spec discussed in the [RFD86](https://github.com/joyent/rfd/blob/master/rfd/0086/mariposa.md#getstatus-get-status):

```json
{
  "Services": [
    {
      "Name": "app",
      "Address": "192.168.1.236",
      "Port": 80,
      "Status": "unknown"
    },
    {
      "Name": "containerpilot",
      "Address": "127.0.0.1",
      "Port": 9090,
      "Status": "healthy"
    }
  ],
  "Watches": [
    "app",
    "otherapp"
  ]
}
```

cc @cheapRoc @geek @misterbisson


tgross:
I've pushed a commit that adds 100% branch coverage for unit tests and also an integration smoke test. This should be good-to-go.

cheapRoc:
This also LGTM but consumers should review this one.

Issue #409, opened 2017-06-19
Many <defunct> processes when its parent is killed by containerpilot


roengram:
Hi,

I'm using containerpilot 2.7.4.
I created a task that executes a shell script periodically, which in turn executes a command.
When the command process takes longer than the timeout configured for the task, containerpilot kills the shell process by sending it SIGKILL. This makes the command process parentless and it eventually becomes <defunct> when finished. Because the task is executed periodically, the number of defunct processes grows until it hits the max process number. 

I've created a tiny demo for clarity: 
[containerpilot.tar.gz](https://github.com/joyent/containerpilot/files/1084207/containerpilot.tar.gz)
The task just invokes sleep longer than its timeout. Several minutes after launching the demo with "docker-compose up --build", you can observe many defunct sleep processes. I also included log file.

Prefferably, I think containerpilot should kill the entire process tree of a timedout task. Alternatively, contaierpilot should send a signal other than SIGKILL so that users can specify trap for the signal.

tgross:
Thanks for the report, @roengram. This is definitely a bug.

ContainerPilot should be reaping the whole process group. I checked on the code and we're [doing it correctly in v3](https://github.com/joyent/containerpilot/blob/837b3ab63142394f885217a128520bf329f11634/commands/commands.go#L140). But somehow this never made its way [back to v2](https://github.com/joyent/containerpilot/blob/c45436a725458c4ead0c6ec1ac5a06b7e77d06e5/commands/commands.go#L175)

Thanks for providing such a clear reproduction! This should be an easy fix and I'll get right on this.

tgross:
Just for the record it looks like this was introduced in 2.7.3 in https://github.com/joyent/containerpilot/pull/334. Simply sending SIGTERM to the top-level command process was fine before that because we had a general SIGCHLD signal handler. By introducing the reaper code that wasn't going to work anymore.

tgross:
Opened https://github.com/joyent/containerpilot/pull/410

tgross:
Not only is this an issue for v2 but I've also been able to confirm it's an issue in v3 now too. ☹️

tgross:
We have an integration test that should be catching this, so I've had a look at that. It looks like unfortunately the test is simply wrong. Here's what I get if I echo the outputs of `test_reap_zombies/run.sh`:

```
2017/06/19 09:49:34 TEST: test_reap_zombies
STAT PPID  PID   COMMAND
S        0     1 containerpilot
[...]

PR #410, opened 2017-06-19
v2: reapChildren should use the pid and not the pgid when pgid==0

tgross:
For #409 

When we swapped out the general SIGCHLD handler for the more specific child reaper in #334 we didn't update the `Kill` method to kill the whole process group (as was done in the v3 branch). This means we get zombies if we kill a task that has children.

This PR fixes the above and changes the unit test; we don't get an error back from the call to `syscall.Kill` so we need to check for its effects rather than errors.

---

*Edit: reopened*

Whenever our pgid is 0, the reapChildren function won't have anything to reap so we end up with zombies. So in that case we'll pass the pid; we need to move this check to after the start of the process so that we have a valid pid.

cc @cheapRoc

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -136,17 +136,7 @@ func RunAndWaitForOutput(c *Command) (string, error) {
 	c.Cmd.Stderr = os.Stderr
 	log.Debugf("%s.Cmd.Output", c.Name)
 
-	// See https://golang.org/src/syscall/exec_linux.go
-	// SysProcAttr.Setpgid:
-	// "Set process group ID to Pgid, or, if Pgid == 0, to new pid"
-	// So in the case where ContainerPilot is PID1 this will fail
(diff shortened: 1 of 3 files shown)


tgross:
Hm, although it looks like the tests will pass with this, the reproduction that @roengram provided us is still showing the zombie processes. I'm going to need to keep digging into why that is.

tgross:
I'm going to close this for now as I've been able to confirm it's a problem in v3 builds as well so this may be the wrong approach.

tgross:
I've had to revert the fix for the `RunAndWaitForOuput` method, which is used to power the `sensors` hooks. We're going to hit a NPE in that call so long as we keep the call to `Cmd.Output`. I want to resolve that but let's make sure we have the primary case fixed first.

(Sensors frankly don't work in general in v2, but that's a design issue and not a bug so I'm less worried about this in terms of backporting the fix.)

PR #413, opened 2017-06-21
factor argument parsing out of LoadApp

tgross:
This PR breaks up the very large `core.LoadApp` in favor of having all the arg parsing in one place, and then letting `main.go` decide what to do with that. Not a critical fix of anything but the size of `LoadApp` was starting to get unwieldy.

cc @cheapRoc

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,7 +13,16 @@ func main() {
 	// contention on the main application
 	runtime.GOMAXPROCS(1)
 
-	app, configErr := core.LoadApp()
+	subcommand, params := core.GetArgs()
+	if subcommand != nil {
+		err := subcommand(params)
+		if err != nil {
+			log.Fatal(err)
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -13,7 +13,16 @@ func main() {
 	// contention on the main application
 	runtime.GOMAXPROCS(1)
 
-	app, configErr := core.LoadApp()
+	subcommand, params := core.GetArgs()
+	if subcommand != nil {
+		err := subcommand(params)
```
cheapRoc: I feel like there's a little misdirection/in-and-out going on here but, hey, it's way cleaner because of it! I like the use of returning a func and combining the args.

tgross:
Had to rebase for a merge conflict... waiting for this test run to pass then I'll merge this.

Issue #423, opened 2017-07-03
Failed to check service health due to logging on closed pipe


ochanism:
Recently, I updated the containerpilot version from 2.7.3 to 2.7.4 and 2.7.5 and I got the error as described below.

> `2017/07/03 11:28:35 us-east-1b-log-zookeeper.health exited with error: io: read/write on closed pipe`

Actually, Zookeeper health was fine and health check script exited with code 0. But Zookeeper health on consul was out-of-service. When starting the service, it looked fine, but the problem occurred in a few days. I looked into containerpilot code to figure out what caused it. The error came out from [here](https://github.com/joyent/containerpilot/blob/2.7.5/commands/commands.go#L245) and it might have been caused by closing the logger in [here](https://github.com/joyent/containerpilot/blob/2.7.5/commands/commands.go#L48). I couldn't figure out why the logger was closed and who did it. When I sent SIGHUP to a containerpilot process, the logger pipe was recreated and the service health was changed to in-service.

Anyway, in this situation, I think service health should be in-service, even if the logger is closed.
I've attached the some scripts below to help you debug this issue.

**health check script**
```
#!/bin/bash
echo ruok | nc ${HOSTNAME} ${ZOOKEEPER_PORT}
```

**containerpilot configuration**
```
{
  "consul": "localhost:8500",
  "preStart": "/zookeeper/{{.ZOOKEEPER_NAME}}/config/preStart.sh",
  "logging": {
    "level": "INFO",
    "format": "default",
    "output": "stdout"
  },
  "stopTimeout": 5,
  "preStop": "/zookeeper/{{.ZOOKEEPER_NAME}}/config/preStop.sh",
  "postStop": "/zookeeper/{{.ZOOKEEPER_NAME}}/config/postStop.sh",
  "services": [
    {
      "id": "{{.DC}}-{{.ZOOKEEPER_CONSUL_ID}}-{{.HOSTNAME}}",
      "name": "{{.DC}}-{{.ZOOKEEPER_CONSUL_NAME}}",
      "port": {{.ZOOKEEPER_PORT}},
      "health": "/zookeeper/{{.ZOOKEEPER_NAME}}/config/healthCheck.sh",
      "tags": [
          {{.ZOOKEEPER_TAGS}}
      ],
      "interfaces": [
          {{.CONTAINER_PILOT_INTERFACES}}
      ],
      "poll": 10,
      "timeout": "10s",
      "ttl": 30
    }
  ],
  "backends": [
  ],
[...]

Issue #393, opened 2017-06-09
Annotate logs by generating process


jhmartin:
Coprocesses and tasks have names. When those processes emit STDOUT logs, include that name tag in some sort of parseable fashion so that log parsers can differentiate between each process. Include a tag on the 'main' process as well.

tgross:
In v2 we didn't add log structure because we had a "main" application which was assumed to take care of its own log structure. The `coprocess` and `task` hooks came later and probably should have had log structure but didn't because we reused most of the mechanisms we were using to run the "main" application. It looks like you're proposing adding this as a feature but we're not adding any new features to v2 of ContainerPilot.

However in v3 of ContainerPilot (the master branch and [release candidate](https://github.com/joyent/containerpilot/releases/tag/3.0.0-RC1)) no longer has a concept of a "main" application and we do wrap all the log lines for each `job` in the log structure. I was pretty sure we have the job names in the fields already but I'll double-check. In the meantime I'll mark this as a `3.x` proposal.

tgross:
@jhmartin I dug into this area while looking at #423 and discovered a few things (ref https://github.com/joyent/containerpilot/issues/423#issuecomment-312725831):

- we're definitely supposed to be setting up the commands so that they log their names
- we accidentally broke the logger that was supposed to pick up those names in v3
- we had a regression in v2 on the same item in 2.7.1.

I've got a fix in the works and I'll have it up for review sometime Weds morning.

tgross:
PR open for v3 https://github.com/joyent/containerpilot/pull/424

tgross:
I've merged #424 and #425 and I'll release that fix in ContainerPilot 2.7.6 and ContainerPilot 3.1.3. I want to make sure we've done what we can about #423 first, but I won't hold the patch release up for long if that proves to be a lengthy process.

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/3.1.3 and https://github.com/joyent/containerpilot/releases/tag/2.7.6

mterron:
Does not log job names on 3.6.2 @tgross 
Example log there should be a "redis" after the timestamp and before the 19:M:
```
2018-01-15T00:03:49.607445593Z 19:M 15 Jan 00:03:49.607 * 10000 changes in 60 seconds. Saving...
2018-01-15T00:03:49.609117243Z 19:M 15 Jan 00:03:49.608 * Background saving started by pid 995
[...]

PR #424, opened 2017-07-03
update logrus to 1.0, use Entry.Writer to catch exec stdout

tgross:
For #393 and possibly #423 if it applies to v3

Fixes a regression in log formatting for exec stdout/stderr; we were creating a new logger without any of our configuration. In Logrus 1.0 we can access the Writer of a contextual logger and attach it to our process's stdout/stderr.

Will need to backport this to v2 as well.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -3,7 +3,6 @@ package commands
 import (
 	"context"
 	"fmt"
-	"io"
 	"os/exec"
 	"sync"
 	"syscall"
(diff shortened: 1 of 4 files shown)


tgross:
Per discussion in #423 (ref https://github.com/joyent/containerpilot/issues/423#issuecomment-312969599), this won't fix that problem. But it will still fix the regression in #393 and remove any possible raciness in the stdout/stderr capture of our `exec`.

tgross:
I'm going to fix the logrus package name problem (ref https://github.com/sirupsen/logrus/issues/553) under a separate PR so as not to confuse the hell out of this PR.

Issue #422, opened 2017-07-03
Emit RFC3339/ISO8601 dates


mterron:
From observation, Containerpilot seem to be using UTC time for it's log (please correct me if I'm wrong).
Currently Containerpilot emits dates on the following format:
YYYY/MM/DD hh:mm:ss

Instead, it should output its log dates in a std format like [RFC3339](https://www.rfc-editor.org/rfc/rfc3339.txt)

tgross:
We're currently emitting the default configuration of the [formatters](https://github.com/sirupsen/logrus#formatters) provided with Logrus. Digging into the Logrus code this suggests we should be getting [`RFC3339`](https://github.com/sirupsen/logrus/blob/7dd06bf38e1e13df288d471a57d5adbac106be9e/formatter.go#L5) by default, but maybe this was changed in a newer version of Logrus. I'll have to look into it.

Arguably we might want to [pass options](https://godoc.org/github.com/sirupsen/logrus#TextFormatter) to the formatter structs so as to format them in a particular way determined by the user, and just have a sane default.

tgross:
I dug into this a bit more and the output varies depending on the log formatter chosen:

- `default` uses the golang stdlib logger and the `LstdFlags` format. The stdlib doesn't give us the option of a RFC3339 format as far as I can tell.
- `json` uses the RFC3339 format w/ TZ: `{"level":"debug","msg":"app1.kill","time":"2017-07-03T13:18:04-04:00"}`
- `text` uses logrus Text format, which is frankly a little weird:
  - If there's a TTY attached, we get messages like `DEBU[0006] app1.kill`, where "0006" is the _number of seconds elapsed since start_. (Wtf?)
  - If there's no TTY attached, we get a format like the following: `time="2017-07-03T17:26:00Z" level=debug msg=app1.kill`

So the only case where we're not getting the RFC3339 format is in the stdlib log format. Unless you know of a way to coerce the stdlib log format to do otherwise, it's my recommendation to use the `text` format or `json` format. These are also structured formats which are better for log ingest.

mterron:
This snippet emits rfc3339 dates using the std log lib.
https://play.golang.org/p/RC9JxV_LaA
Can you take a look Tim?

tgross:
Seems sane as far as the timestamp goes, but how do we get the level block in there? I'm definitely open a PR on this.

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/3.2.0

PR #426, opened 2017-07-03
Emit RFC3339 dates in logs

mterron:
This PR formats the date emited by Containerpilot according to RFC3339 as per [#442](https://github.com/joyent/containerpilot/issues/422).
This is useful when consuming CP logs for correlation with other time sources as it remove ambiguity related to timezones.

If this patch works, dates in the logs should look like:
```
2017-07-04T23:00:00.00000Z REST OF THE LOG LINE
```

### DIFF
--- a/config/logging.go
+++ b/config/logging.go
@@ -76,11 +76,10 @@ func (l *LogConfig) init() error {
 type DefaultLogFormatter struct {
 }
 
-//RFC3339 formatter
+// RFC3339logWriter io.Writer that outputs RFC3339 dates
 type RFC3339logWriter struct {
 }
 
-// Define new time io.writer that outputs RFC3339 dates
 func (writer RFC3339logWriter) Write(bytes []byte) (int, error) {
 	return fmt.Print(time.Now().UTC().Format(time.RFC3339Nano) + " " + string(bytes))
 }
--- a/config/logging.go
+++ b/config/logging.go
@@ -81,7 +81,7 @@ type RFC3339logWriter struct {
 }
 
 func (writer RFC3339logWriter) Write(bytes []byte) (int, error) {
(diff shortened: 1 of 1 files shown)


mterron:
This initial PR assumed time is in UTC, will need to add TZ parsing to be able to output in local time I think.

tgross:
> This initial PR assumed time is always output in UTC, will need to add TZ parsing to be able to output in local time I think.

If you take a look at https://golang.org/pkg/log/#pkg-constants you'll see that the `LstdFlags` format we're using is the local time. You can just drop the `.UTC()` call and you'll be good-to-go.

```go
package main

import (
	"fmt"
	"time"
)

func main() {
	fmt.Println(time.Now().UTC().Format(time.RFC3339Nano))
	fmt.Println(time.Now().Format(time.RFC3339Nano))
}
```

```sh
$ go run main.go
2017-07-05T12:45:56.944643712Z
2017-07-05T08:45:56.944688271-04:00
```

(You can try this at https://play.golang.org/p/wbC1AnDmBV but they're locked to UTC time there so that's useful.)

---

Also just for the record, I asked in https://github.com/joyent/containerpilot/issues/422#issuecomment-312729108 how we'd get the log level in there, but I was mistaken -- the golang stdlib logger doesn't include the log level anyways.

tgross:
LGTM 👍

tgross:
Pushed a quick commit https://github.com/joyent/containerpilot/commit/9489dbf56e4f6a2aa2e21e0c6cfb81603c33f682 to update the docs too.

tgross:
Released in https://github.com/joyent/containerpilot/releases/tag/3.2.0

cheapRoc:
This was found to have caused unnecessary zero byte writes stemming from the fact that the returned buffer held no data. It looks like printing was being done outside logrus all together. 

Ref: #540.

PR #427, opened 2017-07-05
change tll -> ttl and fix schema format

JeromeGe:
Thanks for contributing to ContainerPilot! Please include with your pull request:

- A description of what you did for the changelog
- An explanation of why ContainerPilot needs this change
- How to verify that it works (most PRs need tests!)
- A link to the GitHub issue that it addresses

If you're contributing a new feature, it's usually better to open an issue to discuss the feature rather than open a pull request unless the feature is trivial.


tgross:
@JeromeGe we've asked in the PR template that contributors include a description of what / why they're doing in the PR. But in this case it looks like it's a trivial docs fix, so I'll merge this.

PR #428, opened 2017-07-05
update logrus import to use lower case package name

tgross:
The logrus package had its name changed from `github.com/Sirupsen/logrus` to `github.com/sirupsen/logrus` (ref https://github.com/sirupsen/logrus/issues/553). This PR updates all our imports and our glide config to use the new package name.

Once this PR is merged, anyone developing on ContainerPilot will need to do a `make clean` to ensure they don't have any debris from the old vendor directory or project glide cache floating around.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -8,8 +8,8 @@ import (
 	"syscall"
 	"time"
 
-	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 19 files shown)


tgross:
(Backported this to `v2_backports` as well)

PR #432, opened 2017-07-06
add internal prometheus metrics

tgross:
For https://github.com/joyent/containerpilot/issues/374

This PR gives us a lot of the items described in issue #374, because most of the things that ContainerPilot does pass through the event bus `Publish` method. The list of events this will catch are:

- *ExitSuccess*: any `exec` (job or health check) completes with 0 exit code
- *ExitFailed*: any `exec` (job or health check) completes with 0 exit code
- *Stopping*: a job in this container is about to stop
- *Stopped*: a job in this container has stopped
- *StatusHealthy*: a job (or watched job) is healthy
- *StatusUnhealthy*: a job (or watched job) is unhealthy
- *StatusChanged*: a job (or watched job) status has changed
- *EnterMaintenance*: entered maintenance mode
- *ExitMaintenance*: exited maintenance mode
- *Error*: an `exec` couldn't start or was forced to cancel with an error b/c it timed out (will always have *ExitFailed* too, but *ExitFailed* might not include *Error*).
- *Metric*: a user-defined metric was recorded
- *Startup*: global ContainerPilot startup
- *Shutdown*: global ContainerPilot shutdown

(The internal *TimerExpired* and *Quit* events you can find in [events/events.go](https://github.com/joyent/containerpilot/blob/master/events/events.go) are never broadcast across the event bus so won't show up here.)

This Prometheus collector here is a vec counter, or in other words a counter that's annotated with labels. The output looks like this:

```
curl -s localhost:9090/metrics | grep containerpilot
# HELP containerpilot_events count of ContainerPilot events, partitioned by type and source
# TYPE containerpilot_events counter
containerpilot_events{code="ExitSuccess",source="check.app1"} 3
containerpilot_events{code="Startup",source="global"} 1
containerpilot_events{code="StatusHealthy",source="app1"} 3
```

This PR needs tests too, but I wanted to use this for discussion.
cc @cheapRoc @misterbisson @geek

PR #433, opened 2017-07-06
refactor: split utils package out to config subpackages

tgross:
As part of an effort to make future contributors have an easier time figuring out what the heck is going on, this PR removes the "junk drawer" of the utils package and places its responsibilities as a subpackage of `config`. As much as I'd like to have these be just within the `config` package, that'll cause circular references when they're imported into the `config.go` files associated with most of the other packages.

cc @cheapRoc 
(Note this might have merge conflicts if #429 and #432 are merged first, but those are easy cleanups)


cheapRoc:
Thinking of best ways to reduce stutter across this series of changes. Arguably, I am not considering implementations here.

- `decoding.DecodeRaw()` => `decoding.Raw()`
- `templating.ApplyTemplate()` => `templating.Apply()`
- `decoding.DecodeArray()` => `decoding.Array()`. Not sure if `ToArray()` might work? I see `ToStringArray`.

tgross:
Ok, I've pushed a commit that does:
- `services.ValidateServiceName` --> `services.ValidateName`
- `templating.ApplyTemplate` --> `templating.Apply`
- `logging.LogConfig` --> `logging.Config`
- `decoding.DecodeRaw` --> `decoding.ToStruct`
- `decoding.DecodeArray` --> `decoding.ToSlice`
- `decoding.ToStringArray` --> `decoding.ToStrings`

tgross:
Yup, that works. Did:
- `config/logging` -->  `config/logger`
- `config/templating` -->  `config/template`
- `config/decoding` -->  `config/decode`

Issue #439, opened 2017-07-14
some child processes not being reaped


tgross:
A user has reported a problem with ContainerPilot 3.1.2 and a RabbitMQ configuration that leaves a lot of zombies. The relevant anonymized portion of their configuration looks like the following:

```json5
{
  logging: { level: "DEBUG" },
  jobs: [
    {
      name: "rabbitmq",
      exec: "rabbitmq-server",
      health: {
        exec: "rabbitmqctl node_health_check",
        interval: 30,
        ttl: 45
        timeout: "20s"
      }
    }
}
``` 

The relevant anonymized log sections:

```
2017/07/14 19:31:40 event: {Startup global}
2017/07/14 19:31:40 control: serving at /var/run/containerpilot.socket
2017/07/14 19:31:45 rabbitmq.Run start
...
(lots and lots of startup)
...
2017/07/14 19:32:10 check.rabbitmq.Run start
2017/07/14 19:32:11 Timeout: 70.0 seconds ...
2017/07/14 19:32:11 Checking health of node rabbit@0064bb4a4ea6 ...
2017/07/14 19:32:11 Health check passed
2017/07/14 19:32:11 check.rabbitmq exited without error
2017/07/14 19:32:11 event: {ExitSuccess check.rabbitmq}
2017/07/14 19:32:11 check.rabbitmq.Run end
2017/07/14 19:32:11 check.rabbitmq.kill
2017/07/14 19:32:11 killing command 'check.rabbitmq' at pid: 357
2017/07/14 19:32:11 event: {StatusHealthy rabbitmq}

```

These logs don't tell us much other than that our health checks appear to be working and getting killed as expected.

The process table shows many zombie entries for `erl_child_setup`. I'm not an expert on RabbitMQ, but it appears these are children of `/usr/lib/erlang/erts` (the Erlang service manager), which is in turn the child of `rabbitmq-server` -- our `job.exec` and not our `health.exec`. So this doesn't appear to be a regression of #409 which was fixed in 3.1.0.

My working theory is that although the fix in 3.1.0 helps out with health checks and other short-lived job `exec`s, the long-lived `rabbitmq-server` process doesn't exit and so we never hit [the defer that triggers child reaping](https://github.com/joyent/containerpilot/blob/a78b65aa62e7cbe17bc152e301498220a2b924d8/commands/commands.go#L105). My first step will be to try to verify this, but I'm having trouble creating a minimum failing test case.

[...]

PR #440, opened 2017-07-17
move child reaping into supervisor process

tgross:
This is a work-in-progress PR for figuring out a test case and fix for #439

So far:
- I've dropped the `test_probe` fixture and `zombie_maker` fixture in favor of just mounting the test code to the target container; this makes the test more self-contained and improves iteration speed on it.
- I've added a spot in the test target ContainerPilot config for a job called "slow-zombies", because the case I think we're hitting has to do with not reaping during long-lived job `exec`.


tgross:
The latest "wip" commit passes tests locally. I want to smoke test it with the rabbitmq container that discovered this whole issue, and then clean up this PR for review.

tgross:
Ok, I've smoke-tested this PR with the config described in https://github.com/joyent/containerpilot/issues/439#issuecomment-315766276 and it looks good:

```
docker exec -it rabbit ps -eo stat,ppid,pid,args
STAT  PPID   PID COMMAND
Ssl+     0     1 /containerpilot -config /rabbit.json5
Sl+      1    14 /containerpilot -config /rabbit.json5
S       14    19 /bin/sh /usr/lib/rabbitmq/bin/rabbitmq-server
S        1    96 /usr/lib/erlang/erts-8.2.1/bin/epmd -daemon
Sl      19   120 /usr/lib/erlang/erts-8.2.1/bin/beam.smp -W w -A 64 -P 1048576 -
Ss     120   218 erl_child_setup 1048576
Ss     218   232 inet_gethost 4
S      232   233 inet_gethost 4
Rs+      0   252 ps -eo stat,ppid,pid,args
```

tgross:
cc @cheapRoc for review.

commands/commands.go
```diff
@@ -92,56 +84,29 @@ func (c *Command) Run(pctx context.Context, bus *events.EventBus) {
...
-		if pgid == 0 {
-			pgid = c.Cmd.Process.Pid
-		}
-		defer reapChildren(pgid)
-
 		// blocks this goroutine here; if the context gets cancelled
-		// we'll return from wait() and do all the cleanup
-		if err := c.wait(); err != nil {
+		// we'll return from Wait() and publish events
+		if err := c.Cmd.Wait(); err != nil {
```
tgross: Strictly speaking we _could_ now use `Cmd.Run()` instead of `Cmd.Start()` + `Cmd.Wait()` but this would lose the nice error messages where we differentiate "user gave us the wrong exec" vs. "something went wrong after it started"

cheapRoc:
Tests are passing for me as well.

Issue #456, opened 2017-08-01
'when:timeout' incorrectly terminates pipeline of jobs


wheleph:
It seems like `when:timeout` incorrectly terminates pipeline of jobs.

Here's my config file:

```
{
  consul: "localhost:8500",
  logging: {
    level: "DEBUG"
  },
  jobs: [
    {
      name: "job 1",
      exec: "echo === job 1 completed.",
      restarts: "never"
    },
    {
      name: "job 2",
      when: {
        source: "job 1",
        once: "exitSuccess",
        timeout: "10s"
      },
      exec: "/script.sh 2 100"
    }
  ]
}
```

Here `job 1` simulates a setup job which completes fast and `job 2` simulates a long-running service (`script.sh` just prints some info and sleeps for a given amount of seconds).

I expect that both jobs will complete successfully but that's not the case:

```
2017-08-01T13:55:28.703464146Z loaded config: {"Discovery":{},"LogConfig":{"level":"DEBUG","format":"default","output":"stdout"},"StopTimeout":5,"Jobs":[{"Name":"job 1","Exec":"echo === job 1 completed.","Port":0,"Interfaces":null,"Tags":null,"ConsulExtras":null,"Health":null,"ExecTimeout":"","Restarts":"never","StopTimeout":"","When":{"Frequency":"","Source":"","Once":"","Each":"","Timeout":""}},{"Name":"job 2","Exec":"/script.sh 2 100","Port":0,"Interfaces":null,"Tags":null,"ConsulExtras":null,"Health":null,"ExecTimeout":"","Restarts":null,"StopTimeout":"","When":{"Frequency":"","Source":"job 1","Once":"exitSuccess","Each":"","Timeout":"10s"}}],"Watches":null,"Telemetry":null,"Control":{"SocketPath":"/var/run/containerpilot.socket"}}
2017-08-01T13:55:28.703556436Z control: initialized router for control server
2017-08-01T13:55:28.703837501Z event: {Startup global}
2017-08-01T13:55:28.703912711Z control: serving at /var/run/containerpilot.socket
2017-08-01T13:55:28.704122161Z job 1.Run start
2017-08-01T13:55:28.705163619Z === job 1 completed.
2017-08-01T13:55:28.70519698Z job 1 exited without error
2017-08-01T13:55:28.705220352Z event: {ExitSuccess job 1}
2017-08-01T13:55:28.705251514Z job 1.Run end
2017-08-01T13:55:28.705284575Z job 1.term
[...]

PR #458, opened 2017-08-01
Fix when->timeout canceling running jobs

tgross:
For #456 

This PR includes:
- make separate calls to `Job.Subscribe` and `Job.Run` during startup so that we're sure all job that start later than others don't miss events emitted by the earlier jobs.
- make jobs ignore the `when->timeout` timer once they've started by setting their timeout event to the never-published `NonEvent`.

We could probably handle the latter by creating a new context in `events.NewTimeout` from the job's context, and returning a `context.CancelFunc` which `Job.StartJob` could call, but when I actually tried that we end up passing the `CancelFunc` all over the place and it got really messy. We can probably come back to that the next time we take a pass through to clean up the `Job` type's functions.

cc @cheapRoc @wheleph @ewjmulder

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -148,8 +148,15 @@ func (a *App) reload() error {
 // HandlePolling sets up polling functions and write their quit channels
 // back to our config
 func (a *App) handlePolling() {
+
+	// we need to subscribe to events before we Run all the jobs
+	// to avoid races where a job finishes and fires events before
+	// other jobs are even subscribed to listen for them.
 	for _, job := range a.Jobs {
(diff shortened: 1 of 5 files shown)


ewjmulder:
I confirm that the changes in this branch indeed fix our problem!

> We could probably handle the latter by creating a new context in events.NewTimeout from the job's context, and returning a context.CancelFunc which Job.StartJob could call

I agree that would be a cleaner fix (I tried to achieve something along those lines), but this'll do fine for now. Thanks!

Issue #437, opened 2017-07-13
Add an option to leave job's stdout/stderr alone


inetfuture:
As of 3.x, containerpilot warps job's stdout/stderr in its own format as msg field, this behavior is different with 2.x, which just redirect service's stdout/stderr to container's.

I would love to have that old behavior back, perhaps via an option, because our underlying log collecting tool has assumptions about how the service's log should look like, with 3.x, the new format breaks our tool.

tgross:
This would cause job logs to be interleaved and undifferentiated. It doesn't work in any case where there is more than one job, and that includes all uses of ContainerPilot where a job responds to a `watch`.

tgross:
If we were to do this (with the caveat that I think it's not particularly usable), what would you propose as the configuration syntax?

inetfuture:
Sorry for the delay, how about something like this:

```json5
{
  jobs: [
    {
      name: "app",
      exec: "/bin/app",
      restarts: "unlimited",
      port: 80,
      captureStdout: false,
      captureStderr: false
    },
  ]
}
```

Where `captureStdout` and `captureStderr` by default are true, means containerpliot will capture them with its logger. 

Then users like me can choose to set them to false, means containerpilot will not capture them and let them propagate up freely.

inetfuture:
@tgross any suggestions?

tgross:
Sorry for the delay in response here @inetfuture. That config seems somewhat sane. Maybe @misterbisson has opinions about naming of the fields too (he's good at naming things). From an implementation perspective, the ContainerPilot worker process would be the one capturing the logs, so I'm not 100% sure that we wouldn't just be dropping the logs on the floor if we tried piping them to stdout/stderr in a Docker environment. I'll have to dig into that.

misterbisson:
> he's good at naming things

I want that as a merit badge ;-)

The question I'm struggling with, and perhaps it's a bit naive, is if this should be a job-level config item, or if it should be a [log format option](https://www.joyent.com/containerpilot/docs/configuration/logging). I mean, the effect of `captureStdout: false` will be to pass the output straight through to stdout, but not suppress other logging that ContainerPilot is passing through to stdout on its own, right?

inetfuture:
yes

I also thought of whether we should put it inside log config section, the reason I choose jobs section is because I want job configuration to be independent, not involving modifying other parts. So when I need to include a job dynamically, like based on ENV var, there will be only one `if` needed instead of two in the template.

[...]

PR #462, opened 2017-08-03
add raw logging field to bypass logger for exec

tgross:
For #437

This PR implements a "raw" logging option for `exec` fields using the field name `passthru` (which I thought might be more descriptive than "raw"), so that we can pass the stdout/stderr for either job exec or health check exec without wrapping entries in a log. The intent is that the exec structures its own logs, and that users will configure the ContainerPilot logging to one of the structured log formats that matches. Some example configurations from the included integration test:

```json5
jobs: [
  {
    name: "job1",
    exec: "echo job1 exec",  // this echo gets wrapped in a log line
  },
  {
    name: "job2",
    exec: "echo job2 exec", // this echo is not wrapped in a log line
    passthru: true,
  },
  {
    name: "job3",
    port: 8000,
    exec: "sleep 5",
    health: {
      exec: "echo job3 health", // this echo is not wrapped in a log line
      passthru: true,
      interval: 1,
      ttl: 5
    }
  },
  {
    name: "job4",
    port: 9000,
    exec: "sleep 5",
    health: {
      exec: "echo job4 health", // this echo gets wrapped in a log line
      interval: 1,
      ttl: 5
    }
  }
]
```

cc @inetfuture @cheapRoc @misterbisson

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -59,7 +59,7 @@ func (c *Command) Run(pctx context.Context, bus *events.EventBus) {
 	c.lock.Lock()
 	log.Debugf("%s.Run start", c.Name)
 
-	cmd := ArgsToCmd(c.Exec, c.Args)
+	cmd := exec.Command(c.Exec, c.Args...)
 	if c.logger.Logger != nil {
 		cmd.Stdout = c.logger.Writer()
 		cmd.Stderr = c.logger.Writer()
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"context"
 	"fmt"
+	"os"
 	"os/exec"
 	"sync"
 	"syscall"
@@ -29,15 +30,19 @@ func NewCommand(rawArgs interface{}, timeout time.Duration, fields log.Fields) (
 	if err != nil {
 		return nil, err
 	}
-	logger := log.WithFields(fields)
 	cmd := &Command{
 		Name:    exec, // override this in caller
 		Exec:    exec,
 		Args:    args,
 		Timeout: timeout,
 		lock:    &sync.Mutex{},
-		logger:  *logger,
 	} // exec.Cmd created at Run
+
+	if fields != nil {
+		// don't attach the logger if we don't have fields set, so that
+		// we can pass-thru the logs raw
+		cmd.logger = *log.WithFields(fields)
+	}
 	return cmd, nil
 }
 
@@ -55,8 +60,13 @@ func (c *Command) Run(pctx context.Context, bus *events.EventBus) {
 	log.Debugf("%s.Run start", c.Name)
 
 	cmd := ArgsToCmd(c.Exec, c.Args)
-	cmd.Stdout = c.logger.Writer()
-	cmd.Stderr = c.logger.Writer()
+	if c.logger.Logger != nil {
+		cmd.Stdout = c.logger.Writer()
+		cmd.Stderr = c.logger.Writer()
+	} else {
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+	}
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 	c.Cmd = cmd
 	ctx, cancel := getContext(pctx, c.Timeout)
--- a/integration_tests/tests/test_logging/containerpilot.json5
+++ b/integration_tests/tests/test_logging/containerpilot.json5
@@ -14,7 +14,7 @@
       name: "job2",
       // this echo is not wrapped in a log line
       exec: "echo job2 exec",
-      passthru: true,
+      logging: { raw: true },
     },
     {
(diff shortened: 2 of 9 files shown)


commands/commands.go
```diff
@@ -55,8 +60,13 @@ func (c *Command) Run(pctx context.Context, bus *events.EventBus) {
 	log.Debugf("%s.Run start", c.Name)
 
 	cmd := ArgsToCmd(c.Exec, c.Args)
```
cheapRoc: Nit, I realize `ArgsToCmd` isn't a part of this commit but I had to look up the implementation in order to discover that its better represented as `ExecWithArgs`. Even if the exec may not happen with args. `ArgsToCmd` sounds like message passing or type conversion.

commands/commands.go
```diff
@@ -55,8 +60,13 @@ func (c *Command) Run(pctx context.Context, bus *events.EventBus) {
...
-	cmd.Stdout = c.logger.Writer()
-	cmd.Stderr = c.logger.Writer()
+	if c.logger.Logger != nil {
```
cheapRoc: Is it possible for `c.logger == nil`?

integration_tests/tests/test_logging/containerpilot.json5
```diff
@@ -0,0 +1,43 @@
...
+    {
+      name: "job1",
+      // this echo gets wrapped in a log line
+      exec: "echo job1 exec",
+    },
+    {
+      name: "job2",
+      // this echo is not wrapped in a log line
+      exec: "echo job2 exec",
+      passthru: true,
```
cheapRoc: I'm not crazy about the word `passthru` but unable to think of a better alternative for this feature.

commands/commands.go
tgross: The function signature is `ArgsToCmd(executable string, args []string) *exec.Cmd`, so it takes command arguments and makes a `Cmd` out of them. The name is probably lispy though (I'm guilty of that sometimes)... maybe more idiomatic would be `NewCmd`? Or maybe just inline it, as I think there's only one or two callers at this point.

commands/commands.go
tgross: I thought that at first but the field is `logger log.Entry` in the struct, so it'll always be zero value at worst. You can't even make the `c.logger == nil` comparison without the compiler telling you you're wrong. 😀

integration_tests/tests/test_logging/containerpilot.json5
tgross: Yeah, me neither.

commands/commands.go
cheapRoc: Ah, didn't catch that.

integration_tests/tests/test_logging/containerpilot.json5
tgross: Does He-Who-Is-Good-At-Naming-Things (@misterbisson 😀 ) have a suggestion?

commands/commands.go
tgross: On second look, I'm honestly not sure why that function exists at all. The function it wraps is variadic, so we can just call it directly and safely pass in the zero-value for `Command.Args`. Maybe there was more logic there at some point in the past... I don't feel like code-spelunking over it now.

Just pushed a commit removing `ArgsToCmd` entirely and just calling `os/exec.Command` instead.

inetfuture:
Thanks for the support of our requirement.

May I ask, when will this be merged and released?

tgross:
I think we're still bikeshedding the name of the field. If there's no movement on that by end of the day I'll just merge and release.

cheapRoc:
I don't think we're going to get any more descriptive than `passthru` for the moment. I'm 👍 with this getting merged and released.

integration_tests/tests/test_logging/containerpilot.json5
misterbisson: The only question I have about the name is if there's any other interaction besides stdout/logs that might get passed through. If so, the name might be ambiguous.

I'm wondering if the following might make more sense:

- `output: capture` (default, not needed in the config)
- `output: pas.*`

That might be easier to understand without referencing the docs, but it has the disadvantage of requiring a more sophisticated answer than true/false/1/0, etc.

cheapRoc: @misterbisson Unless the value was entirely unambiguous which, in this case, you'd still need to reference the docs for the non-bool value.

tgross:
Sorry for the delay on this. We needed to do some more bikeshedding! 🚲 

We've settled on having a `logging` option that currently has only one field, `raw`. This gives us what we need today but gives us a spot for future expansion if need be. (I think we're all in agreement that individual job-specific loggers is problematic from an implementation standpoint.)

```
jobs: [
  {
    name: "myjob",
    exec: "doStuff.sh",
    logging: { raw: true }
  }
]
```

Will update the PR with this shortly.

Issue #477, opened 2017-08-15
enable log to a file


benoittoulme:
We worked on a fork of containerpilot which enables logging to a file.
We used https://github.com/client9/reopen to enable reopen of a file (in order to support logrotate)
You can see our fork here: https://github.com/fanhattan/containerpilot/tree/reopen
We will link a PR to this issue.

Please let us know what you think about it. You are welcome to merge it.

cheapRoc:
Thanks @benoittoulme, after reviewing everything I can't think of a reason why this wouldn't be something to upstream into ContainerPilot. As you've suggested, it definitely sounds like a useful feature that would help users outputting their logs to files vs. streaming them. 

It would be great to have an integration test around this functionality in order to wrap-up the PR. I can review that part a bit further if that's something you'll need help with.

benoittoulme:
We ll add a test and let you know.

benoittoulme:
Added the tests, you can review when you have time.

PR #478, opened 2017-08-15
support log to file

benoittoulme:
linked to issue https://github.com/joyent/containerpilot/issues/477

### DIFF
--- a/config/logger/logging_test.go
+++ b/config/logger/logging_test.go
@@ -1,9 +1,14 @@
 package logger
 
 import (
+	"fmt"
+	"io/ioutil"
 	"os"
 	"reflect"
+	"strings"
+	"syscall"
 	"testing"
+	"time"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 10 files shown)


config/logger/logging_test.go
```diff
@@ -65,3 +70,66 @@ func TestDefaultFormatterPanic(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("Did not expect error: %v", err)
+	}
+	logMsg = "this is another test"
+	logrus.Info(logMsg)
+	testContainsLog(rotatedFilename, logMsg, t)
+
+	// reopen file
+	syscall.Kill(os.Getpid(), syscall.SIGUSR1)
+	time.Sleep(1*time.Second)
```
cheapRoc: Everything about this unit test feels right except that its adding 3 or 4 seconds to a full 15 second unit suite run. We should keep this unit test but move just the "reopen file" section into an integration test. 

I've begun looking into this but also wouldn't mind a PR enhancement from the author's end. A good integration test to look at would be `integration_tests/tests/test_logging/` and see what can be mimicked for the log re-open behavior. 

Fair enough?

benoittoulme: Sure. Could not get to it yet. Will try to get it done today

cheapRoc:
👍 Perfect contribution! I've got this squashed for merge over at #486.

PR #481, opened 2017-08-17
discovery: fix tls config for Consul

sarbash:
- A description of what you did for the changelog:
I've changed discovery/config.go to correct assignment of parsed values

- An explanation of why ContainerPilot needs this change
Registering in TLS-enabled Consul doesn't work

- How to verify that it works (most PRs need tests!)
Try to register your service in TLS-enabled Consul. It'll fail constantly without this fix.

- A link to the GitHub issue that it addresses
#482

### DIFF
--- a/discovery/config.go
+++ b/discovery/config.go
@@ -40,7 +40,7 @@ func getTLSConfig(parsed *parsedConfig) api.TLSConfig {
 		parsed.TLS.HTTPClientKey = clientKey
 	}
 	if serverName := os.Getenv("CONSUL_TLS_SERVER_NAME"); serverName != "" {
-		parsed.TLS.HTTPClientKey = serverName
+		parsed.TLS.HTTPTLSServerName = serverName
 	}
 	verify := os.Getenv("CONSUL_HTTP_SSL_VERIFY")
 	switch strings.ToLower(verify) {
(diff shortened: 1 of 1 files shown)


tgross:
I'm not wild that we don't have tests for this or a bug report that explains how it "doesn't work" in the first place. But in terms of design this LGTM.

sarbash:
It seems that nobody uses tls with Consul. I've had followed the guide on Consul's site howto run Consul in Docker but microservices are running in internal's Docker network so they have not so much to get access to Consul.

tgross:
> It seems that nobody uses tls with Consul.

I know we had some people asking about it, but the common deployment is having the agent on the same instance as the application. So while encrypting the gossip protocol might be important, there's a lot less value in encrypting the localhost traffic to the agent.

That being said, we do want to make sure it works. Given you added the logs I'm ok with merging this if @cheapRoc gives me the 👍

sarbash:
I'm with You with both my hands regarding the agent maybe that info is old.

>>>
Consul should always be run with --net=host in Docker because Consul's consensus and gossip protocols are sensitive to delays and packet loss, so the extra layers involved with other networking types are usually undesirable and unnecessary.
>>>

[Consul and Docker](https://store.docker.com/images/consul)

tgross:
Ok, this LGTM. Thanks!

sarbash:
Not at all! I'm glad to be of service.

Issue #480, opened 2017-08-16
ContainerPilot.Socket Bind: Address Already in use on restart


Smithx10:
While restarting containerpilot on version 3.3.0 I am receiving the following error 

```
# Add Containerpilot and set its configuration
ENV CONTAINERPILOT_VER=3.3.0
ENV CONTAINERPILOT=/etc/containerpilot.json5
RUN export CP_VER=3.3.0 \
    && export CP_PKG=containerpilot-${CP_VER}.tar.gz \
    && export CP_SHA1=62621712ef6ba755e24805f616096de13e2fd087 \
    && export CP_URL=https://github.com/joyent/containerpilot/releases/download/${CP_VER}/${CP_PKG} \
    && curl -Ls --fail -o /tmp/${CP_PKG} ${CP_URL} \
    && echo "${CP_SHA1} /tmp/${CP_PKG}" | sha1sum -c \
    && tar zxf /tmp/${CP_PKG} -C /usr/local/bin \
    && rm /tmp/${CP_PKG}
```
```
 2017/08/16 17:32:12 error listening to socket at /var/run/containerpilot.socket: listen unix /var/run/containerpilot.socket: bind: address already in use
``` 

```
{
  consul: '{{ if .CONSUL_AGENT }}localhost{{ else }}{{ .CONSUL | default "consul" }}{{ end }}:8500',
  jobs: [
    {
      name: 'preStart',
      exec: '/bin/config-rethinkdb.sh preStart'
      {{ if .CONSUL_AGENT }},
      when: {
        source: 'consul-agent',
        once: 'healthy'
      }{{ end }}
    },
    {
      name: 'rethinkdb',
      port: 28015,
      exec: [
        'rethinkdb',
        '--config-file', '/etc/rethinkdb.conf',
        '--bind', 'all',
        '--no-update-check'
      ],
      health: {
        exec: 'pgrep rethinkdb',
        interval: 10,
        ttl: 25
      },
      when: {
        source: 'preStart',
        once: 'exitSuccess'
      },
      restarts: 'unlimited'
    },{{ if .CONSUL_AGENT }}
    {
      name: 'consul-agent',
      exec: ['/usr/local/bin/consul', 'agent',
                  '-bind={{`{{ GetInterfaceIP \"eth0\" }}`}}',
                  '-data-dir=/data',
                  '-config-dir=/config',
                  '-log-level=err',
                  '-rejoin',
[...]

PR #484, opened 2017-08-17
control: HTTPServer should handle existing control socket files

cheapRoc:
This commit adds support for the control `HTTPServer` to unlink existing socket
files prior to listening. This works around a case where ContainerPilot receives
a `SIGKILL` which prevents it from self cleaning up the socket properly. I've
also abstracted validation into it's own function, called `Validate`, which
ensures the state of `HTTPServer` prior to continuing `control.NewHTTPServer`.

Ref: joyent/containerpilot#480

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -5,6 +5,7 @@ package control
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
 	"net/http"
 	"os"
@@ -17,7 +18,10 @@ import (
 )
 
 // SocketType is the default listener type
-var SocketType = "unix"
+var (
+	SocketType     = "unix"
+	ErrMissingAddr = errors.New("control server not loading due to missing config")
+)
 
 var collector *prometheus.CounterVec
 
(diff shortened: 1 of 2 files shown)


control/control.go
```diff
@@ -41,17 +45,35 @@ type HTTPServer struct {
...
+// socket does not exist prior to setting up the listener (bind).
+func (srv HTTPServer) Validate() error {
+	if srv.Addr == "" {
+		return MissingAddrErr
+	}
+
+	// TODO(justinwr@): maybe we should check file ownership and/or perms
+	if _, err := os.Stat(srv.Addr); err == nil {
+		log.Debugf("control: unlinking previous socket at %s", srv.Addr)
+		if err := os.Remove(srv.Addr); err != nil && os.IsExist(err) {
```
tgross: if `err != nil` but `!os.IsExist(err)` then I think we end up returning `nil` for the validation error, right?

cheapRoc: We've got `os.IsExist(err)` and my hope was that the file would be found and removed under most normal conditions. Do you think if there's any error we should return it, removing the call to `IsExist`?

tgross: Well it we want to wrap the error that's fine. But if, say, the file permissions got changed out from under us we'll get a different error than the one returned by `IsExist`. We probably don't want to simply continue in that case.

cheapRoc: Ok, I see what you mean now. Let me play with a few scenarios.

tgross:
LGTM :+1:

Issue #479, opened 2017-08-16
documentation on stopTimeout field is confusing


JeromeGe:
containerpilot --version 
Version: 3.3.4
GitHash: f7d6212

lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.3 LTS
Release:        14.04
Codename:       trusty

uname -a
Linux 0743e8140205 3.13.0-32-generic #57-Ubuntu SMP Tue Jul 15 03:51:08 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux

``` containerpilot.json5
{
  consul: "127.0.0.1:8500",
  logging: {
    level: "DEBUG",
    format: "text",
  },
  jobs: [
    {
      name: "test-stopTimeout",
      exec: "/root/test",
      stopTimeout: "120s",
      restarts: "unlimited",
    }
  ],
  control: {
    socket: "/var/run/containerpilot.socket"
  }
}
```

test.go
```
package main

import "fmt"
import "os"
import "os/signal"
import "syscall"
import "time"

func main() {
        sigs := make(chan os.Signal, 1)
        done := make(chan bool, 1)
        signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
        go func() {
                fmt.Println("start")
                sig := <-sigs
                fmt.Println(sig)
                for i := 0; i < 30; i++ {
                        time.Sleep(1 * time.Second)
                        fmt.Println(i)
                }
                done <- true
        }()
        <-done
        fmt.Println("exiting")
}
```

```
time="2017-08-16T18:02:03+08:00" level=debug msg="event: {Shutdown global}" 
time="2017-08-16T18:02:03+08:00" level=debug msg="control: stopping control server" 
time="2017-08-16T18:02:03+08:00" level=debug msg="control: completed graceful shutdown of control server" 
time="2017-08-16T18:02:03+08:00" level=debug msg="control: stopped serving at /var/run/containerpilot.socket" 
time="2017-08-16T18:02:03+08:00" level=debug msg="event: {Stopping test-stopTimeout}" 
time="2017-08-16T18:02:03+08:00" level=debug msg="event: {Stopped test-stopTimeout}" 
time="2017-08-16T18:02:03+08:00" level=debug msg="killing all processes in 5 seconds" 
time="2017-08-16T18:02:03+08:00" level=debug msg=test-stopTimeout.term 
[...]

PR #485, opened 2017-08-17
docs: Attempt to articulate better language around `stopTimeout`

cheapRoc:
Ref: #479

Let me know if this clears up the behavior expressed in the docs.


tgross:
This LGTM.

PR #488, opened 2017-08-18
move defer out of loop so as not to leak a closure

tgross:
For https://github.com/joyent/containerpilot/issues/464

This may not be the complete fix for that #464 but it's definitely contributing. We’re creating a `defer` in the `for` loop, but that loop never exits, and so on each tick we just create a new closure. By moving that defer to the top of the goroutine we create it only once.

cc @cheapRoc

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -97,10 +97,10 @@ func getEnvVarNameFromService(service string) string {
 
 // Run starts the application and blocks until finished
 func (a *App) Run() {
+	a.handleSignals()
 	for {
 		a.Bus = events.NewEventBus()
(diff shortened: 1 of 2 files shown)


tgross:
I've added a commit moving `handleSignals` out of the main loop of `App.Run()` because we never shut that down. This would cause us to leak a goroutine every time we reload the configuration (a very tiny leak to be sure but a leak nontheless). h/t to @sean- on that.

cheapRoc:
Now that we're looking for leaks within reload behavior, I'm curious how we're cleaning up the goroutine that `a.ControlServer.Run(a.Bus)` creates.

Issue #327, opened 2017-04-21
Build portability improvements


tgross:
After some offline discussions with @jen20 where he's trying to get ContainerPilot running on SmartOS, I want an issue to collect thoughts on things in our build/test infrastructure that aren't portable across OS's.

So far I've got:
- our Makefile is a GNU Make file, which isn't portable to BSD or Illumos/SmartOS. I know there are difference but I'm not really sure what they are.
- our integration tests run in Docker, which means they aren't portable to any platform that doesn't have Docker.
- we don't have a way of running automated tests in any environment except what TravisCI gives us, which I don't believe includes BSD or SmartOS (although I'd be happy to be wrong here).

jen20:
The last two points I would add to this for discussion when we get round to looking at this are:

- our dependencies are not checked into this repository, meaning we have a build time dependency on glide (currently). This significantly increases cycle time, risks 'left-pad' style issues of missing dependencies due either to mistake/malice or GitHub being down, and prevents repeatable builds from easily being made without internet access from a simple clone of the project.
- our builds and unit tests run in Docker by default and need special casing to run on SmartOS (via the `make local` target.

tgross:
> our dependencies are not checked into this repository

I'm going to table this sub-topic as unrelated to this particular issue. Our plan has been to use whatever `dep` says we need to do once it's been deemed production-ready, so that we're not doing something out-of-the-ordinary from the rest of the golang community.

cheapRoc:
I started looking into the process changes to support building ContainerPilot under SmartOS given the existing build system and tooling. I took some notes...

- `pkgin update && pkgin install go-1.8.3 git unzip gmake`
- I had to comment out [the glide install](https://github.com/joyent/containerpilot/blob/master/makefile#L99-L101) and build manually via `go get github.com/Masterminds/glide && cd $GOPATH && make install`
- The makefile needs to support `go env` values for `GOPATH`, `GOOS`, and `GOARCH` (I'll open a PR for this).
- `glide` still requires `GOPATH` in the environment.
- For whatever reason, `make local build` doesn't remove Docker commands from the rest of the process so I had to comment [a few lines out](https://github.com/joyent/containerpilot/blob/master/makefile#L47L49) for the build to work.
- All unit tests pass 💚 

As a note, I'd be happy to continue down this path and upstream some of my changes if we had a SmartOS/ContainerPilot customer or community member that required it. Please leave a comment on this issue if you are such a user.

tgross:
[...]

PR #489, opened 2017-08-21
tools: Go env vars should be derived from `go env`

cheapRoc:
This simply adjusts a couple small things I found while exploring #327. Makefile changes include...

* Go env vars should be derived from `go env`
* remove the hard restriction on go1.8 during `make tools`
* alpha sort `make debug` env vars

### DIFF
--- a/makefile
+++ b/makefile
@@ -15,8 +15,9 @@ docker := docker run --rm -e LDFLAGS="${LDFLAGS}" $(RUNNER)
 export PATH :=$(PATH):$(GOPATH)/bin
 
 # flags for local development
-GOOS := $(shell uname -s | tr A-Z a-z)
-GOARCH := amd64
+GOPATH ?= $(shell go env GOPATH)
+GOOS ?= $(shell go env GOOS)
+GOARCH ?= $(shell go env GOARCH)
 CGO_ENABLED := 0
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -93,7 +94,7 @@ dep-add: build/containerpilot_build
 # run 'GOOS=darwin make tools' if you're installing on MacOS
 ## set up local dev environment
 tools:
-	@go version | grep 1.8 || (echo 'go1.8 not installed'; exit 1)
+	@go version | grep 1.8 || (echo 'go1.8 should be installed')
```
tgross: Wouldn't this unpin us from go1.8?

cheapRoc: Yes! I've been testing locally under 1.9rc2 and would like to continue moving my local Go version. My hope is that our build process enforces our [Go version](https://github.com/joyent/containerpilot/blob/master/Dockerfile#L1) target and not our local toolchain. I think a developer warning should be enough here.

cheapRoc: Does this change/step have other repercussions I'm not considering?

tgross: Ok, that's fair so long as we're pinning for the build process. The "casual contributor" case I think we're covered for with the Docker builder. I can't think of any other impact, so LGTM 👍

Issue #490, opened 2017-08-22
containerpilot spawns many threads (LWP)


roengram:
containerpilot version: 2.7.7
cloud: samsung private cloud (based on JPC)

I noticed that containerpilot process is creating many LWPs.
Below is the output of */native/usr/bin/prstat*
```
  PID USERNAME  SIZE   RSS STATE  PRI NICE      TIME  CPU PROCESS/NLWP         
 32716 root     1217M 1186M sleep   59    0  11:28:25 0.1% prometheus/65
 32715 root       82M   65M sleep   59    0   1:10:21 0.0% griffin-log-col/63
 32722 root       52M   32M sleep   59    0   1:27:03 0.0% consul/67
 30693 root      247M  196M sleep   54    0   0:01:02 0.0% java/228
 70092 root        0K    0K zombie   0    -   0:00:00 0.0% /0
 37293 root       14M 3536K sleep   59    0   0:00:00 0.0% bash/1
 84507 root       14M 3560K sleep   59    0   0:00:00 0.0% bash/1
 38034 root       14M 3552K sleep   59    0   0:00:00 0.0% bash/1
 76483 root       14M 3588K sleep   59    0   0:00:00 0.0% bash/1
 74903 root       25M 3552K sleep   43    0   0:00:00 0.0% vi/1
 30676 root       14M  408K sleep   59    0   0:00:00 0.0% run-griffin-age/1
 84546 root     4340K 3472K cpu14    1    0   0:00:00 0.0% prstat/1
 30679 root       14M  404K sleep   59    0   0:00:00 0.0% bash/1
 69399 root       14M 3568K sleep   45    0   0:00:00 0.0% bash/1
 32714 root       14M  392K sleep   59    0   0:00:00 0.0% run.sh/1
 32138 root       23M   10M sleep    1    0   0:12:45 0.0% containerpilot/28
 59901 root       14M 3592K sleep   59    0   0:00:00 0.0% bash/1
 42998 root        0K    0K zombie   0    -   0:00:00 0.0% /0
 69398 root       14M 2288K sleep   59    0   0:00:00 0.0% bash/1
 31778 root        0K    0K sleep   60    -   0:00:00 0.0% zsched/1
 31874 root      268M  254M sleep   59    0   1:14:37 0.0% containerpilot/3520

Total: 21 processes, 3984 lwps, load averages: 0.83, 1.44, 1.49
```

[...]

PR #492, opened 2017-08-22
split SIGCHLD from all other signal handlers in supervisor (v2)

tgross:
For https://github.com/joyent/containerpilot/issues/490, in v2

This PR splits out the goroutine that handles the `SIGCHLD` signal from all other signal handlers. This lets us loop on that handler without worrying about spawning additional goroutines for each signal received and should prevent resource growth in the case where many `SIGCHLD` signals land during the handling of one of them.

cc @cheapRoc @roengram


tgross:
For some reason the tasks test is now failing. I would say it's flaky in v2 (which is certainly a possibility) but it's failing in a consistent way. Lemme look into that:

```
2017/08/22 15:30:50 TEST: test_tasks
Creating testtasks_consul_1
Creating testtasks_app_1
Expected task1 to have 7 or 8 executions: got 6
No stopped containers
2017/08/22 15:30:57 FAIL: test_tasks
```

PR #495, opened 2017-08-23
Fix join

### LINKED ISSUE #494: putEnv with more then one equal sign will cut string
Hi,

As usual thank you guys for a great project.
I am facing an issue when trying to use the `putEnv` with a string with multi equal value

here is an example
```bash
container pilot -putenv POSTGRESQL=localhost:5432/test?user=fred&password=secret&ssl=true";
```

The desired behavior would be to assign `localhost:5432/test?user=fred&password=secret&ssl=true"` to `POSTGRESQL`
the it will ony assign `localhost:5432/test?user`

The issue is due to the limit in [indices](https://github.com/joyent/containerpilot/blob/e2fb5ec4a2c0cdbf8e01cf1a85badd97a0800875/core/flags.go#L36)

I can do PR with the following and this should fix the issue.
```
key, val := strings.Join(pair[0:1], ""), strings.Join(pair[1:], "="))
```

Waiting for your feedback.

ahelal:
Fixes #494

### DIFF
--- a/core/flags.go
+++ b/core/flags.go
@@ -33,7 +33,7 @@ func (f *MultiFlag) Set(value string) error {
 		return fmt.Errorf(
 			"flag value '%v' was not in the format 'key=val'", value)
 	}
-	key, val := strings.Join(pair[0:1], ""), strings.Join(pair[1:2], "")
+	key, val := strings.Join(pair[0:1], ""), strings.Join(pair[1:], "=")
 	f.Values[key] = val
 	return nil
 }
--- a/core/flags.go
+++ b/core/flags.go
@@ -28,13 +28,11 @@ func (f *MultiFlag) Set(value string) error {
 	if f.Len() == 0 {
 		f.Values = make(map[string]string, 1)
 	}
-	pair := strings.Split(value, "=")
+	pair := strings.SplitN(value, "=", 2)
 	if len(pair) < 2 {
-		return fmt.Errorf(
-			"flag value '%v' was not in the format 'key=val'", value)
+		return fmt.Errorf("flag value '%v' was not in the format 'key=val'", value)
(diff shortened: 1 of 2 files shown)


core/flags.go
```diff
@@ -33,7 +33,7 @@ func (f *MultiFlag) Set(value string) error {
 		return fmt.Errorf(
 			"flag value '%v' was not in the format 'key=val'", value)
 	}
-	key, val := strings.Join(pair[0:1], ""), strings.Join(pair[1:2], "")
+	key, val := strings.Join(pair[0:1], ""), strings.Join(pair[1:], "=")
```
tgross: With fresh eyes the original line here looks questionable to me and I'm not sure what it's supposed to be getting us. Couldn't we do?:

```golang
func (f *MultiFlag) Set(value string) error {
 	if f.Len() == 0 {
 		f.Values = make(map[string]string, 1)
 	}
	pair := strings.SplitN(value, "=", 2)
	if len(pair) < 2 {
		return fmt.Errorf("flag value '%v' was not in the format 'key=val'", value)
	}
	f.Values[pair[0]] = pair[1]
	return nil
}
```

Do we have an example where the original line does something we need? It'd be nice to have a unit test for this.

ahelal:
Hi @tgross,

I update the PR. I attempted to add a test, but was not sure I am in the right direction.
I wanted to check for `-putenv ENV_VALUE=PART1` and `-putenv ENV_VALUEPART1` to cover all cases. I kind of have the urge to test `Set` directly instead of using `os.Args and then GetArgs()`

But feeling lost somehow. So your input would be helpful.

tgross:
I think this LGTM. Let's get it in!

PR #496, opened 2017-08-23
Refactor shutdown with new pub/sub interfaces ~ WIP

cheapRoc:
**\*WIP\*** This is highly in flux, no tests, and I've broken everything. I'm working through the interfaces to see where I need to focus and make changes so I needed review as I step through this.

Included...
- Passing `context` up from `app.go`.
- Stubbed out new `Publisher` and `Subscriber` under `events/`.
- I've begun integrating `Publisher` and `Subscriber` into `Job`.
- I think I've fully covered `Publisher` into `Watch` (minus refactoring out Timeout event).
- Working `Subscriber` into `Metric`.
- Ripping `EventHandler` out (it isn't used anymore).
- Passing `context` into the control and telemetry servers and handling shutdown there.

Missing...
- Control and Telemetry reloading behavior.
- Control bus for publishing Endpoint events.
- Test changes to accommodate everything.

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -11,10 +11,9 @@ import (
 	"os"
 	"time"
 
+	"github.com/joyent/containerpilot/events"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
-
-	"github.com/joyent/containerpilot/events"
 )
 
 // SocketType is the default listener type
@@ -37,9 +36,9 @@ func init() {
 // HTTP transport control plane. Currently this is listening via a UNIX socket
 // file.
 type HTTPServer struct {
+	Addr string
+
 	http.Server
-	Addr                string
-	events.EventHandler // Event handling
 }
 
 // NewHTTPServer initializes a new control server for manipulating
@@ -52,7 +51,7 @@ func NewHTTPServer(cfg *Config) (*HTTPServer, error) {
 		return nil, fmt.Errorf("control: validate failed with %s", err)
 	}
 
-	srv.InitRx()
+	// srv.InitRx()
 	return srv, nil
 }
 
@@ -73,15 +72,15 @@ func (srv HTTPServer) Validate() error {
 }
 
 // Run executes the event loop for the control server
-func (srv *HTTPServer) Run(ctx context.Context) {
+func (srv *HTTPServer) Run(ctx context.Context, bus *events.EventBus) {
 	defer srv.Stop(ctx)
-	srv.Start()
+	srv.Start(bus)
 }
 
 // Start sets up API routes with the event bus, listens on the control
 // socket, and serves the HTTP server.
-func (srv *HTTPServer) Start() {
-	endpoints := &Endpoints{srv.Bus}
+func (srv *HTTPServer) Start(bus *events.EventBus) {
+	endpoints := &Endpoints{bus}
 
 	router := http.NewServeMux()
 	router.Handle("/v3/environ", PostHandler(endpoints.PutEnviron))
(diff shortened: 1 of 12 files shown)


events/bus.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // EventBus manages the state of and transmits messages to all its Subscribers
 type EventBus struct {
-	registry map[Subscriber]bool
+	registry map[*Subscriber]bool
```
cheapRoc: One of the things I'm not sure about was this change. I had to make it in order for the compiler to pass once I started passing `EventPublisher` through other changes, but, admittedly, I don't fully understand the repercussions here.

events/handler.go
```diff
@@ -30,17 +30,17 @@ func (evh *EventHandler) InitRx() {
 
 // Subscribe adds the EventHandler to the list of handlers that
 // receive all messages from the EventBus.
-func (evh *EventHandler) Subscribe(bus *EventBus, isInternal ...bool) {
+func (evh *EventHandler) Subscribe(bus *EventBus) {
 	evh.wg.Add(1)
-	bus.Register(evh, isInternal...)
+	bus.Subscribe(evh)
```
cheapRoc: Only to satisfy the compiler here...

events/bus.go
cheapRoc: ...beyond the fact that its now a reference and not an instance.

tgross: `Subscriber` was an interface, not an instance. You can't take a pointer to an interface, so when we switched here to using a concrete object we want to make sure that's a reference so that it can unsubscribe itself later.

jobs/jobs.go
```diff
@@ -121,37 +124,37 @@ func (job *Job) Kill() {
 }
 
 // Run executes the event loop for the Job
-func (job *Job) Run() {
-	ctx, cancel := context.WithCancel(context.Background())
+func (job *Job) Run(ctx context.Context) {
```
tgross: Just a style note: rather than having a `ctx` and `ctx2`, I like using variable names that make it clear which is a parent and which is the child. So I think elsewhere you'll see I've used `pctx` for "parent context", but if you're not wild about that name you at least get the general idea.

control/control.go
```diff
@@ -73,19 +72,14 @@ func (srv HTTPServer) Validate() error {
...
 
 	go func() {
-		defer srv.Stop()
+		defer srv.Stop(ctx)
 		for {
-			event := <-srv.Rx
-			switch event {
-			case
-				events.QuitByClose,
-				events.GlobalShutdown:
```
cheapRoc: We won't need this anymore.

control/control.go
```diff
@@ -94,8 +88,8 @@ func (srv *HTTPServer) Run(bus *events.EventBus) {
 
 // Start sets up API routes with the event bus, listens on the control
 // socket, and serves the HTTP server.
-func (srv *HTTPServer) Start() {
-	endpoints := &Endpoints{srv.Bus}
+func (srv *HTTPServer) Start(bus *events.EventBus) {
+	endpoints := &Endpoints{bus}
```
cheapRoc: ...but I'm wondering if we'll need to still treat HTTPServer as a publisher for `Endpoints` who publish events. Unsure how to handle reload through the control socket now as well, it's possible I'll need to add that back.

control/control.go
tgross: Because we don't need it we can actually replace the whole goroutine with:

```
go func() {
	 defer srv.Stop(ctx)
	 <- ctx.Done()
	 return
}()
```

control/control.go
tgross: Ah, that's a good point. I'd forgotten that we publish events through the endpoints.

cheapRoc: Hopefully we can look at this together tomorrow because I'm unsure what benefit we're getting by this refactor. I think we both had hoped this change set would remove a lot of cruft around the server sections, which I'm not certain of now.

cheapRoc:
Overall, this refactoring has come together for Jobs, Watches, and Metrics. Control and Telemetry servers are another issue and unit tests need to be cleaned up dramatically.

I'm going to squash this PR and open another tomorrow.

cheapRoc:
Forgot to close this old working branch.

Issue #476, opened 2017-08-15
split publish/subscribe behavior to better-leverage context


tgross:
Cancelling jobs, watches, servers, etc. has been a bit of a bug farm. We don't really leverage `context` as much as we probably could, but any use of this needs to make sure we're sending messages back like stopping/stopped and deregistering from the event bus.

I've worked up a preliminary design that splits the `EventHandler` into separate `Publisher` and `Subscriber` types and uses `context.CancelFunc` as a means of signaling shutdown instead of publishing an event. This _seems_ to work ok and I think will be subtly simpler in the long run.

Here's the worked demo: https://gist.github.com/tgross/6b0e32a85157c045718ee2ef5d3a8da6
If you `go run main.go 1000` this will run for 1000ms and then exit cleanly. It has the pub/sub types, an implementation of watch (pub-only), metric (sub-only), and job (pub+sub).

cc @cheapRoc

tgross:
Updated the worked demo example here: https://gist.github.com/tgross/b167a2b1d4a5b727643d6e25f9962553

This one handles signals (SIGTERM and SIGINT, plus SIGHUP for reloads) rather than having a fixed time interval.

cheapRoc:
There's a lot of work sitting behind this and I definitely feel the urge to complete what I've put in so far. 

At the moment I'm stuck attempting to uncover the reason why CP is not exiting when all jobs have completed cleanly. The primary integration test where this was uncovered is running the following (`test_envvars`)...

```json5
{
  consul: "127.0.0.1:8500",
  jobs: [
    {
      name: "testenvvar",
      port: 8500,
      exec: "env",
      health: {
        exec: "true",
        interval: 10,
        ttl: 25
      }
    }
  ]
}
```

This will halt execution and exits cleanly on master so I know the call stack the bug is in but not the specific code that's causing it. When this fires on my branch, it's still blocking on the `WaitGroup`, which is puzzling since de-registration should have occurred for the `Job` after the `Command` exits.

Still in flux.

cheapRoc:
Ok, just as a note, it looks like the control server is not being shut down when a job exits cleanly/successfully. This makes sense because the existing implementation doesn't `Unregister` itself from the EventBus. Now we're getting somewhere...

cheapRoc:
Another note, I'm on a good path now and down to a single failing integration test, `test_coprocess`.

cheapRoc:
[...]

Issue #497, opened 2017-08-24
Display the PID of a job's exec process as an additional logger field


cheapRoc:
I'm proposing a feature that would add an additional `pid` logger field of a job's exec'd process. It looks something like this...

```
INFO[0005] 'this is test'                                job=echo-test pid=8346
```

The goal is to delineate between various processes on a system under ContainerPilot's supervision. This also helps display changes in process ID's during reloads, etc.

I've already implemented this so this issue is a stub until I can post the diff (it's inline with another feature).

tgross:
I like this idea! 👍

ahelal:
what about the job name ?

tgross:
> what about the job name ?

That's already included. ref [`jobs/config.go#L253`](https://github.com/joyent/containerpilot/blob/c326edae9cf73151e264f590330a6639586a3fb0/jobs/config.go#L253)

cheapRoc:
This is next up on the list now that #512 has been merged. Look for a new PR soon.

PR #498, opened 2017-08-24
Tag logger field with the PID of a Command's exec process

cheapRoc:
Ref: #497 

This adds a Command's process PID as a logger field which displays the PID as a tag in the logs. To implement this I've added a new attribute on the Command struct called `fields` which passes around the state of the logger fields in case they need to be reset (as we do here).

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -25,6 +25,7 @@ type Command struct {
 	Timeout time.Duration
 	logger  log.Entry
 	lock    *sync.Mutex
+	fields  log.Fields
 }
 
 // NewCommand parses JSON config into a Command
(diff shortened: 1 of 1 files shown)


commands/commands.go
```diff
@@ -99,6 +101,19 @@ func (c *Command) Run(pctx context.Context, bus *events.EventBus) {
...
+		// if we're able to, log the PID of our Command's exec process through
+		// our logger fields
+		if c.Cmd != nil && c.Cmd.Process != nil {
+			pid := c.Cmd.Process.Pid
+			if len(c.fields) > 0 {
+				c.fields["pid"] = pid
+			} else {
+				c.fields = log.Fields{"pid": pid}
+			}
+			c.logger = *log.WithFields(c.fields)
```
tgross: I'm pretty sure this breaks the "raw" logging, doesn't it? I think we only want to assign to `c.logger` in the case where `c.fields > 0`, so this line should be in first branch of the `if` above.

tgross: Actually, we attach to the process stdout/stderr well before this part -- so does this work? We should definitely verify the pid field shows up in the `test_logging` integration test.

cheapRoc: You're right, and I completely missed the comment about that.

cheapRoc:
This will be reviewed and merged after I close #476 (or #499) since there was a race discovered that became more prominent after this was added (#500).

cheapRoc:
Since I haven't had the time to complete it's dependencies I'm going to close this for the moment and revisit later.

cheapRoc:
This is next up in line now that #512 has been merged. I'll have a PR open soon.

PR #499, opened 2017-08-24
Refactor shutdown and reload using pub/sub break-up ~ WIP

cheapRoc:
**\*WIP\*** I've squashed my existing branch and continue to make progress on the new shutdown and reload behavior using `context`. There's an issue where the control server is not triggering `Publish` properly in order to handle reloads but signal processed reloads are reloading control server properly. Progress...

Peer review is highly appreciated.


cheapRoc:
Today ended up going a lot better than I had hoped despite a few set backs late yesterday. The old reload behavior is now entirely coupled to the application wide context chain. The only sketchy part is that the reload control endpoint calls context.CancelFunc in order to reload the control server. While this centralizes the shutdown behavior, this is stored on the Endpoints struct which is hopefully GC'd properly.

Telemetry needs a similar refactor then I can clean up tests.

cheapRoc:
Telemetry is cleaned up and reloads properly. I've also cleaned up more of the tests but this one looks unsettling...

```
--- FAIL: TestJobRunSafeClose (10.11s)
        jobs_test.go:44: expected: [{Startup global} {Stopping myjob} {Stopped myjob}]
                got: [{Startup global} {ExitSuccess myjob} {Stopping myjob} {Stopped myjob}]
```

I also have a hang at `TestJobMaintenance/enter_maintenance`.

cheapRoc:
While all lints and unit tests are passing, integrations are stalling further progress and will need to be debugged. In an effort to remain in good standing with my family, this will have to wait until I'm back from vacation. Thus we'll close this for the time being.

PR #508, opened 2017-09-19
add Jobs to status endpoint

### LINKED ISSUE #507: Include jobs with health checks in /status
In the services array we should include any jobs that have a health check. This will mean that they may not have an address or port, but they can at least report on their health status.

For example, given the following jobs config:

```
jobs: [
    {
      name: 'foo',
      exec: 'node /opt/app/',
      restarts: 'unlimited',
      health: {
        exec: 'pgrep node',
        interval: 10,
        ttl: 20
      }
    }
]
```

The current `/status` endpoint doesn't return any values in the `Services` array for the above job, instead, we should return:

```
Services: [{ Name: 'foo', Status: 'healthy' }]
```

This will be useful for checking the status of services that aren't registered with consul but still have health checks.

geek:
Fixes #507

### DIFF
--- a/jobs/config.go
+++ b/jobs/config.go
@@ -127,10 +127,8 @@ func (cfg *Config) Validate(disc discovery.Backend) error {
 	if err := cfg.validateRestarts(); err != nil {
 		return err
 	}
-	if err := cfg.validateExec(); err != nil {
-		return err
-	}
-	return nil
+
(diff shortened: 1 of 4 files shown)


geek:
lint failure is unrelated to any change I made:

```
jobs/config.go:130:2: redundant if ...; err != nil check, just return error instead.
```

PR #512, opened 2017-10-05
Refactor away EventHandler into separate pub/sub interfaces (#476)

cheapRoc:
This commit refactors away EventHandler to replace it with separate Publisher/Subscriber interfaces. The side effect is properly using the `context` package throughout ContainerPilot as the basis for it's shutdown process. This helps provide a more structured shutdown and no longer relies on the event bus and event processing which was bug prone.

Changes include...

* Introduce more `context` handling throughout our shutdown processes
* Introduce Subscriber and Publisher interfaces
* Turn `Job` into a Publisher and Subscriber
* Turn `Watch` into a Publisher
* Turn `Metric` into a Subscriber
* Refactor Control and Telemetry server shutdown/reload
* Feed a `context.CancelFunc` up stack into the reload control endpoint
* Telemetry server reload uses an appropriate `context.CancelFunc`
* Fix all broken unit tests for pub/sub refactoring
* Patch tests around pub/sub to use `QuitByClose` event for testing only
* Reuse `QuitByTest` event for shutting down through internal control channel
* Shut down CP after all Jobs have completed

I'm posting this up to begin RC testing. There's a lot to confirm but all unit and integration tests are passing for me locally and through Travis. I want to sit on this while I test and hunt down code reviewers.


core/app.go
```diff
@@ -97,11 +98,42 @@ func getEnvVarNameFromService(service string) string {
...
+		ctx, cancel := context.WithCancel(context.Background())
+
+		// This provides an escape hatch for shutting down after all jobs have
+		// completed. Each time a job completes, during its cleanup func, it
+		// will be set `IsComplete` to `true`. Then it'll send `true` through
+		// this channel to wake up this goroutine, check all jobs for
+		// `IsComplete`, and shutdown ContainerPilot if all jobs are
+		// completed. This is because ContainerPilot is NOT a server and must
+		// shut down when no longer required (i.e. process containers).
+		completedCh := make(chan bool)
```
cheapRoc: This is the part I'm unhappy about but I don't believe there is any other way around it.

Every other case for using `context` throughout ContainerPilot allows the application to uniformly shutdown when calling `context.CancelFunc` which is passed into various parts of the application. This works great for the Reload endpoint or SIGTERM handling where `context` can be used by a single vector for termination.

This is not the case when handling shutdown when all Jobs have exited.

All Jobs running in ContainerPilot can cause a shutdown upon completion if no other jobs are running. This means we need to check the status of every job before calling `context.CancelFunc`. This goroutine provides that behavior for every Job that is exiting by providing a channel back to a thread which can properly exit ContainerPilot.

jobs/jobs.go
```diff
@@ -70,8 +75,9 @@ func NewJob(cfg *Config) *Job {
 		restartsRemain:    cfg.restartLimit,
 		frequency:         cfg.freqInterval,
 	}
-	job.InitRx()
+	// job.InitRx()
```
geek: can just delete this line

jobs/jobs_test.go
```diff
@@ -43,22 +50,24 @@ func TestJobRunSafeClose(t *testing.T) {
...
 
 	time.Sleep(200 * time.Millisecond)
-	defer func() {
-		if r := recover(); r != nil {
-			t.Fatalf("panicked but should not: sent to closed Subscriber")
-		}
-	}()
-	bus.Publish(events.QuitByClose)
-	job.Quit()
+	// defer func() {
```
geek: can delete these lines

jobs/jobs.go
cheapRoc: Yes, I need to remove these. There might be another somewhere.

watches/watches.go
```diff
@@ -53,45 +56,46 @@ func (watch *Watch) CheckForUpstreamChanges() (bool, bool) {
...
-	watch.Bus = bus
-	ctx, cancel := context.WithCancel(context.Background())
+func (watch *Watch) Run(pctx context.Context, bus *events.EventBus) {
+	watch.Register(bus)
+	ctx, cancel := context.WithCancel(pctx)
 
-	timerSource := fmt.Sprintf("%s.poll", watch.Name)
-	events.NewEventTimer(ctx, watch.Rx,
-		time.Duration(watch.poll)*time.Second, timerSource)
+	// timerSource := fmt.Sprintf("%s.poll", watch.Name)
```
cheapRoc: Removing this one as well.

cheapRoc:
Will get this merged in now that I've received CR from both @geek and @stack72 (offline).

PR #516, opened 2017-10-11
Log the pid of every job as in a logger field (#497)

cheapRoc:
Ref: #497 

This re-introduces my PR for adding a job's PID to log output. Before the shutdown refactor landed this was causing a race condition and blocking the socket from closing properly (because the old shut down was shotty). This is working for me without issue now.


cheapRoc:
@stack72 CR here as well, thanks.

stack72:
Also LGTM! nice addition

Issue #515, opened 2017-10-10
Migrate away from deprecated PassTTL and any other deps


cheapRoc:
Quick issue for migrating away from Consul's API `PassTTL` to the new favored `UpdateTTL` function. This is primarily found underneath the `discovery` package.

We should also take the moment to find out if there are any other deprecations or dependencies we can move ahead.

PR #517, opened 2017-10-11
Migrate from deprecated Consul API call PassTTL to UpdateTTL (#515)

cheapRoc:
Ref: #515 

This starts a PR fixes the Consul API deprecation and begins exploring what other deprecations/dependencies we can clean up.

~~FTM we'll keep this open.~~

### DIFF
--- a/discovery/consul.go
+++ b/discovery/consul.go
@@ -57,10 +57,10 @@ func NewConsul(config interface{}) (*Consul, error) {
 	return consul, nil
 }
 
-// PassTTL wraps the Consul.Agent's PassTTL method, and is used to set a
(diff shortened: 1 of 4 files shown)


cheapRoc:
There are no other dep warnings/changes that I could find. @stack72 if you want to CR that would be great.

stack72:
Apart from the suggestion of using constants, this LGTM :)

cheapRoc:
If you're talking about replacing the strings passed into `UpdateTTL` with constants, than yes I agree. But there may be a feature proposal I'll make regarding that line in the not too distant future.

Issue #513, opened 2017-10-05
Trigger a job by specific UNIX signal hitting CP process


cheapRoc:
The goal of this enhancement is to re-introduce some of the behavior available in ContainerPilot v2 which is now missing in ContainerPilot v3.x. This includes the common case of needing to communicate a specific signal to processes under CP's control. Sending a UNIX signal into a main process running within the container is something a parent scheduler might make use of (like Nomad).

This extends some of the behavior exposed by #504 and #505.

## Proposal

ContainerPilot should be enhanced to handle specific UNIX signals as special task events that can subsequently trigger a job or command within the container. Let's call this a _signal event_.

The most effective solution should be for ContainerPilot to propagate UNIX signals into a job by reusing job triggers that exist today. This removes the need for ContainerPilot to re-broadcast maintenance/custom UNIX signals to children under its control, thus causing potential odd edge cases and scenarios for non-standard software. A signal event also promotes ever more customization for end users of ContainerPilot which can perform any action they please for handled signals.

### Option 1

- The supervisor running as PID 1 will now include `SIGHUP` and `SIGUSR2` handlers.
- The supervisor will also include code that allows it to pass signals through to the worker process.
- The worker will also handle both `SIGHUP` and `SIGUSR2`.
- The signal handler will publish a global event, `events.Signal`, onto the bus similar to how `events.GlobalShutdown` is implemented today.
- A `job.when` clause can be configured to detect this global event and fire a script when a signal event occurs.
- The watching job's exec will be able to call `kill -HUP` or perform any other process within the container when it is fired by this signal event.

### Option 2

- The supervisor running as PID 1 will now include `SIGHUP` and `SIGUSR2` handlers.
- The supervisor will also include code that allows it to connect to the control socket. This will work as a primary mode of IPC.
- The control socket will include a write only `POST /signal` handler which receives a POST from the supervisor when it receives a `SIGHUP` or `SIGUSR2`.
- The control socket handler will publish a global event, `events.Signal`, onto the bus similar to how `events.GlobalShutdown` is implemented today.
- A `job.when` clause can be configured to detect this global event and fire a script when a signal event occurs.
- The watching job's `exec` will be able to call `kill -HUP` or perform any other process within the container when it is fired by this signal event.

Initially, the goal will be to target the implementation at the supervisor process of CP then shortly after refactor this out into the main process not running as PID 1. The control socket behavior should not be required when CP is not running as PID 1 since there is only a single process.

[...]

PR #520, opened 2017-10-17
Implement new signal based events

cheapRoc:
Ref: #513 

This PR implements the process outlined in the tracking issue. I've also included the ability for running jobs to create an environment variable including their PID (`CONTAINERPILOT_<name>_PID`) as well as both the supervisor and normal worker process signal handlers. So far, the functionality is working out great and you can fire jobs based on a container receiving a specific signal (or CP process itself). Obviously this is more of a scheduler based feature for CP and not something useful for the Triton audience.

I'm still testing this and looking to make sure nothing that I've changed is going to cause havoc to the supervisor process.


cheapRoc:
I'm taking my time testing this since I'll be introducing behavior into the supervisor process. I already have plans to change my original implementation under this WIP PR. I've found some optimizations but also weird behavior (dup signals passing).

cheapRoc:
Latest commits include fixes from master as well as simplifying the supervisor logic. I'll perform final testing tomorrow.

cheapRoc:
Alright, I think this is baked enough. Couple things to note...
- I reverted the control endpoint for triggering signal events over HTTP. A part of me is wondering if this would have been handy either way, as one could trigger a custom event within CP running on Triton.
- I've added some documentation.
I'll get CR of this next week.

events/events_test.go
```diff
@@ -23,51 +28,86 @@ type TestSubscriber struct {
...
 		results: []Event{},
 	}
-	my.Rx = make(chan Event, 100)
-	return my
+	sub.Rx = make(chan Event, 100)
+	return sub
 }
 
-func (ts *TestSubscriber) Run(bus *EventBus) {
+func (ts *TestSubscriber) Run(ctx context.Context, bus *EventBus) {
```
cjihrig: Any reason not to make `ctx` a pointer?

docs/30-configuration/34-jobs.md
```diff
@@ -29,6 +29,15 @@ Additionally, jobs may react to these events:
...
 - `exitMaintenance`: published when the [control plane](./30-configuration/37-control-plane.md) is told to exit maintenance mode for the container.
 
+Finally, there are two special `source` values that can be used to trigger a job when ContainerPilot receives a UNIX signal.
+
+- `SIGHUP`: published when a ContainerPilot process receives the UNIX signal `SIGHUP`.
+- `SIGUSR2`: published when a ContainerPilot process receives the UNIX signal `SIGUSR2`.
+
+Signal events come in handy when you need to kick off some type of special process (reloading configs, publishing debug info) inside a container. This type of external communication is only supported when running ContainerPilot within a Docker container running on a Docker host, or under the supervision of a scheduler like Nomad.
+
+Note: Either two signals can be sent to ContainerPilot acting as a PID 1 supervisor or it's standalone worker process.
```
cjihrig: Should "it's" be "its"?

cheapRoc: Yup, definitely should. 😁

events/events_test.go
cheapRoc: We're not using a pointer because the value never changes. We always want to share a static `context` across the entire application.

It's also worth noting the following documentation from the standard library `context` package...

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each 
> function that needs it. The Context should be the first parameter, typically named `ctx`.

So, while we can technically store this `context.Context` value within a struct (and I've seen plenty of [folks do this](https://github.com/joyent/pg_prefaulter/blob/master/agent/agent.go#L46)), it's not recommended. I can't give you a good reason yet why it's not. I'm on the fence, it always looks handy to store the `context.Context` on a struct.

cheapRoc:
Going to test this one last time and get this merged today.

cheapRoc:
Going to close this PR, do the squash manually, and open a second for the merge.

Issue #521, opened 2017-10-18
Reloads are leaking goroutines


cheapRoc:
Outputting `runtime.NumGoroutine()` at the beginning and end of the reload cycle displays an ever increasing number of goroutines. Grabbing a stack trace of all goroutines shows the following two areas that need work (the other 3 goroutines not shown do NOT increase over reloads).

```
goroutine profile: total 12
6 @ 0x102ec5c 0x102ed4e 0x1040f00 0x10610d0 0x1065966 0x106605c 0x11049bf 0x114a87b 0x105aef1
#       0x1040eff       sync.runtime_notifyListWait+0x10f                                                               /usr/local/Cellar/go/1.9.1/libexec/src/runtime/sema.go:507
#       0x10610cf       sync.(*Cond).Wait+0x7f                                                                          /usr/local/Cellar/go/1.9.1/libexec/src/sync/cond.go:56
#       0x1065965       io.(*pipe).read+0xc5                                                                            /usr/local/Cellar/go/1.9.1/libexec/src/io/pipe.go:47
#       0x106605b       io.(*PipeReader).Read+0x4b                                                                      /usr/local/Cellar/go/1.9.1/libexec/src/io/pipe.go:130
#       0x11049be       bufio.(*Scanner).Scan+0xae                                                                      /usr/local/Cellar/go/1.9.1/libexec/src/bufio/scan.go:207
#       0x114a87a       github.com/joyent/containerpilot/vendor/github.com/sirupsen/logrus.(*Entry).writerScanner+0xaa  /Users/justinreagor/go/src/github.com/joyent/containerpilot/vendor/github.com/sirupsen/logrus/writer.go:51

3 @ 0x102ec5c 0x102ed4e 0x1006584 0x100626b 0x13c432b 0x105aef1
#       0x13c432a       github.com/joyent/containerpilot/core.(*App).handleSignals.func1+0x4a   /Users/justinreagor/go/src/github.com/joyent/containerpilot/core/signals.go:15
```

The latter is a bug in the signal handling path and was definitely introduced by me at v3.5.0. This should be easy to fix. 

I'm not sure yet why the former notes `logrus` and increases by 2 goroutines every reload. I'll continue reviewing.

cheapRoc:
[...]

PR #523, opened 2017-10-18
Fix a goroutine leak in the signal handler code path

cheapRoc:
The root cause of this issue was regarding our change to shutting down via
context. While implementating that change set there was a goroutine leak
introduced that did not provide a clean way for shutting down and reloading the
signal handler. This caused signal handling to be duplicated across reloads.

Fixes: #521

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -130,7 +130,7 @@ func (a *App) Run() {
 		}()
 
 		a.Bus = events.NewEventBus()
-		a.handleSignals(cancel)
+		a.handleSignals(ctx, cancel)
 		a.ControlServer.Run(ctx, a.Bus)
 		a.runTasks(ctx, completedCh)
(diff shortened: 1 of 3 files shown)


cheapRoc:
Looks like the postStop job of the `test_sigterm` integration test is not firing. Looking into it...

cheapRoc:
I'm guessing the removal of cancel effected telemetry. I'll revisit.

cheapRoc:
I'm optimizing this a bit further now that I've noticed we no longer need this in the [main application event loop](https://github.com/joyent/containerpilot/blob/master/core/app.go#L133).

cheapRoc:
I think this one is GTG @stack72.

stack72:
LGTM - we discussed this outside of github already

Issue #519, opened 2017-10-17
Migrate to Go 1.9.x and Consul 1.0.x


cheapRoc:
Stub for migrating our build steps and releases to Go 1.9.x.

cheapRoc:
Ok, so I have all tests passing. Interesting side bit... when I bumped Consul to 1.0.0 and made a few small adjustments to how our tests run, it looks like Consul's `testutil` package was broken. Upon further investigation I found that @magiconair and others have made some required fixes into Consul post-1.0.0 for the testing facilities that ContainerPilot also relies on (re: `freeport`). Besides our "unit" tests (which rely on a live consul server) all integration tests pass so I've pegged my feature branch at v1.0.1-rc1 for now. I'll continue tracking the progress until the final GM is released.

/cc @misterbisson

magiconair:
@cheapRoc `testutil` wasn't broken since it is an internal package of consul that is required for consul's own integration testing and you shouldn't rely on that to be stable. The nomad team ran into this issue as well. 

To speed up integration tests and make them more reliable I've initially added an external tool `porter` to provide tests with unused port numbers on the localhost interface. In a subsequent step we replaced that with the `freeport` library to remove the dependency on the `porter` tool.

Having said that, if you need a consul instance for your tests you should start one by executing the binary and control the process. I've significantly sped up the startup time for a `-dev` instance to be ~100ms and added a `-hcl` command line option with 1.0.0 so that you can provide arbitrary configuration on the command line without generating a config file since there is no parity between config flags and config files.

Just run `consul agent -dev -hcl '<your HCL config>'` and you're set.

cheapRoc:
> @cheapRoc testutil wasn't broken since it is an internal package of consul that is required for consul's own integration testing and you shouldn't rely on that to be stable. The nomad team ran into this issue as well.

You're correct, I should have phrased that a tad better, it was broken in our use case. The interface worked perfectly fine outside Consul in another test of mine. I also noticed the Nomad team's commits and your added context makes this much clearer.

> Having said that, if you need a consul instance for your tests you should start one by executing the binary and control the process. I've significantly sped up the startup time for a -dev instance to be ~100ms and added a -hcl command line option with 1.0.0 so that you can provide arbitrary configuration on the command line without generating a config file since there is no parity between config flags and config files.
>
> Just run consul agent -dev -hcl '<your HCL config>' and you're set.

I'm with you, we shouldn't depend on an internal testing utility of a third party dependency. 

[...]

PR #527, opened 2017-11-12
Upgrade ContainerPilot to Go 1.9.x and Consul 1.0.x

cheapRoc:
Ref: #519 

This is a WIP PR. 

EDIT: Note that Consul is pegged at 1.0.1-rc1 and we'll want to finalize once a GM is released.

### DIFF
--- a/discovery/consul_test.go
+++ b/discovery/consul_test.go
@@ -3,9 +3,9 @@ package discovery
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	consul "github.com/hashicorp/consul/api"
-	"github.com/hashicorp/consul/testutil"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -77,18 +77,18 @@ a Consul server for testing. The 'consul' binary must be in the $PATH
 ref https://github.com/hashicorp/consul/tree/master/testutil
 */
 
-var testServer *testutil.TestServer
+var testServer *TestServer
 
 func TestWithConsul(t *testing.T) {
 	var err error
-	testServer, err = testutil.NewTestServerConfigT(t, func(c *testutil.TestServerConfig) {
-		c.LogLevel = "err"
-	})
+	testServer, err = NewTestServer(8500)
 	if err != nil {
 		t.Fatal(err)
 	}
-
 	defer testServer.Stop()
+
+	time.Sleep(300 * time.Millisecond)
+
 	t.Run("TestConsulTTLPass", testConsulTTLPass)
 	t.Run("TestConsulReregister", testConsulReregister)
 	t.Run("TestConsulCheckForChanges", testConsulCheckForChanges)
--- a/discovery/consul_test.go
+++ b/discovery/consul_test.go
@@ -80,9 +80,14 @@ ref https://github.com/hashicorp/consul/tree/master/testutil
 var testServer *testutil.TestServer
 
 func TestWithConsul(t *testing.T) {
-	testServer, _ = testutil.NewTestServerConfigT(t, func(c *testutil.TestServerConfig) {
+	var err error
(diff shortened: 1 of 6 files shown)


discovery/consul_test.go
```diff
@@ -77,13 +77,18 @@ a Consul server for testing. The 'consul' binary must be in the $PATH
...
-		c.LogLevel = "err"
-	})
+	var err error
+	testServer, err = NewTestServer(8500)
+	if err != nil {
+		t.Fatal(err)
+	}
 	defer testServer.Stop()
+
+	time.Sleep(300 * time.Millisecond)
```
cheapRoc: I'm not crazy about this and would much rather add something that checks for Consul's HTTP API endpoint being live. Any suggestions on that?

cheapRoc:
@magiconair Last commit in this PR adds the exec for `consul`. Let me know any suggestions you have.

Also, I didn't yet remove the old dependencies from our dep manager. I'll do that after we've discussed the patch.

discovery/test_server.go
```diff
@@ -0,0 +1,55 @@
...
+
+// NewTestServer constructs a new TestServer by including the httpPort as well.
+func NewTestServer(httpPort int) (*TestServer, error) {
+	path, err := exec.LookPath("consul")
+	if err != nil || path == "" {
+		return nil, fmt.Errorf("consul not found on $PATH - download and install " +
+			"consul or skip this test")
+	}
+
+	args := []string{"agent", "-dev"}
```
magiconair: You're not configuring the port. Add `"-http-port", strconv.Itoa(httpPort)` to `args`

discovery/test_server.go
```diff
@@ -0,0 +1,55 @@
...
+	cmd := exec.Command("consul", args...)
+	cmd.Stdout = io.Writer(os.Stdout)
+	cmd.Stderr = io.Writer(os.Stderr)
+	if err := cmd.Start(); err != nil {
+		return nil, errors.New("failed starting command")
+	}
+
+	httpAddr := fmt.Sprintf("127.0.0.1:%d", httpPort)
```
magiconair: You can inline this in the struct definition.

```go
return &TestServer{cmd, fmt.Sprintf("127.0.0.1:%d", httpPort)}, nil
```

discovery/consul_test.go
magiconair: Query `localhost:8500/v1/status/leader` for a non-empty `200 OK` response but do that as part of the `Start()` call. You can use/clone the `testutil/retry` package for that.

discovery/consul_test.go
```diff
@@ -77,18 +77,18 @@ a Consul server for testing. The 'consul' binary must be in the $PATH
 ref https://github.com/hashicorp/consul/tree/master/testutil
 */
 
-var testServer *testutil.TestServer
+var testServer *TestServer
```
magiconair: Don't use global state for tests. Pass the test server or better just the http address into the sub tests since that is all they need AFAICT.

cheapRoc: I agree, artifact of the previous implementation.

discovery/test_server.go
cheapRoc: Doh!

discovery/test_server.go
cheapRoc: I have a personal preference around inlining. Normally I enjoy the extra bits for readability and searching. In this case I agree, the line can be kept fairly short. 👍

discovery/consul_test.go
cheapRoc: Our tests don't require leadership election so I was able to just throw a naive GET request at `/v1/agent/self` which also uses `testutil/retry`.

cheapRoc:
Rebased against latest master and added @magiconair's CR comments. I also backed down to pinning Consul v1.0.0.

I believe we're in good shape.

Issue #525, opened 2017-11-08
Allow setting timeout to "never" for a job running on an interval.


eliasbrange:
We have multiple running jobs that runs every 15-30 seconds. In very rare cases, these jobs gets stuck. We do not want the jobs to be terminated automatically due to the uncertainty of the job states. Right now we set a huge timeout that will never be reached, and it would instead be nice to be able to set `never`.

One example is, every 30 seconds we poll a S3 bucket for .bak files and restore them if there are any. If the restore fails and gets stuck somehow, we do not want the job to start over until we have fixed the problem.

cheapRoc:
We really appreciate your feedback @eliasbrange. I want to make sure we talk this through to figure out what your needs/options are.

AFAIK most, if not all, `timeout` configuration in ContainerPilot is optional. Not including a `timeout` should dictate the `never` functionality you're requesting.

Another solution is to bake the exception/error handling into the process the job is executing. ContainerPilot has `timeout` features in order to handle it's own performance but any application it runs should subsequently handle it's own processing. Which leads me to want to understand your setup more closely.

It would be helpful if you could post a test configuration (or test repo I can run myself) so we can reference a working solution in this issue. Are you using version 2.x or 3.x?

eliasbrange:
Thanks for looking into it. I will try to be more clear:

According to the documentation the `timeout` defaults to the `when:interval` value. Our use case is that we run a job on an interval every 30 seconds, however some executions of that job might take longer than 30 seconds which will lead to a timeout if we do not explicitly set it to a higher value. Being able to tell the job to never timeout, even when running on an interval is what we are looking for.

I can post a configuration tomorrow, it's kinda late here now!

cheapRoc:
Ok, I was able to confirm the behavior after your last description. I've built a simple example setup that demonstrates the kill after the default timeout hits.

https://gist.github.com/cheapRoc/fb5a81ff590e898c5561d1851c827d9e

Here's the log output showing the process kill when a sleep occurs. 

```
time="2017-11-09T22:34:25Z" level=debug msg="control: initialized router for control server"
time="2017-11-09T22:34:25Z" level=debug msg="control: listening to /tmp/cp-single.socket"
time="2017-11-09T22:34:25Z" level=debug msg="event: {Startup global}"
time="2017-11-09T22:34:25Z" level=debug msg="test-repeats.Run start"
[...]

Issue #500, opened 2017-08-25
Potential race during reload removes control socket file on filesystem


cheapRoc:
This is the report for the issue that @tgross and I found this morning regarding reloading. Apparently there's a race condition which will cause the control socket file to be removed after the control server has reloaded.

The following log output demonstrates the race...

```
<reload sent>
...
DEBU[0019] control: initialized router for control server
DEBU[0019] control: listening to /tmp/cp-simple.socket
DEBU[0019] event: {Startup global}
INFO[0019] control: serving at /tmp/cp-simple.socket
DEBU[0019] main-process.Run start
DEBU[0019] main-process2.Run start
DEBU[0019] control: completed graceful shutdown of control server
```

This appears to be fixed in the early dev version of our pub/sub refactoring in-play for #476. Hopefully that will continue to be developed smoothly and we can close this bug then.

cheapRoc:
I forgot to mention that in order to test for this case in the future, we should add to the existing reload integration test whether the control socket is subsequently removed after reload. I think we can do this with a simple sleep and stat for the control socket file.

cheapRoc:
I have yet to see this crop up now that #512 has been merged. I'm going to close this since I'd like to think we've finally buried it into the ground. 😄

PR #533, opened 2017-11-22
Post release patches for various shutdown bits

cheapRoc:
This PR touches up a few areas of previously released shutdown patches.

- Use a composite literal of type `struct{}` as the sender type for the `completedCh` channel.
- Add a shared lock so that Jobs block when being marked `IsComplete`.
- Remove a redundant `cancel()` during app reload/shutdown.
- Clean-up signal handler since it's only run once per-app process.
- `SIGINT` and `SIGTERM` no longer exit the signal handler and can be called multiple times.

Most of these are just preventive measures for various bits around the new shutdown logic.

### DIFF
--- a/core/app.go
+++ b/core/app.go
@@ -118,7 +118,7 @@ func (a *App) Run() {
 		// signal handler above, and reload endpoint, only need to fire a
 		// GlobalShutdown across the event bus. Context handles everything after
 		// that process.
-		completedCh := make(chan bool)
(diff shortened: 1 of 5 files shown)


cheapRoc:
I should note, some of these came about due to #525 and #500. Having hit a few edge-y race conditions involving shutdown made me want to make some small adjustments. I haven't hit them since, god willing.

Issue #536, opened 2017-12-04
Duplicate metric events in Telemetry endpoint


eliasbrange:
We use `containerpilot -putmetric` to gather certain fluctuating metrics such as memory usage and free disk space. We noticed that after running a container in production for some time the `/metrics` endpoint generates over 100k lines with lines on the form

```
containerpilot_events{code="Metric",source="metric|value"} 1
```

Since the metrics are fluctuating this means that we get a few new events every 15 seconds or so. My guess is that this will continue indefinitely until the container is restarted. For counters that can only ever go up this also means a new event line every time you do `putmetric`.

My question is, is this expected behaviour? Filtering in e.g. `InfluxDB` on `source` is now impossible, and the endpoint is getting slower and slower due to the massive amount of events that are returned.

**Edit:** I have experimented with version `3.4.2`, `3.5.0` and `3.6.0` and they seem to behave the same.

cheapRoc:
Thanks for reporting this @eliasbrange.

Initially the behavior you're describing doesn't sound like what I'd expect on a Prom formatted endpoint. I would expect a single metric counter to increase monotonically and not output multiple lines or counters with a single event value. 

Could you paste a snippet of everything else being logged around when this starts (I don't need all the lines, obviously)?

Is it just a single metric named `metric|value`? Are any other custom metrics performing this way, possibly named in another manner?

A test configuration/setup is always helpful as well and helps me debug quicker but I'm quite sure I can reproduce this. Sounds like I only need to push metrics into ContainerPilot.

eliasbrange:
I guess the config doesn't actually matter. And defining a metric in the config does not matter either I think. Anyways, here comes a little config.

```
{
  consul: "localhost:8500",
  logging: {
    level: "DEBUG",
    format: "default",
    output: "stdout"
  },
  telemetry: {
    port: 9090,
    interfaces: ["static:127.0.0.1"],
    metrics: [
      {
        namespace: "telemetry",
        subsystem: "test",
        name: "counter",
        help: "test",
        type: "counter"
      }
    ]
  }
}
```

Then I run the following commands with the following output:

```
$ curl --silent localhost:9090/metrics | grep telemetry_test_counter
> # HELP telemetry_test_counter test                                          
> # TYPE telemetry_test_counter counter                                       
> telemetry_test_counter 0                                                    

$ for i in {1..10}; do containerpilot -putmetric telemetry_test_counter=$i; done

[...]

PR #538, opened 2017-12-06
Prevent overzealous collection of Metric events through Prometheus

cheapRoc:
Ref: #536 

This prevents `Metric` events from being collected by the event bus Prometheus collector. This removes the issue where any number of unique values hitting a counter or gauge could be recorded as unique Metric events. This would eventually cause the endpoint to get sluggish from so many unique events.


cheapRoc:
There's probably some broken tests for this but I wanted to at least push what I had tested.

Issue #549, opened 2018-02-21
Inconsistent default value for timeouts between job exec and health checks.


eliasbrange:
I've noticed an inconsistency between the default value for timeouts between a job exec timeout and health check timeout.

For jobs not running on an interval, the timeout defaults to `0` if left out which makes the job exec never time out. For jobs running on an interval the timeout defaults to be the same as the interval as seen here: https://github.com/joyent/containerpilot/blob/7208e025644e9d9b0ca06d26ee97a4c612af7788/jobs/config.go#L240

For health checks, if the timeout configuration is left out it defaults to the job exec timeout as seen here: https://github.com/joyent/containerpilot/blob/7208e025644e9d9b0ca06d26ee97a4c612af7788/jobs/config.go#L300

This seems off, why should it default to the timeout of the job it is health checking? 

And also, the above code always sets the health check timeout to `0` since `validateHealthCheck` is executed before `validateExec`, which means that `cfg.execTimeout` which is initialized in `validateExec` is always `0` when running `validateHealthCheck`.

A more sensible default might be to set `checkTimeout = cfg.heartbeatInterval` to make it behave the same as the default for exec timeouts.

cheapRoc:
Released in [3.8.0](https://github.com/joyent/containerpilot/releases/tag/3.8.0).

PR #550, opened 2018-02-21
Set health timeout to health interval if omitted.

eliasbrange:
This sets the health check timeout to be the same as the health check interval if the timeout configuration is omitted. This is to make the configuration for health check timeouts and job exec timeouts consistent. Read more in the issue https://github.com/joyent/containerpilot/issues/549.


cheapRoc:
I haven't had the cycles yet to review your issues/changes. I'll get back to you on them, apologies.

PR #553, opened 2018-04-02
Register critical services in consul.

### LINKED ISSUE #552: Register critical services in consul.
Howdy, we currently have about 200 services in consul, and filtering them on only failed ones is a quick way for us to see that everything is up and running. However, containerpilot only registers a service after it has become healthy for the first time due to `job.sendHeartbeat()` being called in `onHealthCheckPassed`. 

If for some reason the service never gets healthy, the failed filter in consul does not give us that information. Currently, we have to manually search for the service that usually misbehave on startup.

It would be great to register a service as critical in consul in `onHealthCheckPassed`, but only if the services hasn't been seen before to not mess with the TTL.

eliasbrange:
Fixes #552 

This PR adds the functionality to register services in consul that never gets healthy to be able to filter on failed services in consul and see that they are misbehaving. 

I've added a method for registering a service with the status set to critical. This is called in `onHealthChechFailed`, but only if the service hasn't already been registered. I've also made some minor refactoring in `services/services.go`, i.e. having `MarkForMaintenance` call `Deregister` instead of the other way around.

### DIFF
--- a/discovery/service.go
+++ b/discovery/service.go
@@ -15,7 +15,7 @@ type ServiceDefinition struct {
 	Port                           int
 	TTL                            int
 	Tags                           []string
-	InitialStatus									 string
+	InitialStatus                  string
 	IPAddress                      string
 	EnableTagOverride              bool
 	DeregisterCriticalServiceAfter string
--- a/discovery/service.go
+++ b/discovery/service.go
@@ -15,6 +15,7 @@ type ServiceDefinition struct {
 	Port                           int
 	TTL                            int
 	Tags                           []string
+	InitialStatus									 string
 	IPAddress                      string
 	EnableTagOverride              bool
 	DeregisterCriticalServiceAfter string
@@ -49,9 +50,29 @@ func (service *ServiceDefinition) SendHeartbeat() error {
 	return nil
 }
 
-// RegisterUnhealthy registers the service with status set to critical.
-func (service *ServiceDefinition) RegisterUnhealthy() {
-	service.register(api.HealthCritical)
+// Register the service with its configured initial status.
+func (service *ServiceDefinition) RegisterWithInitialStatus() {
+	if service.wasRegistered {
+		return
+	}
+
+	status := ""
+
+	switch service.InitialStatus {
+		case "passing":
+			status = api.HealthPassing
+			break
+		case "warning":
+			status = api.HealthWarning
+			break
+		case "critical":
+			status = api.HealthCritical
+			break
+	}
+
+	log.Infof("Registering service %v with initial status set to %v",
+	          service.Name, service.InitialStatus)
+	service.register(status)
 }
 
 // Register registers the service with the given status in Consul.
--- a/discovery/service.go
+++ b/discovery/service.go
@@ -25,43 +25,51 @@ type ServiceDefinition struct {
 
 // Deregister removes the service from Consul.
 func (service *ServiceDefinition) Deregister() {
-	service.MarkForMaintenance()
-}
-
-// MarkForMaintenance removes the service from Consul.
-func (service *ServiceDefinition) MarkForMaintenance() {
 	log.Debugf("deregistering: %s", service.ID)
 	if err := service.Consul.ServiceDeregister(service.ID); err != nil {
 		log.Infof("deregistering failed: %s", err)
 	}
 }
 
-// SendHeartbeat writes a TTL check status=ok to the consul store.
-// If consul has never seen this service, we register the service and
-// its TTL check.
+// MarkForMaintenance removes the service from Consul.
+func (service *ServiceDefinition) MarkForMaintenance() {
+	service.Deregister()
+}
+
+// SendHeartbeat writes a TTL check status=ok to the Consul store.
 func (service *ServiceDefinition) SendHeartbeat() error {
-	if !service.wasRegistered {
-		if err := service.registerService(); err != nil {
-			log.Warnf("service registration failed: %s", err)
-			return err
-		}
-		service.wasRegistered = true
-		return nil
-	}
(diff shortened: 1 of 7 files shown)


jobs/jobs.go
```diff
@@ -98,8 +98,16 @@ func FromConfigs(cfgs []*Config) []*Job {
...
+// registerUnhealthyService registers this Job's service with status set to
+// critical. Will only register the service if it does not already exist.
+func (job *Job) registerUnhealthyService() {
+	if job.Service != nil {
+		job.Service.RegisterUnhealthy()
+	}
+}
+
+// sendHeartbeat sends a heartbeat for this Job's service
+func (job *Job) sendHeartbeat() {
```
cheapRoc: I figured since the tests pass this isn't a problem but what were your thoughts on changing this from an exported method? `Job.SendHeartbeat` => `Job.sendHeartbeat`?

cheapRoc:
Overall I think this is a great change. I can definitely imagine services that never come up to be critical state yet still registered in a service catalog. One question in the review otherwise I think we can include this in a next minor bump.

cheapRoc:
Actually have a few more thoughts I'll post in the issue.

jobs/jobs.go
eliasbrange: I had a hard time seeing how that method will be used outside of the job itself, but it's a small change so I can change it back if you'd like.

eliasbrange:
Thanks for the feedback, I will try to address those thoughts in the issue.

misterbisson:
@eliasbrange I'd love your feedback on @cheapRoc's suggestion about `initial_status` in https://github.com/joyent/containerpilot/issues/552, and if that works we'd also need to update the docs in this PR to correspond with the change .

discovery/service.go
```diff
@@ -72,7 +101,7 @@ func (service *ServiceDefinition) registerService() error {
 			EnableTagOverride: service.EnableTagOverride,
 			Check: &api.AgentServiceCheck{
```
cheapRoc: After looking up that this is using Consul's `api.AgentServiceCheck` I believe the configuration option should be on the job level and _not_ within the health check. Optionally, we could expose the same `initial_status` option for all health checks independently... but I think that should be a future enhancement warranted from user requests for the feature.

jobs/config.go
```diff
@@ -24,6 +24,7 @@ type Config struct {
 
 	// service discovery
 	Port              int           `mapstructure:"port"`
+	InitialStatus     string        `mapstructure:"initialStatus"`
```
cheapRoc: Let's call it `initial_status` (snake case), I don't believe we use lower camel case anywhere else. In Go we'll continue to use `InitialStatus`.

jobs/jobs.go
```diff
@@ -98,13 +98,20 @@ func FromConfigs(cfgs []*Config) []*Job {
 	return jobs
 }
 
-// SendHeartbeat sends a heartbeat for this Job's service
-func (job *Job) SendHeartbeat() {
+// sendHeartbeat sends a heartbeat for this Job's service
+func (job *Job) sendHeartbeat() {
```
cheapRoc: Rename `sendHeartbeat` back to exported method `SendHeartbeat`.

jobs/jobs.go
```diff
@@ -159,6 +166,9 @@ func (job *Job) Run(pctx context.Context, completedCh chan struct{}) {
 			completedCh <- struct{}{}
 		}()
 		for {
+			// Check if job's service has been registered. Doing it inside the event
+			// loop to retry if consul registration fails.
+			job.checkRegistration()
```
cheapRoc: 👍 Looks handy all by itself.

jobs/testdata/TestJobConfigServiceWithInitialStatus.json5
```diff
@@ -0,0 +1,18 @@
...
+    exec: "/bin/serviceA",
+    health: {
+      exec: "/bin/to/healthcheck/for/service/A.sh",
+      interval: 10,
+      ttl: 30,
+    },
+  },
+  {
+    name: "serviceB",
+    initialStatus: "invalid-value",
```
cheapRoc: Rename to `initial_status` here.

cheapRoc:
Thank you again @eliasbrange for your continued support and patience.

eliasbrange:
Reverted `SendHeartbeat` change and renamed to `initial_status` @cheapRoc .

jobs/config.go
cheapRoc: Ah yes, you're right.

cheapRoc:
I tested this PR with the following (rather convoluted) test rig using Fabio, `http-echo`, and the following configuration file.

```json5
{
  consul: "localhost:8500",
  logging: {
    level: "DEBUG",
    format: "text"
  },
  jobs: [
    {
      name: "foo",
      exec: "http-echo -listen=:8484 -text=hello",
      initial_status: "warning",
      tags: ["urlprefix-/foo"],
      port: 8484,
      health: {
        exec: "/usr/bin/curl -o /dev/null -sfI http://localhost:8484/",
        interval: 60,
        ttl: 60,
      },
    },
  ],
  control: {
    socket: "/tmp/cp.socket"
  }
}
```

I was able to verify that the Consul service check was initially set to `warning` status.

screen shot 2018-04-12 at 10 54 26 pm

With the health check set to `60` seconds I was able to wait and see the health check passing.

screen shot 2018-04-12 at 10 54 53 pm

Once again, I want to thank you for a wonderful addition to ContainerPilot @eliasbrange!

eliasbrange:
Thanks for including it @cheapRoc. Will make it much easier for our monitoring!

PR #574, opened 2018-12-04
fix template/parseEnvironment function

### LINKED ISSUE #573: Error parsing environment variable in config template
ContainerPilot version: 3.8.0

If the environment variable value contains an equal sign, `=`, everything is trimmed from that point. For example:

Environment variable `ABC` is set to `abc=cde==`
Template has `{{ .ABC }}`
Output is just `abc` and the `=cde==` part is trimmed.

MunifTanjim:
fixes #573

### DIFF
--- a/config/template/template.go
+++ b/config/template/template.go
@@ -54,7 +54,7 @@ func parseEnvironment(environ []string) Environment {
 		return env
 	}
 	for _, e := range environ {
-		kv := strings.Split(e, "=")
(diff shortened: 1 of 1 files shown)


mgerdts:
Thanks for this PR.  The code looks good, but we could use a test in config/template/template_test.go.

MunifTanjim:
@mgerdts added a test case

MunifTanjim:
@mgerdts any update on this?

